PR #6, opened 2019-12-24
Update faq.rst

nrolland:
fixing the command to force recalling substituters by deleting the local cache


domenkozar:
best to use an asterix

PR #33, opened 2020-06-12
Add rec anti pattern

### LINKED ISSUE #29: Anti pattern: rec

domenkozar:
Fixes #29

### DIFF
--- a/source/anti-patterns/language.rst
+++ b/source/anti-patterns/language.rst
@@ -1,5 +1,5 @@
-Nix expressions
-===============
+Nix language
+============
 
 
 Unqouted URLs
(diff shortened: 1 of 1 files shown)


source/anti-patterns/language.rst
```diff
@@ -9,10 +9,46 @@ Nix syntax supports URLs as verbatim, so one can write ``https://example.com`` i
...
+    b = a + 2;
+  }
+
+evaluating to ``{ a = 1; b = 3; }``.
+
+``b`` refers to ``a`` as ``rec`` makes all keys available within.
+
+There are a couple of pitfals:
+
+- it's possible to introduce recursion with the wrong shadowing, resulting in a weird error message
```
Valodim, line 34: resulting in "a" weird error message? so what does that look like?

also, is "shadowing" a term that's clear on its own, or would that perhaps benefit from a link to a place where it's explained?

source/anti-patterns/language.rst
```diff
@@ -9,10 +9,46 @@ Nix syntax supports URLs as verbatim, so one can write ``https://example.com`` i
...
+
+- combining with overriding logic such as `overrideAttrs` function in nixpkgs it has suprising behavour
```
Valodim, line 36: "surprising behavior" is a very fuzzy term here, can we find at least a rough description of what the bad things™ are that might happen?

source/anti-patterns/language.rst
```diff
@@ -9,10 +9,46 @@ Nix syntax supports URLs as verbatim, so one can write ``https://example.com`` i
...
+  rec {
+    a = 1;
+    b = a + 2;
+  }
+
+evaluating to ``{ a = 1; b = 3; }``.
+
+``b`` refers to ``a`` as ``rec`` makes all keys available within.
+
+There are a couple of pitfals:
```
Valodim, line 32: I always wondered, is there a notable difference in performance with `rec` attrsets?

source/anti-patterns/language.rst
domenkozar: Good point, made some changes to address that.

source/anti-patterns/language.rst
domenkozar: Same here.

source/anti-patterns/language.rst
domenkozar: Me too, but that's out of scope here :)

source/anti-patterns/language.rst
```diff
@@ -9,10 +9,48 @@ Nix syntax supports URLs as verbatim, so one can write ``https://example.com`` i
...
+  rec {
+    a = 1;
+    b = a + 2;
+  }
+
+evaluating to ``{ a = 1; b = 3; }``.
+
+``b`` refers to ``a`` as ``rec`` makes all keys available within the attribute set.
+
+There are a couple of pitfals:
```
cransom, line 32: Spelling on `pitfalls`

source/anti-patterns/language.rst
Valodim: Well, I think if use of `rec` has a significant impact on performance, it might be another reason to avoid it :)

domenkozar: Right, going to add that once someone confirms it :)

domenkozar:
Feel free to review further, going to merge this for now :)

PR #38, opened 2020-06-13
feature(thispage): add edit, history and source links

ghuntley:


domenkozar:
I think you need to wrap the template in `{%- if show_source and has_source and sourcename %}`

domenkozar:
Thanks! <3

PR #42, opened 2020-06-14
Minor improvements

qrilka:
This is a bit subjective for s/`/``/ replacement but even in HTML generated by Sphinx gray spans on https://nix.dev/anti-patterns/language.html look much more prominent than italic (which is easy to miss and sentences using "with" become a bit confusing)


qrilka:
No idea what CI complaint is about :(

domenkozar:
Yeach CLA check is kind of broken. Thanks!

PR #78, opened 2020-07-09
reproducibile shell edits

sixhobbits:
small edits

### DIFF
--- a/source/tutorials/declarative-and-reproducible-developer-environments.rst
+++ b/source/tutorials/declarative-and-reproducible-developer-environments.rst
@@ -1,9 +1,6 @@
-.. _declarative-reproducible-envs:
-
 Declarative and reproducible developer environments
 ===================================================
-
-In :ref:`ad-hoc-envs` tutorial we took a dive into providing shell
+In the :ref:`ad-hoc-envs` tutorial we looked at providing shell
 environments for when we need a quick'n'dirty way of getting hold
 of some tools.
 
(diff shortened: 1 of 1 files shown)


sixhobbits:
@domenkozar these are some smaller edits/touch ups. My larger questions after reading this one which I think would be good to address are 

* Why might it be useful to run the bash hook every time I run the shell (I imagine it can be used as part of git workflows etc, but mentioning a 'real world' example to supplement the 'hello' demonstration would help clarify this I think
* Why is the ``direnv`` command useful - I imagine developers would not need to modify their project shell.nix file that often and would probably restart the shell at least daily when they begin working on a project anyway? Or is that not a common workflow and therefore it's worth installing ``direnv`` globally on all dev machines to avoid having to restart the shell?

sixhobbits:

sixhobbits:
recheckcla

domenkozar:
>    Why might it be useful to run the bash hook every time I run the shell (I imagine it can be used as part of git workflows etc, but mentioning a 'real world' example to supplement the 'hello' demonstration would help clarify this I think

That's a good point! I'll add an example.

>   Why is the direnv command useful - I imagine developers would not need to modify their project shell.nix file that often and would probably restart the shell at least daily when they begin working on a project anyway? Or is that not a common workflow and therefore it's worth installing direnv globally on all dev machines to avoid having to restart the shell?

It's more that you often switch between projects during development and you want the project to just work.

It depends a bit on the worfklow, if you have one project and a terminal open all the time, then it's not that useful :)

domenkozar:
Merging these changes since they look good! Thanks :)

domenkozar:
Opened https://github.com/nix-dot-dev/nix.dev/issues/80 for your feedback :)

PR #110, opened 2020-11-18
Add tutorial for deploying NixOS with Terraform

domenkozar:
cc @zimbatm @zupo @grahamc 

I haven't really tested this out, so it's a draft.

If for some reason this is incorrect nor the best practice let me know :)

[preview](https://deploy-preview-110--nixdev.netlify.app/tutorials/deploying-nixos-using-terraform.html)

### DIFF
--- a/source/tutorials/deploying-nixos-using-terraform.rst
+++ b/source/tutorials/deploying-nixos-using-terraform.rst
@@ -19,7 +19,7 @@ Booting NixOS image
 
    nix-shell -p terraform
 
-2. We are using `Terraform Cloud <app.terraform.io/>`_ as a `state/locking backend <https://www.terraform.io/docs/state/purpose.html>`_:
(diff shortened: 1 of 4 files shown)


domenkozar:
Alright, I have tested it now and it works. Going to merge tomorrow morning.

source/tutorials/deploying-nixos-using-terraform.rst
```diff
@@ -0,0 +1,179 @@
+.. _deploying-nixos-using-terraform:
+
+Deploying NixOS using Terraform
```
andir, line 3: `s/Terraform/Terraform Cloud/`

While it can probably be translated to the non cloud version this would be more correct right now.

domenkozar: That's not really correct, since it doesn't run on Cloud but it only stores the state there.

PR #152, opened 2021-06-03
Add tutorial for integration testing using virtual machines

### LINKED ISSUE #111: NixOS tests

domenkozar:
Fixes #111

[rendered](https://fab33bb8.nix-dev.pages.dev/tutorials/integration-testing-using-virtual-machines)

cc @zimbatm @zupo @NorfairKing

### DIFF
--- a/source/tutorials/integration-testing-using-virtual-machines.rst
+++ b/source/tutorials/integration-testing-using-virtual-machines.rst
@@ -1,0 +1,231 @@
+Integration testing using Virtual Machines
+==========================================
+
+One of the most powerful features in the Nix ecosystem is **the ability
+to provide a set of declarative NixOS configurations and use a simple
+Python interface** to interact with them using `QEMU <https://www.qemu.org/>`_
+as the backend.
+
+Those tests are widely used to ensure that NixOS works as intended, so in general they are called **NixOS tests**.
+They can be written and launched outside of NixOS, on any Linux machine (with 
+`MacOS support coming soon <https://github.com/NixOS/nixpkgs/issues/108984>`_).
+
+Integration tests are reproducible due to the design properties of Nix,
+making them a valuable part of a Continuous Integration pipeline.
+
+
+Testing a typical web application backed by PostgreSQL
+------------------------------------------------------
+
+This tutorial follows `PostgREST tutorial <https://postgrest.org/en/stable/tutorials/tut0.html>`_,
+a generic `RESTful API <https://restfulapi.net/>`_ for PostgreSQL.
+
+If you skim over the official tutorial, you'll notice there's quite a bit of setup
+in order to test if all the steps work.
+
+We are going to set up:
+
+- VM named ``server`` running postgreSQL and postgREST.
+
+- VM named ``client`` running HTTP client queries using ``curl``.
+  
+- A ``testScript`` orchestrating testing logic between ``client`` and ``server``.
+
+
+Writing the test
+----------------
+
+Create ``postgrest.nix``:
+
+.. TODO: highlight nix https://github.com/pygments/pygments/issues/1793
+
+
+.. code-block::
+  :linenos:
+
+  let
+    # Pin nixpkgs, see pinning tutorial for more details
+    nixpkgs = fetchTarball "https://github.com/NixOS/nixpkgs/archive/0f8f64b54ed07966b83db2f20c888d5e035012ef.tar.gz";
+    pkgs = import nixpkgs {};
+
+    # Single source of truth for all tutorial constants
+    database      = "postgres";
+    schema        = "api";
+    table         = "todos";
+    username      = "authenticator";
+    password      = "mysecretpassword";
+    webRole       = "web_anon";
+    postgrestPort = 3000;
+
+    # NixOS module shared between server and client
+    sharedModule = {
+      # Since it's common for CI not to have $DISPLAY available, we have to explicitly tell the tests "please don't expect any screen available"
+      virtualisation.graphics = false;
+    };
+
+  in pkgs.nixosTest ({
+    # NixOS tests are run inside a virtual machine, and here we specify system of the machine.
+    system = "x86_64-linux";
+
+    nodes = {
+      server = { config, pkgs, ... }: {
+        imports = [ sharedModule ];
+
+        networking.firewall.allowedTCPPorts = [ postgrestPort ];
+
+        services.postgresql = {
+          enable = true;
+
+          initialScript = pkgs.writeText "initialScript.sql" ''
+            create schema ${schema};
+
+            create table ${schema}.${table} (
+                id serial primary key,
+                done boolean not null default false,
+                task text not null,
+                due timestamptz
+            );
+
+            insert into ${schema}.${table} (task) values ('finish tutorial 0'), ('pat self on back');
+
+            create role ${webRole} nologin;
+            grant usage on schema ${schema} to ${webRole};
+            grant select on ${schema}.${table} to ${webRole};
+
+            create role ${username} inherit login password '${password}';
+            grant ${webRole} to ${username};
+          '';
+        };
+
+        users = {
+          mutableUsers = false;
+          users = {
+            # For ease of debugging the VM as the `root` user
+            root.password = "";
+
+            # Create a system user that matches the database user so that we
+            # can use peer authentication.  The tutorial defines a password,
+            # but it's not necessary.
+            "${username}".isSystemUser = true;
+          };
+        };
+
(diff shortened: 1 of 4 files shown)


source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+    webRole       = "web_anon";
+    postgrestPort = 3000;
+
+    # NixOS module shared between server and client
+    sharedModule = {
+      # don't assume graphical interface is possible
+      virtualisation.graphics = false;
+    };
+
+  in import "${nixpkgs}/nixos/tests/make-test-python.nix" ({ pkgs, ...}: {
```
tfc, line 68: you can use `pkgs.nixosTest` as a function to call

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+      ]
+
+      actual = json.loads(
+          client.succeed(
+              "${pkgs.curl}/bin/curl http://server:${toString postgrestPort}/${table}"
+          )
+      )
+
+      if expected != actual:
+          sys.exit(1)
```
tfc, line 158: does anything speak against this?

```python
assert expected == actual, "table query returns expected content"
```

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+
+
+Next steps
+----------
+
+- Running integration tests on CI requires hardware acceleration, which many CIs do not support. 
+  To run integration tests on :ref:`GitHub Actions <github-actions>` see 
+  `how to disable hardware acceleration <https://github.com/cachix/install-nix-action#how-can-i-run-nixos-tests>`_.
+
+- NixOS comes with a large set of tests that serve also as educational examples, a good inspiration is `Matrix bridging with an IRC <https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/matrix-appservice-irc.nix>`_.
```
tfc, line 228: my personal favorite example is the bittorrent test. i can also provide a nice picture that shows what kind of complexity the test maintains with few lines that i would share if you are interested.

Screenshot from 2021-06-03 14-18-08

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+    table         = "todos";
+    username      = "authenticator";
+    password      = "mysecretpassword";
+    webRole       = "web_anon";
+    postgrestPort = 3000;
+
+    # NixOS module shared between server and client
+    sharedModule = {
+      # don't assume graphical interface is possible
+      virtualisation.graphics = false;
```
NorfairKing, line 65: Why is this necessary? Does it speed up the test? If so, I should add that to my own tests.

domenkozar: It makes sure the tests works on CI where there's no $DISPLAY.

source/tutorials/integration-testing-using-virtual-machines.rst
domenkozar: That one is pretty huge, I'd like to give smaller examples that demonstrate the power of what it can do.

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
+Integration testing using Virtual Machines
+==========================================
+
+One of the most powerful features in the Nix ecosystem is **the ability
+to provide a set of declarative NixOS configurations and use a simple
+Python interface** to interact with them using `QEMU <https://www.qemu.org/>`_
+as the backend.
+
+They are also known as **NixOS tests**.
```
Patryk27, line 9: Maybe it'd be worth pointing out that those tests don't require using NixOS as the host system?

Something in terms of:

> Those tests are widely used to ensure that NixOS works as intended, so in general they are called **NixOS tests** - though they can be written and launched outside of NixOS, on any Linux machine (with MacOS support coming soon).

source/tutorials/integration-testing-using-virtual-machines.rst
Patryk27: This one caught me off-guard, too - how about phrasing such as this?

> since it's common for CI not to have $DISPLAY available, we have to explicitly tell the tests "please don't expect any screen available"

(this concrete wording is kinda flake-y, but hopefully will work as a sales pitch here 🙂)

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+    sharedModule = {
+      # it's common for CI not to have $DISPLAY available
+      virtualisation.graphics = false;
+    };
+
+  in pkgs.nixosTest ({
+    # In case we're evaluating from macOS, pin the system
+    system = "x86_64-linux";
+
+    # Disable linting for simpler debugging
```
Patryk27, line 73: Because the `testScript =` thingie this parameter refers to appears waay below, how about we either move this option below `testScript`, or change the comment a bit:

> In a few lines below, there's an option called `testScript` - it contains the test's body written in Python. By the default, before the test is executed, `pkgs.nixosTest` ensures the Python code is neatly formatted and, if it's not, prevents the test from running. This is great in the long run, but makes the development a bit harder and so for the time being we're disabling linting that Python code.

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+    postgrestPort = 3000;
+
+    # NixOS module shared between server and client
+    sharedModule = {
+      # it's common for CI not to have $DISPLAY available
+      virtualisation.graphics = false;
+    };
+
+  in pkgs.nixosTest ({
+    # In case we're evaluating from macOS, pin the system
```
Patryk27, line 70: How about something like this?

> NixOS tests are run inside a virtual machine, and here we specify which machine should get emulated. Currently only `x86_64-linux` is supported.

(though tbh i'm not sure if only `x86_64-linux` is supported - can it actually virtualize _arbitrary_ qemu-and-nixos-compatible machine? I've never checked it before 😄)

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+
+  $ nix-build postgrest.nix
+
+You'll notice an error message if something goes wrong.
+
+In case the tests succeed, you should see at the end:
+
+.. code:: shell-session
+
+  1 out of 1 tests succeeded
```
Patryk27, line 187: Hmm, in my case what's returned is:
```
<!-- logs from the VMs -->
client: must succeed: /nix/store/g8w72aanshmw5gxi02ril1qlb427rjzy-curl-7.76.1-bin/bin/curl http://server:3000/todos
client: waiting for the VM to finish booting
client: connected to guest root shell
client: (connecting took 0.00 seconds)
(0.00 seconds)
server # [   10.554648] postgrest-start[908]: 192.168.1.1 - - [03/Jun/2021:17:53:41 +0000] "GET /todos HTTP/1.1" 200 - "" "curl/7.76.1"
(0.09 seconds)
(10.93 seconds)
test script finished in 10.96s
cleaning up
killing client (pid 10)
killing server (pid 22)
(0.00 seconds)
/nix/store/bx7z3imvxxpwkkza10vb23czhw7873w2-vm-test-run-unnamed
```

Patryk27 (commented):
❤️

source/tutorials/integration-testing-using-virtual-machines.rst
```diff
@@ -0,0 +1,228 @@
...
+
+
+Developing and debugging tests
+------------------------------
+
+When developing tests or when something breaks, it's useful to interactively fiddle
+with the script or access a terminal for a machine.
+
+To interactively start a Python session with testing framework:
```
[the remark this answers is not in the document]
domenkozar: I try hard to not offer two solutions to the same problem and since `nixos-run-vms` doesn't allow fiddling with the test script, it's inferior to the driver method. Does that make sense?

source/tutorials/integration-testing-using-virtual-machines.rst
domenkozar: `aarch64-linux` should work too I imagine.

domenkozar:
PTAL :)

tzarsquared:

PR #153, opened 2021-06-06
Flakify with poetry2nix

garbas:
Here is a list of changes I did:
- switch from `requirements.txt` to `pyproject.toml` (via poetry)
- use `poetry2nix` to get a reproducible python environment
- fix `flake.nix` since it was broken. `nix develop` and `nix build` work.
- ensure `nix-shell` and `nix-build` still work (via `flake-compat`)
- use nix for ci

I hope that Cloudflare Pages work with `pyproject.toml`, if not we can figure something out (eg. like pushing from Github Actions or something else)

This will help me to be able to continue syncing nix.dev to nixos.org.

PR #156, opened 2021-06-08
Flakify with poetry2nix

garbas:
was #153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,20 +8,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2.3.4
-    - uses: actions/setup-python@v2
-      with:
-        python-version: 3.8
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -8,7 +8,12 @@ jobs:
...
-    - uses: ammaraskar/sphinx-action@0.4
       with:
-        docs-folder: "."
-        build-command: make html linkcheck
+        fetch-depth: 0
+    - uses: cachix/install-nix-action@v13
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - name: Setup tmate session
+      uses: mxschmitt/action-tmate@v3
```
domenkozar, line 17: @garbas won't this block?

domenkozar: The problem is that it never gets past this action as it just listens for SSH?

garbas: @domenkozar I removed the tmate session. Can you approve the build once again?

Issue #113, opened 2020-11-30
Cross compile


domenkozar:

domenkozar:
https://nix.dev/tutorials/cross-compilation

Issue #157, opened 2021-06-10
Remote building


domenkozar:
Existing tutotrials:

- https://nixos.org/manual/nix/unstable/advanced-topics/distributed-builds.html
- https://eipi.xyz/blog/nix-aarch64-remote/
- https://eipi.xyz/blog/building-on-remote-server-for-nixos-on-raspberry-pi/

Notes:

- builders-use-substitutes = true
- https://github.com/nix-dot-dev/nix.dev/pull/123#issuecomment-889786823

Eisfunke:
I have a RPi3 onto which I deploy NixOS remotely using QEMU, I have some notes on that here: https://www.eisfunke.com/article/nixos-on-raspberry-pi.html#installing, might be helpful as well..

fricklerhandwerk:
Troubleshooting tips: https://github.com/NixOS/nix/pull/5345

fricklerhandwerk:
Duplicated in https://github.com/NixOS/nix.dev/issues/1013.
Closed by https://github.com/NixOS/nix.dev/pull/1045.

PR #160, opened 2021-06-14
Tutorial for cross-compilation

### LINKED ISSUE #148: Cross-compilation tutorial
- https://nixos.wiki/wiki/Cross_Compiling
- https://matthewbauer.us/blog/beginners-guide-to-cross.html

domenkozar:
Fixes #148 #113 

[rendered](https://b5839745.nix-dev.pages.dev/tutorials/cross-compilation)

### DIFF
--- a/source/tutorials/cross-compilation.rst
+++ b/source/tutorials/cross-compilation.rst
@@ -1,0 +1,257 @@
+.. _ref-cross-compilation:
+
+Cross-compilation
+=================
+
+Cross-compilation is the act of **compiling code** on the **build platform**
+to the **host platform**, where the compiled **executable runs**. [#]_
+
+It's needed when the host platform has limited resources (such as CPU)
+or when it's not easily accessible for development.
+
+Nix community has world-class support for cross-compilation,
+after years of hard work from our community.
+
+.. [#] Terminology for cross-compilation platforms differs between build systems,
+       Nix community has chosen to follow 
+       `autoconf terminology <https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html>`_.
+
+.. note:: macOS/Darwin is a special case, as not the whole OS is Open Source. 
+          It's only possible to cross-compile between ``aarch64-darwin`` and ``x86_64-darwin``.
+
+
+What's a target platform?
+-------------------------
+
+There's actually a third platform named target.
+
+It matters in cases where you'd like to distribute a compiler binary, 
+as you'd then like to build a compiler on the build platform, compile code on the
+target plaform and run the final executable on the host platform.
+
+Since that's rarely needed, we'll treat target platform the same as the build.
+
+
+Determining the host platform
+-----------------------------
+
+The build platform is determined automatically by Nix
+as it can just guess it during the configure phase.
+
+The host platform is best determined by running on the host platform:
+
+.. code:: shell-session 
+
+  $ bash $(nix-build '<nixpkgs>' -A gnu-config)/config.guess
+  aarch64-unknown-linux-gnu
+
+In case that's not possible (when the host platform is not easily accessible
+for development), it has to be constructed manually via the following template:
+
+.. code::
+
+  <cpu>-<vendor>-<os>-<abi>
+
+Note that ``<vendor>`` is often ``unknown`` and ``<abi>`` is optional. 
+There's also no unique identifier for a platform, for example ``unknown`` and 
+``pc`` are interchangeable (hence it's called config.guess).
+
+Some other common examples of platforms:
+
+- aarch64-apple-darwin14
+- aarch64-pc-linux-gnu
+- x86_64-w64-mingw32
+- aarch64-apple-ios
+
+
+Choosing the host platform with Nix
+-----------------------------------
+
+Nixpkgs comes with a set of predefined host plaform applied to all packages.
+
+It's possible to list predefined sets via shell completion:
+
+.. code:: shell-session
+
+  $ nix-build '<nixpkgs>' -A pkgsCross.<TAB>
+  pkgsCross.aarch64-android             pkgsCross.musl32
+  pkgsCross.aarch64-android-prebuilt    pkgsCross.musl64
+  pkgsCross.aarch64be-embedded          pkgsCross.muslpi
+  pkgsCross.aarch64-darwin              pkgsCross.musl-power
+  pkgsCross.aarch64-embedded            pkgsCross.or1k
+  pkgsCross.aarch64-multiplatform       pkgsCross.pogoplug4
+  pkgsCross.aarch64-multiplatform-musl  pkgsCross.powernv
+  pkgsCross.amd64-netbsd                pkgsCross.ppc64
+  pkgsCross.arm-embedded                pkgsCross.ppc64-musl
+  pkgsCross.armhf-embedded              pkgsCross.ppc-embedded
+  pkgsCross.armv7a-android-prebuilt     pkgsCross.ppcle-embedded
+  pkgsCross.armv7l-hf-multiplatform     pkgsCross.raspberryPi
+  pkgsCross.avr                         pkgsCross.remarkable1
+  pkgsCross.ben-nanonote                pkgsCross.remarkable2
+  pkgsCross.fuloongminipc               pkgsCross.riscv32
+  pkgsCross.ghcjs                       pkgsCross.riscv32-embedded
+  pkgsCross.gnu32                       pkgsCross.riscv64
+  pkgsCross.gnu64                       pkgsCross.riscv64-embedded
+  pkgsCross.i686-embedded               pkgsCross.scaleway-c1
+  pkgsCross.iphone32                    pkgsCross.sheevaplug
+  pkgsCross.iphone32-simulator          pkgsCross.vc4
+  pkgsCross.iphone64                    pkgsCross.wasi32
+  pkgsCross.iphone64-simulator          pkgsCross.x86_64-embedded
+  pkgsCross.mingw32                     pkgsCross.x86_64-netbsd
+  pkgsCross.mingwW64                    pkgsCross.x86_64-netbsd-llvm
+  pkgsCross.mmix                        pkgsCross.x86_64-unknown-redox
+  pkgsCross.msp430                      
+
+
+From the attribute name it can't always be immediately clear what is the platform.
+
+It's possible to query the platform name using::
+
+  $ nix-instantiate '<nixpkgs>' -A pkgsCross.aarch64-darwin.hostPlatform.config --eval
+  "aarch64-apple-darwin"
+
+.. note:: In case the plaforms hasn't been defined yet, feel free to contribute one
+          by `adding it upstream <https://github.com/NixOS/nixpkgs/blob/master/lib/systems/examples.nix>`_.
+
+
+Cross-compiling for the first time!
+-----------------------------------
+
+To cross-compile a package like `hello <https://www.gnu.org/software/hello/>`_,
+pick the platform target name like ``aarch64-multiplatform`` in our case and run:
+
+.. code:: shell-session 
+
+  $ nix-build '<nixpkgs>' -A pkgsCross.aarch64-multiplatform.hello
+  ...
+  /nix/store/pzi2h0d60nb4ydcl3nn7cbxxdnibw3sy-hello-aarch64-unknown-linux-gnu-2.10
+
+`Search for a package <https://search.nixos.org/packages>`_ attribute name to find the
+one that you're interested in to build.
+
+
+Real world cross-compiling of a Hello World example
+---------------------------------------------------
+ 
+To show off the power of cross-compilation in Nix, let's build our own Hello World program 
+by cross-compiling it as static executables to ``armv6l-unknown-linux-gnueabihf``
+and ``x86_64-w64-mingw32`` (Windows) platforms and run the resulting executable
+with `an emulator <https://en.wikipedia.org/wiki/Emulator>`_.
+
+.. code:: nix 
+
+  { pkgs ? import <nixpkgs> {} 
+  }:
+
+  let
+    # Create a C program that prints Hello World
+    helloWorld = pkgs.writeText "hello.c" ''
+      #include <stdio.h>
+      
+      int main (void)
+      {
+        printf ("Hello, world!\n");
+        return 0;
+      }
+    '';
+
+    # A function that takes host platform packages
+    crossCompileFor = hostPkgs:
+      # Run a simple command with the compiler available
+      hostPkgs.runCommandCC "hello-world-cross-test" {} ''
+        # Wine requires home directory
+        HOME=$PWD
+
+        # Compile our example using the compiler specific to our host platform
+        $CC ${helloWorld} -o hello 
+
+        # Run the compiled program using an emulator
+        # Usually Qemu, but on windows it is Wine
+        ${hostPkgs.stdenv.hostPlatform.emulator hostPkgs.buildPackages} hello > $out
+
+        # print to stdout program stdout
+        cat $out
+      '';
+  in {
+    # Statically compile our example using the two platform hosts
+    rpi = crossCompileFor pkgs.pkgsCross.raspberryPi;
+    windows = crossCompileFor pkgs.pkgsCross.mingwW64;
+  }
+
+If we build this example and print both resulting derivations, we should see "Hello, world!" for each: 
+
+.. code:: shell-session
+
+  $ cat $(nix-build cross-compile.nix)
+  Hello, world!
+  Hello, world!
+
+
+Developer environment with a cross-compiler
+-------------------------------------------
+
+In :ref:`tutorial for declarative reproducible environments <declarative-reproducible-envs>`,
+we've looked at how Nix helps us provide tooling and system libraries for our project.
+
+It's also possible to provide an environment with a compiler configured for cross-compilation!
+
+Given we have a ``shell.nix``:
+
+.. code:: nix
+
+  { nixpkgs ? fetchTarball "https://github.com/NixOS/nixpkgs/archive/bba3474a5798b5a3a87e10102d1a55f19ec3fca5.tar.gz"
+  , pkgs ? (import nixpkgs {}).pkgsCross.aarch64-multiplatform
+  }:
+
+  # pkgs.callPackage is needed due to https://github.com/NixOS/nixpkgs/pull/126844
+  pkgs.callPackage ({ mkShell, zlib, pkg-config }: mkShell {
+    # these tools run on the build platform, but are configure to target the target platform
+    nativeBuildInputs = [ pkg-config file ];
+    # libraries needed for the target platform
+    buildInputs = [ zlib ];
+  }) {}
+
+And ``hello.c``:
+
+.. code:: c 
+
+  #include <stdio.h>
+
+  int main (void)
+  {
+    printf ("Hello, world!\n");
+    return 0;
+  }
+
+We can cross-compile it:
+
+.. code:: shell-session 
+
+  $ nix-shell --run '$CC hello.c -o hello' cross-compile-shell.nix
+
+And confirm it's aarch64:
+
+.. code:: shell-session 
+
+  $ nix-shell --run 'file hello' cross-compile-shell.nix 
+  hello: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /nix/store/733hzlw1hixdm6dfdsb8dlwa2h8fl5qi-glibc-2.31-74-aarch64-unknown-linux-gnu/lib/ld-linux-aarch64.so.1, for GNU/Linux 2.6.32, with debug_info, not stripped
+  
+
+Next steps
+----------
+
+- `Official binary cache <https://cache.nixos.org>`_ doesn't come with binaries
+  for packages that are cross-compiled, so it's important to set up
+  :ref:`a binary cache and CI (GitHub Actions and Cachix) <github-actions>`.
+
+- While many compilers in nixpkgs support cross-compilation,
+  not all of them do.
+
+  On top of that, supporting cross-compilation is not trivial
+  work and due to many possible combinations of what would
+  need to be tested, packages some might not build.
+
+  `A detailed explanation how cross-compilation is implemented in Nix can help fixing those issues <https://nixos.org/manual/nixpkgs/stable/#chap-cross>`_.
+
+- Nix community has a `dedicated Matrix room <https://matrix.to/#/#cross-compiling:nixos.org>`_
+  for help around cross-compiling.
--- a/source/tutorials/cross-compilation.rst
+++ b/source/tutorials/cross-compilation.rst
@@ -12,24 +12,25 @@ or when it's not easily accessible for development.
 The Nix community has world-class support for cross-compilation,
 after years of hard work from our community.
 
-.. [#] Terminology for cross-compilation platforms differs between build systems,
-       we have chosen to follow 
+.. [#] Terminology for cross-compilation platforms differs between build systems.
+       We have chosen to follow 
        `autoconf terminology <https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html>`_.
 
-.. note:: macOS/Darwin is a special case, as not the whole OS is Open Source. 
+.. note:: macOS/Darwin is a special case, as not the whole OS is open-source. 
           It's only possible to cross-compile between ``aarch64-darwin`` and ``x86_64-darwin``.
 
 
 What's a target platform?
 -------------------------
 
-There's actually a third platform named target.
+There's actually a third platform named the target platform.
 
 It matters in cases where you'd like to distribute a compiler binary, 
 as you'd then like to build a compiler on the build platform, compile code on the
 host plaform and run the final executable on the target platform.
 
-Since that's rarely needed, we'll treat target platform the same as the host.
+
+Since that's rarely needed, we'll treat the target platform the same as the host.
 
 
 Determining the host platform config
@@ -46,7 +47,7 @@ The host platform is best determined by running on the host platform:
   aarch64-unknown-linux-gnu
 
 In case that's not possible (when the host platform is not easily accessible
-for development), platform config has to be constructed manually via the following template:
+for development), the platform config has to be constructed manually via the following template:
 
 .. code::
 
@@ -106,9 +107,10 @@ It's possible to list predefined attribute sets via shell completion:
   pkgsCross.msp430                      
 
 
-Cross-compilation package attribute names are made up, so it isn't always clear what is the corresponding platform config.
+Cross-compilation package attribute names are made up, so it isn't always clear 
+what is the corresponding platform config.
 
-Query platform config via:
+It's possible to query the platform config using:
 
   $ nix-instantiate '<nixpkgs>' -A pkgsCross.aarch64-darwin.hostPlatform.config --eval
   "aarch64-apple-darwin"
@@ -121,7 +123,7 @@ Cross-compiling for the first time!
 -----------------------------------
 
 To cross-compile a package like `hello <https://www.gnu.org/software/hello/>`_,
-pick the platform attribute like ``aarch64-multiplatform`` in our case and run:
+pick the platform attribute - ``aarch64-multiplatform`` in our case - and run:
 
 .. code:: shell-session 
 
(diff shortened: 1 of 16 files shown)


source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+
+Nix community has world-class support for cross-compilation,
+after years of hard work from our community.
+
+.. [#] Terminology for cross-compilation platforms differs between build systems,
+       Nix community has chosen to follow 
+       `autoconf terminology <https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html>`_.
+
+.. note:: macOS/Darwin is a special case, as not the whole OS is Open Source. 
+          It's only possible to cross-compile between ``aarch64-darwin`` and ``x86_64-darwin``.
```
sternenseemann, line 20: Technically only from `x86_64-darwin` to `aarch64-darwin`. Not sure if I'd mention this here though, as `aarch64-darwin` is far from mature at this point and cross compilation is not working at all in any channel  currently…

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
+.. _ref-cross-compilation:
+
+Cross-compilation
+=================
+
+Cross-compilation is the act of **compiling code** on the **build platform**
+to the **host platform**, where the compiled **executable runs**. [#]_
```
sternenseemann, line 7: Those platforms also need to differ.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+
+It's needed when the host platform has limited resources (such as CPU)
+or when it's not easily accessible for development.
+
+Nix community has world-class support for cross-compilation,
+after years of hard work from our community.
```
sternenseemann, line 13: Style: community is repeated here a lot which doesn't read to nice. Also I feel like “The Nix community” sounds more correct, but unsure.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+
+
+What's a target platform?
+-------------------------
+
+There's actually a third platform named target.
+
+It matters in cases where you'd like to distribute a compiler binary, 
+as you'd then like to build a compiler on the build platform, compile code on the
+target plaform and run the final executable on the host platform.
```
sternenseemann, line 30: This is wrong: You would run the compiler on the *host* platform and it would generate code for the *target* platform.

This is also why targetPlatform usually doesn't matter: host platform means the same for everything and for everything but build tools and compilers targetPlatform makes zero difference .

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+
+Since that's rarely needed, we'll treat target platform the same as the build.
```
sternenseemann, line 32: Just write that you ignore the platform for this tutorial. The default case of targetPlatform is the same as hostPlatform afaik.

Also interesting to note: On the outermost interface nixpkgs doesn't support `targetPlatform`, it only accepts `localSystem` (~ build) and `crossSystem` (~ host) since cross compiling a cross compiler is a rare use case outside of stdenv bootstrapping.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+
+
+Determining the host platform
+-----------------------------
+
+The build platform is determined automatically by Nix
+as it can just guess it during the configure phase.
```
sternenseemann, line 39: You can also set it by passing `localSystem`.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+There's also no unique identifier for a platform, for example ``unknown`` and 
+``pc`` are interchangeable (hence it's called config.guess).
+
+Some other common examples of platforms:
+
+- aarch64-apple-darwin14
+- aarch64-pc-linux-gnu
+- x86_64-w64-mingw32
+- aarch64-apple-ios
+
```
sternenseemann, line 65: This section feels superfluous if you never mention the `crossSystem` argument which can actually be used to pass a `config` such as we have gotten in this section to use as the host platform for nixpkgs.

Mentioning `crossSystem` is important because it is the most basic case, i. e. `pkgsCross` is just a convenient shortcut for (re-)importing nixpkgs with `crossSystem` set.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+  pkgsCross.iphone64-simulator          pkgsCross.x86_64-embedded
+  pkgsCross.mingw32                     pkgsCross.x86_64-netbsd
+  pkgsCross.mingwW64                    pkgsCross.x86_64-netbsd-llvm
+  pkgsCross.mmix                        pkgsCross.x86_64-unknown-redox
+  pkgsCross.msp430                      
+
+
+From the attribute name it can't always be immediately clear what is the platform.
+
+It's possible to query the platform name using::
```
sternenseemann, line 108: maybe we should stick to terminology of config instead of name in the tutorial?

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+      hostPkgs.runCommandCC "hello-world-cross-test" {} ''
+        # Wine requires home directory
+        HOME=$PWD
+
+        # Compile our example using the compiler specific to our host platform
+        $CC ${helloWorld} -o hello 
+
+        # Run the compiled program using an emulator
+        # Usually Qemu, but on windows it is Wine
+        ${hostPkgs.stdenv.hostPlatform.emulator hostPkgs.buildPackages} hello > $out
```
sternenseemann, line 170: Especially `buildPackages` and maybe how the emulator attribute works should be explained.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+.. code:: shell-session 
+
+  $ nix-shell --run 'file hello' cross-compile-shell.nix 
+  hello: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /nix/store/733hzlw1hixdm6dfdsb8dlwa2h8fl5qi-glibc-2.31-74-aarch64-unknown-linux-gnu/lib/ld-linux-aarch64.so.1, for GNU/Linux 2.6.32, with debug_info, not stripped
+  
+
+Next steps
+----------
+
+- `Official binary cache <https://cache.nixos.org>`_ doesn't come with binaries
```
sternenseemann, line 243: This is technically not true, we have some cache for certain things as part of the `cross-trunk` jobset although it should probably be expanded one of these days…

sternenseemann (changes requested):
Rough first pass.

source/tutorials/cross-compilation.rst
domenkozar: > Technically only from x86_64-darwin to aarch64-darwin.

Is that because we're missing `pkgsCross.x86_64-darwin` or something else?

> Not sure if I'd mention this here though, as aarch64-darwin is far from mature at this point and cross compilation is not working at all in any channel currently…

Since the fix from @alyssais should work, the whole cross-compilation is not well supported but I'd like this tutorial to show that it's entirely feasible with enough maintainers. WDYT?

source/tutorials/cross-compilation.rst
domenkozar: You mean it's not possible for build platform to also be the host platform?

source/tutorials/cross-compilation.rst
sternenseemann: > Is that because we're missing pkgsCross.x86_64-darwin or something else?

I'm not sure as to why exactly, but I can confirm that nixpkgs can't even cross compile from x86_64-darwin to x86_64-darwin due to an evaluation problem very early in `stdenv`.

> the whole cross-compilation is not well supported but I'd like this tutorial to show that it's entirely feasible with enough maintainers

Still, aarch64-linux for example has had more time to mature, accumulate fixes etc. Maybe not mentioning aarch64-darwin saves someone a bit of frustration?

source/tutorials/cross-compilation.rst
domenkozar: I'd like to clarify since it took me quite some time to figure out why target is not important and explain why it can be ignored to save confusion. The user will eventually stumble upon it as they dive into cross-compilation.

source/tutorials/cross-compilation.rst
domenkozar: In what cases is that really needed? In most cases guessing should be sufficient?

source/tutorials/cross-compilation.rst
sternenseemann: I've thought about this yesterday and maybe it would actually *help* the tutorial to explain target completely. This would make it very easy to clear the following things up:

* Why are binutils, pkg-config and cc prefixed? What does the prefix mean? This is quite easy to answer if you explain the relation of the target prefix to the target platform
* What is `buildPackages`? Here having an understanding of target is nice: `buildPackages` usually has build and host platform equivalent to `localSystem`, but its target is `crossSystem` — having an understanding of target automatically explains how we bootstrap our cross package sets

source/tutorials/cross-compilation.rst
domenkozar: How often do we imagine the user provides their own crossSystem? Currently I want to teach that the single truth is the platform identifier and how it one can go between nixpkgs attributes and the identifier.

In case the user needs a new crossSystem, there's a note how to add one to nixpkgs.

Can you think of a use case where that's not enough?

source/tutorials/cross-compilation.rst
domenkozar: I wanted to avoid that to set expectations low and simplify explanation. Maybe something like `comes with almost no binaries`?

source/tutorials/cross-compilation.rst
sternenseemann: That is possible, but that is native compilation, usually. However, you can also do “native” cross compilation (like exherbo) which is also [a goal for nixpkgs I subscribe to](https://github.com/NixOS/nixpkgs/issues/21471).

Since cross compilation is the more generic concept, it is probably hard disambiguate — in a way native compilation is a special case of cross compilation. Following that train of thought you could talk about cross compilation as soon as your compiler has a notion of a target platform.

I think my original suggestion may not be the best one, but the initial paragraph reads to me as if it is missing something, maybe?

source/tutorials/cross-compilation.rst
sternenseemann: The two main use cases would be remote build (if you have a builder with a different platform than your local machine) and testing evaluation of cross-related things by simulating being on another platform.

domenkozar:
cc @zimbatm @zupo (for all the previous help with tutorials!) @ericson2314 @matthewbauer @alyssais (for cross-compilation expertise)

k-bx:
Apologies for a noob question, but I'm getting this:

```
$ nix-shell --run '$CC hello.c -o hello' cross-compile-shell.nix
error: undefined variable 'file' at /home/kb/workspace/nix-crosscompile-test/cross-compile-shell.nix:8:36
```

```
$ cat cross-compile-shell.nix
{ nixpkgs ? fetchTarball "https://github.com/NixOS/nixpkgs/archive/bba3474a5798b5a3a87e10102d1a55f19ec3fca5.tar.gz"
, pkgs ? (import nixpkgs {}).pkgsCross.aarch64-multiplatform
}:

# pkgs.callPackage is needed due to https://github.com/NixOS/nixpkgs/pull/126844
pkgs.callPackage ({ mkShell, zlib, pkg-config }: mkShell {
  # these tools run on the build platform, but are configure to target the target platform
  nativeBuildInputs = [ pkg-config file ];
  # libraries needed for the target platform
  buildInputs = [ zlib ];
}) {}
```

Am I doing something wrong?

nh2:
@k-bx In

```
nativeBuildInputs = [ pkg-config file ];
```

the `pkg-config` comes from the lambda arguments `{ mkShell, zlib, pkg-config }: mkShell`, but `file` doesn't. You need to add it to those arguments so that it is in scope.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,257 @@
...
+Given we have a ``shell.nix``:
+
+.. code:: nix
+
+  { nixpkgs ? fetchTarball "https://github.com/NixOS/nixpkgs/archive/bba3474a5798b5a3a87e10102d1a55f19ec3fca5.tar.gz"
+  , pkgs ? (import nixpkgs {}).pkgsCross.aarch64-multiplatform
+  }:
+
+  # pkgs.callPackage is needed due to https://github.com/NixOS/nixpkgs/pull/126844
+  pkgs.callPackage ({ mkShell, zlib, pkg-config }: mkShell {
```
sternenseemann, line 207: @k-bx probably this is the issue you're encountering.

```suggestion
  pkgs.callPackage ({ mkShell, zlib, file, pkg-config }: mkShell {
```

domenkozar:
> Apologies for a noob question, but I'm getting this:
> 
> ```
> $ nix-shell --run '$CC hello.c -o hello' cross-compile-shell.nix
> error: undefined variable 'file' at /home/kb/workspace/nix-crosscompile-test/cross-compile-shell.nix:8:36
> ```
> 
> ```
> $ cat cross-compile-shell.nix
> { nixpkgs ? fetchTarball "https://github.com/NixOS/nixpkgs/archive/bba3474a5798b5a3a87e10102d1a55f19ec3fca5.tar.gz"
> , pkgs ? (import nixpkgs {}).pkgsCross.aarch64-multiplatform
> }:
> 
> # pkgs.callPackage is needed due to https://github.com/NixOS/nixpkgs/pull/126844
> pkgs.callPackage ({ mkShell, zlib, pkg-config }: mkShell {
>   # these tools run on the build platform, but are configure to target the target platform
>   nativeBuildInputs = [ pkg-config file ];
>   # libraries needed for the target platform
>   buildInputs = [ zlib ];
> }) {}
> ```
> 
> Am I doing something wrong?

Not at all, I added `file` and didn't do the full test again. Should be fixed now!

k-bx:
Aha! It worked now, thanks. Sorry for hijacking the thread, but I'm trying to build hello.c for my Raspberry Pi Zero. What should I put instead of `aarch64-multiplatform` in that cross-compile-shell.nix exactly?

domenkozar:
> Aha! It worked now, thanks. Sorry for hijacking the thread, but I'm trying to build hello.c for my Raspberry Pi Zero. What should I put instead of `aarch64-multiplatform` in that cross-compile-shell.nix exactly?
> 
> UPDATE: ok, figured it's just `raspberryPi` hehe :)

https://b5839745.nix-dev.pages.dev/tutorials/cross-compilation#determining-the-host-platform is supposed to teach how to determine platform config.

https://b5839745.nix-dev.pages.dev/tutorials/cross-compilation#choosing-the-host-platform-with-nix is supposed to teach how to pick the right attribute with that platform config in mind.

Let me know where my explaining went south :)

k-bx:
@domenkozar oh, I see. I was confused by the terminology, assuming that the host is the machine that does the building, and the "target" (or some other term) is my raspberryPi, since that command runs nix to determine the architecture, and I would assume you wouldn't need to install Nix on raspberry pi. Now I see that you would.

k-bx:
Again apologies for spamming, just trying to get that raspberry pi zero hello world to work.

Installing nix is not an option (sorry, there is no binary distribution of Nix for your platform) to determine the host platform. I would assume it's armv6*. If I were to try `armv6l-unknown-linux-gnueabihf` mentioned in that tutorial, I'd get an error of that attribute missing:

```
$ nix-shell --run '$CC hello.c -o hello' cross-compile-shell.nix
error: attribute 'armv6l-unknown-linux-gnueabihf' missing, at /home/kb/workspace/nix-crosscompile-test/cross-compile-shell.nix:2:10
(use '--show-trace' to show detailed location information)

$ cat ./cross-compile-shell.nix
{ nixpkgs ? fetchTarball "https://github.com/NixOS/nixpkgs/archive/bba3474a5798b5a3a87e10102d1a55f19ec3fca5.tar.gz"
, pkgs ? (import nixpkgs {}).pkgsCross.armv6l-unknown-linux-gnueabihf
}:

# pkgs.callPackage is needed due to https://github.com/NixOS/nixpkgs/pull/126844
pkgs.callPackage ({ mkShell, zlib, pkg-config, file }: mkShell {
  # these tools run on the build platform, but are configure to target the target platform
  nativeBuildInputs = [ pkg-config file ];
  # libraries needed for the target platform
  buildInputs = [ zlib ];
}) {}
```

Am I not reading the tutorial correctly? Again, sorry for noob stuff.

domenkozar:
@k-bx thanks, this is really useful! I've added a sentence to hint how to get platform config if it's not possible to use Nix:

> If you can't install Nix, find a way to run ``config.guess`` (usually comes with
 the autoconf package) from the OS you're able to run on the host platform.


`pkgsCross` attributes are made-up names that map into the platform config. I've clarified that in the tutorial.

Unfortunately, there's no simple way to show us all attribute name -> platform config pairs.

source/tutorials/cross-compilation.rst
domenkozar: I'd like to mention that then in #157 tutorial (and consider developing nixpkgs out of scope here).

source/tutorials/cross-compilation.rst
domenkozar: Agreed! I don't think I know enough to really explain those, unfortunately.

source/tutorials/cross-compilation.rst
domenkozar: How about a second paragraph that says:

> Cross-compilation is generalized from native compilation, where host and build platforms are the same.

k-bx:
Aha! Interestingly, `config.guess` is not in PATH, but found it in `/usr/share/misc/config.guess`. It's `armv6l-unknown-linux-gnueabihf` indeed. Due to https://github.com/NixOS/nixpkgs/blob/master/lib/systems/examples.nix#L38 the value `raspberryPi` was indeed the right one. However, if I copy the resulting executable on my rpi0, I get this:

```
pi@raspberrypi:~ $ ls -la
total 88
drwxr-xr-x 10 pi   pi    4096 Jun 15 12:45 .
drwxr-xr-x  3 root root  4096 Mar  4 22:47 ..
-rwxr-xr-x  1 pi   pi   11600 Jun 15 15:46 hello
pi@raspberrypi:~ $ ./hello 
-bash: ./hello: No such file or directory
pi@raspberrypi:~ $ ldd ./hello 
        /usr/lib/arm-linux-gnueabihf/libarmmem-${PLATFORM}.so => /usr/lib/arm-linux-gnueabihf/libarmmem-v6l.so (0xb6f86000)
        libc.so.6 => /usr/lib/arm-linux-gnueabihf/libc.so.6 (0xb6e38000)
        /nix/store/bs0yybprj2jdz4gk4g2738pnfbraaxh2-glibc-2.31-74-armv6l-unknown-linux-gnueabihf/lib/ld-linux-armhf.so.3 => /lib/ld-linux-armhf.so.3 (0xb6f99000)
```

That's why I thought `raspberryPi` wasn't the right value initially. Do you know what could be the reason for this?

> Unfortunately, there's no simple way to show us all attribute name -> platform config pairs.

That is indeed one of the pain-points of Nix: you see the final spell, but it's hard to inspect what other keys and values are available. I hope to dive into Nix near time and see if I can help improve this situation.

sternenseemann:
@k-bx The `no such file or directory` is the dynamic linker telling you that it is not finding the libraries linked against. To fix that you can either a) compile statically by passing `-static` to `$CC` or b) copy all dependencies of your executable. Finding this out for a binary built with `nix-shell` is a bit annoying, but if you have a proper derivation you could use `nix-store --requisites <realised store path>`.

source/tutorials/cross-compilation.rst
sternenseemann: How about something like “We have extremely limited binary-cache for cross compilation, so it is very possible that you have to compile the respective toolchain yourself”? This would also clarify the risk (compiling binutils and gcc).

source/tutorials/cross-compilation.rst
sternenseemann: Well the immediate use case would of course be (as it has come up here) that you can pass a triple to `crossSystem` and it will figure things out (or try to).

source/tutorials/cross-compilation.rst
sternenseemann: Why not!

k-bx:
@sternenseemann thank you! May I just mention that adding `static` will produce an error:

```
/nix/store/s8j8w9kik99ffxcgwz3bdpk71ijr3qj7-armv6l-unknown-linux-gnueabihf-binutils-2.31.1/bin/armv6l-unknown-linux-gnueabihf-ld: cannot find -lc
```

I've tried adding glibc, that didn't help. `libc` is not found when accepted as lambda argument, not sure what exactly to do next (again, probably due to lack of linker knowledge/experience).

sternenseemann:
@k-bx Oh yeah, you need to use the `static` output for `glibc`, i. e. `glibc.static` or alternatively musl, but in that case you probably want to switch stdenv (static compilation is also a lot easier if you actually use a static `stdenv` of course…).

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,262 @@
+.. _ref-cross-compilation:
+
+Cross-compilation
+=================
+
+Cross-compilation is the act of **compiling code** on the **build platform**
+to the **host platform**, where the compiled **executable runs**. [#]_
```
Ericson2314, line 7:
```suggestion
When compiling code, we can distinguish between the **build platform**, where the executable
is *built*, and the **host platform**, where the compiled executable *runs*. [#]_
**Native compilation** is the special case where those two platforms are the same.
**Cross compilation** is the general case where those two platforms are not.
```

Also I wouldn't hyphenate "cross-compilation". It's just a noun phrase. Only "cross-compiled" the adjective as in "let's use this cross-compiled library" needs it.

k-bx:
@sternenseemann thank you! Changing `buildInputs` argument to `glibc.static` indeed helped!

I guess I finally have a starting point of how to get from here to a point where I can cross-compile my Rust project that is using some native libraries :)

sternenseemann:
@k-bx Yeah, this is something we should take into consideration as well. There is some special casing to be done for rust possibly, I'm not quite sure.

In general the way would be:

* Package your project normally as a nix derivation, take extra care to properly distinguish between `nativeBuildInputs` and `buildInputs` ([see manual](https://nixos.org/manual/nixpkgs/unstable/#variables-specifying-dependencies)) and add it via an [overlay](https://nixos.org/manual/nixpkgs/unstable/#chap-overlays) to nixpkgs
* Then compiling `pkgsCross.<platform>.my-package` should just work (if everything is supported)
    * For shared executables (the default case) you need to copy the entire closure (i. e. what `nix-store --requisites` prints) while preserving its location (i. e. everything must be under `/nix/store`)
    * static executables can just be copied as is, you can build your derivation statically using `pkgsCross.<platform>.pkgsStatic.my-package`

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,262 @@
...
+- x86_64-w64-mingw32
+- aarch64-apple-ios
+
+
+Choosing the host platform with Nix
+-----------------------------------
+
+Nixpkgs comes with a set of predefined host platforms applied to all packages.
+
+It's possible to list predefined attribute sets via shell completion:
```
pingiun, line 76: Shell completion doesn't always work fully, so it might be nice to give the command to list these as well. I'm often scratching my head when thinking about how to search for things in nixpkgs so it's good to be reminded often I feel

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,262 @@
...
+or when it's not easily accessible for development.
+
+The Nix community has world-class support for cross-compilation,
+after years of hard work from our community.
+
+.. [#] Terminology for cross-compilation platforms differs between build systems.
+       We have chosen to follow 
+       `autoconf terminology <https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html>`_.
+
+.. note:: macOS/Darwin is a special case, as not the whole OS is open-source.
```
pingiun, line 19: I think for added clarity, it would be nice to mention it's a special case to use as a build platform.

domenkozar: It's it a special case for the host platform?

If macOS is a build platform, you can still cross compile to linux as a target for example.

source/tutorials/cross-compilation.rst
domenkozar: Good point! I should maybe show how to use a reply here.

source/tutorials/cross-compilation.rst
domenkozar: How about a disclaimer that we've just added support for it?

source/tutorials/cross-compilation.rst
domenkozar: Good point :laughing:

source/tutorials/cross-compilation.rst
sternenseemann: `lib/systems/examples.nix` is not a scary file I think it is usable as a list of platforms.

domenkozar: I'd like to avoid pointing the user to a file (through github) as a reference to what attributes `pkgsCross` has as it's hard to explain the connection between the two :(

source/tutorials/cross-compilation.rst
domenkozar: I took the inspiration from nixpkgs manual: https://nixos.org/manual/nixpkgs/stable/#chap-cross

domenkozar:
Thanks everyone for the feedback!

I've included most of it in the last commit, please take another look!

source/tutorials/cross-compilation.rst
sternenseemann: I mean the alternative is telling them to run

```console
$ nix-instantiate --eval --strict -E 'builtins.attrNames (import <nixpkgs/lib>).systems.examples'
```

pretty much…

source/tutorials/cross-compilation.rst
Ericson2314: Yeah macos -> true FOSS is fine. I might also move this minor caviat down, as none of the text above talks about what platforms are supported yet or anything like that.

source/tutorials/cross-compilation.rst
```diff
@@ -0,0 +1,284 @@
...
+  pkgsCross.mingwW64                    pkgsCross.x86_64-netbsd-llvm
+  pkgsCross.mmix                        pkgsCross.x86_64-unknown-redox
+  pkgsCross.msp430                      
+
+Cross compilation package attribute names are made up, so it isn't always clear 
+what is the corresponding platform config.
+
+It's possible to query the platform config using:
+
+  $ nix-instantiate '<nixpkgs>' -A pkgsCross.aarch64-multiplatform.hostPlatform.config --eval
```
Ericson2314, line 127:
```suggestion
  $ nix-instantiate '<nixpkgs>' -A pkgsCross.aarch64-multiplatform.stdenv.hostPlatform.config --eval
```
We would like to deprecate `hostPlatform` at the top level.

source/tutorials/cross-compilation.rst
pingiun: Oh I was confused here, wouldn't it also be the case with windows then? As windows is also closed source largely?

Ericson2314: It's possible to make macOS cross work better; it's just no one did yet. With WIndows it has been done.

source/tutorials/cross-compilation.rst
domenkozar: I've added a disclaimer and move this further down.

source/tutorials/cross-compilation.rst
domenkozar: Went with `nix repl` as that's a good thing to teach anyway (although it adds more things to introduce here, it's for the better).

domenkozar:
> @sternenseemann thank you! Changing `buildInputs` argument to `glibc.static` indeed helped!
> 
> I guess I finally have a starting point of how to get from here to a point where I can cross-compile my Rust project that is using some native libraries :)

The example for developer environment uses now a cross compiler that emits static binaries. Thanks for bringing this up!

domenkozar:
I'm going to merge this in a few hours to prevent perfect becoming the enemy of the good.

Let me know if I missed some feedback and happy to address it further, but I might convert it into issues to keep things going.

PR #164, opened 2021-06-24
cross-compilation: fix platform terminology in shell example

sternenseemann:
* `nativeBuildInputs` is roughly equivalent to taking a package from `buildPackages` which
  means that *its* host platform is the build platform of the derivation (so we can execute it during
  the build) and its target platform is the host platform of the derivation (so e. g. compilers produce
  binaries that can be executed where we want them to)
* `buildInputs` are needed on the *host* platform of the derivation
* What would the target platform be in this case? It would only be relevant if we were to build a
  cross-compiler: Then we'd need runtime libraries for the target platform, for example.


sternenseemann:

domenkozar:
Good catch!

PR #184, opened 2021-09-30
add short hacking tutorial for nixpkgs with docker

Mic92:
This example is intentionally a bit more verbose as docker might be used
by upstream authors to test package in nixpkgs without having to install
nix.


Mic92:

PR #197, opened 2021-11-13
Add instructions to install nix with a package manager.

loreb:
Prompted by https://github.com/nix-dot-dev/nix.dev/issues/194

This came out longer than I hoped; feel free to criticize it to death (or merge and then modify to your pleasure).

### DIFF
--- a/source/tutorials/install-nix.rst
+++ b/source/tutorials/install-nix.rst
@@ -25,7 +25,7 @@ for instance on debian based distributions:
 
 .. code:: bash
 
-   sudo apt install nix
+   sudo apt install nix-bin nix-setup-systemd
 
 .. note::
 
--- a/source/tutorials/install-nix.rst
+++ b/source/tutorials/install-nix.rst
@@ -3,8 +3,8 @@
 Install Nix
 ===========
 
-Linux
------
+Linux (manual installation)
+---------------------------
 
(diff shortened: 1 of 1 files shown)


loreb:

loreb:
Slightly rewritten because I forgot a step (it didn't show in my history because I ran it as root).

source/tutorials/install-nix.rst
```diff
@@ -17,6 +17,37 @@ Install Nix on via the recommended `multi-user installation <https://nixos.org/m
...
 
+Linux using your package manager
+--------------------------------
+
+If Nix is available in your favorite distribution, you can alternatively use its package manager;
+for instance on debian based distributions:
+
+.. code:: bash
+
+   sudo apt install nix
```
SuperSandro2000, line 28:
```suggestion
   sudo apt install nix-bin nix-setup-systemd
```

loreb:
@fricklerhandwerk np, shit happens

If you feel like rewriting the thing yourself be my guest, it's basically:
1. install using your package manager (because maintainers made sure it works)
2. may or may not need to enable the nix service depending on distribution (some distros will *not* start a service unless you personally say so)
3. will likely need to set a few environment variables by sourcing some file, hopefully /etc/profile.d/nix-something.sh, maybe /usr/share/examples/nix/whatever.sh or similar (eg debian)
4. subscribe to a channel and `nix-channel --update`
5. double check if there are distro-specific instructions (eg debian&arch require users to be part of a group)

PR #208, opened 2021-12-09
Fix image name in dd command of the raspberrypi tutorial

nikstur:
The dd command used an older image instead of the newer image downloaded earlier in the tutorial


nikstur:

PR #209, opened 2021-12-09
Update Python env

zupo:
Update to the latest package versions.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,11 +12,24 @@
           inherit system;
           overlays = [ poetry2nix.overlay ];
         };
+        poetryOverrides = self: super: {
+
+          # Refs https://github.com/nix-community/poetry2nix/issues/218#issuecomment-981615612
+          typing-extensions = super.typing-extensions.overridePythonAttrs (
+            old: {
+              buildInputs = (old.buildInputs or [ ]) ++ [ self.flit-core ];
+            }
+          );
+        };
       in rec {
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -12,11 +12,22 @@
           inherit system;
           overlays = [ poetry2nix.overlay ];
         };
+        poetryOverrides = self: super: {
+          typing-extensions = super.typing-extensions.overridePythonAttrs (
```
zupo, line 16: Refs https://github.com/nix-community/poetry2nix/issues/218#issuecomment-981615612

pyproject.toml
```diff
@@ -7,12 +7,14 @@ license = "CC BY-SA"
...
 python = "^3.8"
-sphinx = "4.3.1"
-livereload = "2.6.3"
-sphinx-book-theme = "0.1.7"
-sphinx-copybutton = "0.4.0"
+
+sphinx = "*"
+livereload = "*"
+sphinx-book-theme = "*"
+sphinx-copybutton = "*"
```
zupo, line 14: Hard pins don't work as then `poetry update` cannot do anything.

Some people might argue that we should have `"^4.3.1"` instead of `"*"`, because "that carries more information", but over time, this information gets more and more incorrect. Incorrect documentation is worse than no documentation.

flake.nix
domenkozar: Let's add that as a comment

PR #210, opened 2021-12-09
Initial support for MyST

zupo:
For now, we can have both .rst and .md files, so we can do the convertion to Markdown file-by-file. In this PR I converted only a single file, just to show how it is done. No changes in styling.

Before:
Screenshot 2021-12-09 at 12 45 22

After:
Screenshot 2021-12-09 at 12 45 30

The rationale to convert to Markdown:
* less frictions for newcomers to submit patches because more people are comfortable with Markdown than ReST,
* official Nix(OS) manuals are Markdown too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,10 +8,14 @@ license = "CC BY-SA"
 [tool.poetry.dependencies]
 python = "^3.8"
 
-sphinx = "*"
 livereload = "*"
+myst-parser = {version = "*", extras = ["linkify"]}
+sphinx = "*"
 sphinx-book-theme = "*"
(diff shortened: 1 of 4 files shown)


zupo:
Before

Screenshot 2021-12-09 at 10 53 26

pyproject.toml
```diff
@@ -8,11 +8,15 @@ license = "CC BY-SA"
 [tool.poetry.dependencies]
 python = "^3.8"
 
-sphinx = "*"
```
zupo, line 11: Alphabetizing packages, sorry for OCDing.

PR #221, opened 2021-12-15
Clarify explanations on cross compilation

fricklerhandwerk:
From my impression that most software developers have no idea about cross compiling with `nix` before coming here (including myself), I think it is important to carefully introduce terminology and the mechanism used in `nixpkgs`.

I tried to make shorter, clearer sentences, and give some context on how and why things are the way they are.


fricklerhandwerk:

domenkozar:
Thanks!

Issue #219, opened 2021-12-13
The example in "Tutorials > Ad hoc developer environments >Towards reproducibility" fails


rldp:
> The following is a fully reproducible example and something that different colleagues with different machines, for example, could share.

```
$ nix-shell --pure -p git -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/82b5f87fcc710a99c47c5ffe441589807a8202af.tar.gz
```

On nixos-version = 22.05pre338278.b0bf5f888d3 (Quokka), I get

```
error: anonymous function at /nix/store/y2fh997ssq4i68di86q60mwfviqjvh28-source/pkgs/top-level/default.nix:20:1 called with unexpected argument 'inNixShell'

       at /nix/store/y2fh997ssq4i68di86q60mwfviqjvh28-source/pkgs/top-level/impure.nix:84:1:

           83|
           84| import ./. (builtins.removeAttrs args [ "system" "platform" ] // {
             | ^
           85|   inherit config overlays crossSystem crossOverlays;
```

domenkozar:
Fixed in 5e0ba5f66b80401ba51c43c6230f6196d1356462

PR #240, opened 2022-03-16
Update archive urls

Michael-J-Ward:
I was still hitting #219 even with `nix --veersion` of 2.7.0

Updating the nixpkgs archive url to the latest fixes it.

The final comment in this thread which I believe to be the cause says no fix will be backported
- https://github.com/NixOS/nix/pull/5543


Michael-J-Ward:

domenkozar:
recheck

PR #255, opened 2022-05-02
Fixing ./result/bin/myapp.py to ./result/bin/myapp

geekbrother:
After the `nix-build` command the resulting app bin is `myapp` not `myapp.py`.

Tested on: 
macOS 12.3.1
nix-build (Nix) 2.8.0

Commands history:
```
1. nix-build
2. ls result/bin/
myapp
```


geekbrother:

PR #256, opened 2022-05-04
Declarative reproducible developer environments

worktheclock:
RITZA language edit.

### DIFF
--- a/source/tutorials/first-steps/declarative-and-reproducible-developer-environments.md
+++ b/source/tutorials/first-steps/declarative-and-reproducible-developer-environments.md
@@ -64,7 +64,8 @@ Now try `which htop` to check where the `htop` executable is stored in the files
 
 ## Customizing your developer environment
 
-Given the following `shell.nix`:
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Sorry it took so long for a review.

This is really good, thank you very much!

PS: Please do the CLA thing.

proofconstruction:
I guess we have a setting to automatically request review when force pushing or something?

Anyway, I rebased* the original commit onto master and made several mostly-small changes to the prose as well as larger changes to the way some things are discussed, in particular:
- more information in the introduction describing why this tutorial differs from the ad-hoc developer environments,
- better explaining why the declarative approach is (more) useful,
- better motivating the `shell.nix` customization, and
- using more precise language to describe what's happening

* I hope I did this correctly; I can't tell whether the first commit by the original author was preserved, or if I accidentally included some of my changes to the file that got committed there. Apologies if so.

fricklerhandwerk (approved):
Looks great, awesome work!

PR #261, opened 2022-07-07
add contribution guide

fricklerhandwerk:
just adding what is already there, up to discussion.

    
based on Nix documentation team [discussion on 2022-07-07](https://discourse.nixos.org/t/2022-07-07-documentation-team-meeting-notes-4/20198)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,107 @@
 # Contributing to nix.dev
 
+nix.dev is a community effort to collect, create, and maintain world-class learning resources for Nix.
+
+We strongly encourage everyone interested to contribute by asking informed questions or directly proposing changes.
+
+Make a [pull request] if you want to introduce an incremental change.
+Note our [considerations on licensing and attribution](#licensing-and-attribution).
+
+Open an [issue] if you want to clarify something not evident from what is provided in this repository, or if you want to discuss a significant change before starting to work on it.
+
+[issues]: https://github.com/NixOS/nix.dev/issues
+[pull request]: https://github.com/NixOS/nix.dev/pulls
+
+## You want to learn and use Nix?
+
+This project would not be possible without you.
+Try to use it as your primary resource, however incomplete it may appear.
+
+We ask you to liberally open issues and document all problems and questions that arise.
+Please also state your learning goals and the paths you have taken so far.
+
+Sharing your first-hand experience is invaluable to better guide our efforts, and will immediately help improve these guides for yourself and everyone else.
+
+## You have experience teaching Nix?
+
+You will probably have observed where learners get stuck most often, and which typical needs and questions they have.
+You may have your own written notes for classes, trainings, or presentations.
+
+Please share your experience and help us inform the structure and detailed contents the guides.
+It would be great if you could contribute examples, wordings, or illustrations that proved helpful to your students.
+
+## You are a domain expert using Nix?
+
+If you are proficient in applying Nix to a domain-specific problem, and want to share your expertise on best practices, please check the table of contents.
+Does existing material on your subject meet your standards?
+How could we improve it?
+Is there a popular application of Nix' capabilities not yet covered?
+We would be glad to incorporate your insights.
+
+# Guidelines
+
+Here are values and practical guidelines to go by when contributing.
+
+## Values
+
+### Be kind
+
+Adapted from [Contributor Covenant] and [The Carpentries Code of Conduct]:
+
+- Use welcoming and inclusive language
+- Show empathy and respect towards other people
+- Be respectful of different viewpoints and experiences
+- Give and gracefully accept constructive criticism
+- Focus on what is best for the community
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -30,3 +132,49 @@ Use culturally neutral language:
...
+Many web services offer permalinks.
+Examples:
+- [GitHub URLs to specific commits][github-permalink]
+- [Wikipedia URLs to specific page versions][wikipedia-permalink]
+- [Internet Archive "Save Page Now" for persisting web pages][internet-archive-save]
+
+[permalink]: https://en.m.wikipedia.org/wiki/Permalink
+[github-permalink]: https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files
+[wikipedia-permalink]: https://en.m.wikipedia.org/wiki/Wikipedia:Linking_to_Wikipedia#Permanent_links_to_old_versions_of_pages
+[internet-archive-save]: https://web.archive.org/save
```
SuperSandro2000, line 165: Lets take this as an example:

```suggestion
- [GitHub URLs to specific commits](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files)
- [Wikipedia URLs to specific page versions](https://en.m.wikipedia.org/wiki/Wikipedia:Linking_to_Wikipedia#Permanent_links_to_old_versions_of_pages)
- [Internet Archive "Save Page Now" for persisting web pages](https://web.archive.org/save)
```
This is what I would write normally and I would prefer here. It is shorter and easier to know what links where.

CONTRIBUTING.md
```diff
@@ -30,3 +132,49 @@ Use culturally neutral language:
...
+
+### Markdown
+
+Write one sentence per line.
+This makes long sentences immediately visible, and makes it easier to manage changes.
```
SuperSandro2000, line 170: I don't think that is a good rule because for short sentences this adds excessive new lines and very long sentences to little.

CONTRIBUTING.md
```diff
@@ -30,3 +132,49 @@ Use culturally neutral language:
...
+
+The rule is unambiguous and does not require tooling support to be applied easily.
+[Here is a discussion of different line wrapping styles.][markdown-wrapping-styles]
+
+Use [reference links][reference-link] to keep the plain text readable.
+Collect links at the end of each section, which are delimited by headings.
+
+[markdown-wrapping-styles]: https://web.archive.org/web/20220519121408/https://mtsknn.fi/blog/4-1-wrapping-styles-for-markdown-prose-and-code-comments/
+[reference-link]: https://github.github.com/gfm/#reference-link
```
fricklerhandwerk, line 179: > IMO this is only useful if the same link is used multiple times or is very long

Most links are very long. Definitely long enough to break reading flow in the source.

> enforcing it for every link is annoying

We don't have to enforce it, and apply some discretion when reviewing. It's only a guide after all. If we wanted to strictly enforce it for some reason we could add a formatter to CI.

@SuperSandro2000 In any case, the point of the proposal is to strike a balance when improving convenience for authors, reviewers, and maintainers (or anyone who wants to change the source).

For instance, we don't have to name the references if they are not reused, and keeping them close to their use site should not impose too much of a burden on maintenance. In fact, I think it is *easier* to update links when they are on a separate line, and especially easier to inspect the resulting diff.

Does that sound more reasonable? 

```suggestion
The rule is unambiguous and does not require tooling support to be applied easily.
[Here is a discussion of different line wrapping styles.]

Use [reference links] to keep the plain text readable.
Collect links close to their references:
at the end of each paragraph (delimited by an empty lines) or section (delimited by headings).

[Here is a discussion of different line wrapping styles.]: https://web.archive.org/web/20220519121408/https://mtsknn.fi/blog/4-1-wrapping-styles-for-markdown-prose-and-code-comments/
[reference links]: https://github.github.com/gfm/#reference-link
```

README.md
```diff
@@ -15,4 +15,6 @@ Run `./live` and open a browser at <http://localhost:5500>.
 
 As you make changes your browser should auto-reload within a few seconds.
 
+For contents and style see [contribution guide][CONTRIBUTING.md].
```
fricklerhandwerk, line 18:
```suggestion
For contents and style see [contribution guide](CONTRIBUTING.md).
```

README.md
```diff
@@ -15,4 +15,6 @@ Run `./live` and open a browser at <http://localhost:5500>.
 
 As you make changes your browser should auto-reload within a few seconds.
 
+For contents and style see [contribution guide](CONTRIBUTING.md).
```
fricklerhandwerk, line 18:
```suggestion
For contents and style see the [contribution guide](CONTRIBUTING.md).
```

CONTRIBUTING.md
infinisil: Too short and too long sentences are discouraged anyways for writing in general, so I think this is a very good idea. It encourages people towards medium-length sentences.

fricklerhandwerk:
- @fricklerhandwerk would like to merge something at all, so we can refer to it in reviews
    - let's figure out what is least controversial and discuss that here
    - will split out the rest into different PRs and discuss async separately
        - @infinisil is ok, unless any topic turns out to be more controversial
            - @fricklerhandwerk okay, what about the licensing section?
                - @domenkozar it's good that it documents the purpose
- @infinisil like the sentence breaking proposal, it's reasonable

PR #263, opened 2022-07-11
codify terminology

fricklerhandwerk:
we observe confusion around terminology.
- newcomers find it hard to distinguish all the things that are called Nix
- Nix is a very generic term that is hard to search for online

this proposal is to introduce just enough structure to clearly distinguish concepts, and give names that are not wrong, without invalidating existing material. therefore it is strictly oriented around existing conventions.

it is *not* intended as an attempt to radically rename ecosystem components, even if that may be desirable for better readability and discoverability.

closes https://github.com/NixOS/nix.dev/issues/275

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -183,3 +183,36 @@ Collect links at the end of each section, which are delimited by headings.
 
 [Here is a discussion of different line wrapping styles.]: https://web.archive.org/web/20220519121408/https://mtsknn.fi/blog/4-1-wrapping-styles-for-markdown-prose-and-code-comments/
 [reference links]: https://github.github.com/gfm/#reference-link
+
+### Nomenclature
+
+To avoid confusion around all the things called Nix, always use the following terms with capitalization as given:
+
+- Nix ecosystem
+
+  All of the below.
+
+  - Nix package manager
+
+    Command line tool.
+
+  - Nix language
+
+      Programming langauge to declare packages and configurations for the Nix package manager.
+
+    - Nix expression
+
+      Expression written in Nix language.
+
+    - Nix file
+
+      File containing Nix expression, usally ends in `.nix`.
+
+  - Nix package collection (`nixpkgs`)
+
+    Read `nixpkgs` as "Nix packages".
+
+  - NixOS
+
+    Linux distribution based on Nix package manager and `nixpkgs`.
+    Read "Nix Oh Es".
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -186,7 +186,10 @@ Collect links at the end of each section, which are delimited by headings.
 
 ### Terminology
 
-To avoid confusion around all the things called Nix, always use the following terms with capitalization as given:
+To avoid confusion around all the things called Nix, always use the following terms with capitalization as given.
+
+Terms in brackets (`[  ]`) are optional.
+Terms in parentheses (`(  )`) are abbreviations.
 
 - Nix ecosystem
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -197,7 +197,7 @@ Terms in parentheses (`(  )`) are abbreviations.
 
   - Nix package manager
 
-    Command line tool.
+    - `nix` [command [line interface]] (Nix CLI)
 
   - Nix language
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,7 +184,7 @@ Collect links at the end of each section, which are delimited by headings.
 [Here is a discussion of different line wrapping styles.]: https://web.archive.org/web/20220519121408/https://mtsknn.fi/blog/4-1-wrapping-styles-for-markdown-prose-and-code-comments/
 [reference links]: https://github.github.com/gfm/#reference-link
 
-### Nomenclature
+### Terminology
 
 To avoid confusion around all the things called Nix, always use the following terms with capitalization as given:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -215,7 +215,9 @@ Terms in parentheses (`(  )`) are abbreviations.
 
     Read `nixpkgs` as "Nix packages".
 
-  - NixOS
+  - NixOS [Linux distribution]
 
     Linux distribution based on Nix package manager and `nixpkgs`.
     Read /nɪks oʊ ɛs/ ("Nix Oh Es").
+
+    - NixOS module collection (NixOS modules)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -200,7 +200,7 @@ To avoid confusion around all the things called Nix, always use the following te
 
   - Nix expression
 
-    Expression written in Nix language.
+    Expression written in the Nix language.
 
   - Nix file
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -188,36 +188,32 @@ Collect links at the end of each section, which are delimited by headings.
 
 To avoid confusion around all the things called Nix, always use the following terms with capitalization as given.
 
-Terms in brackets (`[  ]`) are optional.
-Terms in parentheses (`(  )`) are abbreviations.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -14,3 +14,36 @@ Address the reader with "you" when necessary.
...
+      File containing Nix expression, usally ends in `.nix`.
+
+  - Nix package collection (`nixpkgs`)
+
+    Read `nixpkgs` as "Nix packages".
+
+  - NixOS
+
+    Linux distribution based on Nix package manager and `nixpkgs`.
+    Read "Nix Oh Es".
```
fricklerhandwerk, line 48:
```suggestion
    Read /nɪks oʊ ɛs/ ("Nix Oh Es").
```

CONTRIBUTING.md
```diff
@@ -30,3 +30,41 @@ Use culturally neutral language:
...
+Terms in brackets (`[  ]`) are optional.
+Terms in parentheses (`(  )`) are abbreviations.
+
+- Nix ecosystem
+
+  All of the below.
+
+  - Nix package manager
+
+    - `nix` [command [line interface]] (Nix CLI)
```
blaggacao, line 46: Because of the many things that the Nix CLI is (and becoming): I would at least swap these two.

The _Nix package manager_ would be reflected in a subset of subcommands.

Maybe like so (actual names are no firm suggestion / only an examplification):

- `nix` ...
  - Implements: Nix Package Manager
  - Implements: Nix Language Evaluator
  - Implements: Nix REPL
  - Implements: Nix Flake Interface
  - ...

CONTRIBUTING.md
```diff
@@ -30,3 +30,41 @@ Use culturally neutral language:
...
   What you may consider well-known may be entirely obscure and distracting to people from different backgrounds.
 
+### Terminology
+
+To avoid confusion around all the things called Nix, always use the following terms with capitalization as given.
+
+Terms in brackets (`[  ]`) are optional.
+Terms in parentheses (`(  )`) are abbreviations.
+
+- Nix ecosystem
```
blaggacao, line 40: Nix Ecosystem.

In order to clearly mark these as tokenize d terms, they should be all Title Case.

Same for the rest...

CONTRIBUTING.md
```diff
@@ -30,3 +30,41 @@ Use culturally neutral language:
...
+  - Nix package collection (`nixpkgs`)
+
+    Read `nixpkgs` as "Nix packages".
+
+  - NixOS [Linux distribution]
+
+    Linux distribution based on Nix package manager and `nixpkgs`.
+    Read /nɪks oʊ ɛs/ ("Nix Oh Es").
+
+    - NixOS module collection (NixOS modules)
```
blaggacao, line 69: I would introduce:

- The `NixOS/nixpkgs`-Repository implements:
  - Nix Package Collection &mdash; `./pkgs`
  - NixOS Module Collection &mdash; under `./nixos`
  - Nix Library Collection &mdash; `./lib` (also inndividually as smaller `nix-community/nixpkgs.lib`)

blaggacao (commented):
Since Nix is very fundamentally a source-based distribution mechanism, in my opinion, it makes sense to acknowledge that fact by relating glossary terms to sources. In this light, suggestions above.

CONTRIBUTING.md
blaggacao: I would explicitly call-out the misnomer. Not addressing this upfront makes people question their own best judgement at some point which is unpleasant.

fricklerhandwerk:
@rapenne-s I agree, but I strongly doubt we can pull of a large renaming without doing a lot of damage. This proposal is to provide just enough structure to make sense of things, in a backwards compatible way, even if it all sounds redundant.

CONTRIBUTING.md
fricklerhandwerk: > The `NixOS/nixpkgs`-Repository implements:

I follow your thought but I think your proposal gets it backwards. The goal right now is to give things a sufficiently distinguishable name which is not misleading. It should not matter where these things live (it may change), but what they are.

CONTRIBUTING.md
```diff
@@ -30,3 +30,41 @@ Use culturally neutral language:
...
+
+  All of the below.
+
+  - Nix package manager
+
+    - `nix` [command [line interface]] (Nix CLI)
+
+  - Nix language
+
+      Programming langauge to declare packages and configurations for the Nix package manager.
```
bew, line 50: Typo
```suggestion
      Programming language to declare packages and configurations for the Nix package manager.
```

CONTRIBUTING.md
fricklerhandwerk: The "ecosystem" or "package collection" should be part of the proper name, this is why it's not capitalized.

Tangent: This is just my taste, but it think capitalization is largely pointless. Yes, it can help with readability and distinguishing proper names, but it also has a cost of abiding to fairly arbitrary rules just to keep consistency. This is just how writing systems of European descent happen to work. Many languages, [in fact most of them](https://en.m.wikipedia.org/wiki/Letter_case#Bicameral_script), don't have a notion of letter cases, and it works just fine.

CONTRIBUTING.md
```diff
@@ -30,3 +30,41 @@ Use culturally neutral language:
...
+
+      Expression written in Nix language.
+
+    - Nix file
+
+      File containing Nix expression, usally ends in `.nix`.
+
+  - Nix package collection (`nixpkgs`)
+
+    Read `nixpkgs` as "Nix packages".
```
edolstra, line 62: The short name is Nixpkgs (capitalized).

CONTRIBUTING.md
```diff
@@ -183,3 +183,37 @@ Collect links at the end of each section, which are delimited by headings.
...
+
+  Read /nɪks/ ("Niks").
+
+- Nix language
+
+    Programming language to declare packages and configurations for Nix.
+
+  - Nix expression
+
+    Expression written in Nix language.
```
fricklerhandwerk, line 203: @matta suggested that making Nix Language proper name by capitalization would avoid having to write "the Nix language" all the time.

```suggestion
- Nix Language

    Programming language to declare packages and configurations for Nix.

  - Nix expression

    Expression written in Nix Language.
```

Otherwise it would have to be

```suggestion
- Nix language

    Programming language to declare packages and configurations for Nix.

  - Nix expression

    Expression written in the Nix language.
```

matta: I was credited with the suggestion for "written in Nix Language" but please don't take that as indication I personally think it is a good idea.  :-)

I do think "written in Nix **L**anguage" (bold to emphasize capitalization) in this context is, possibly, _grammatically_ correct, this is true only if we are to assume that "Nix **L**anguage" is the proper name of the language and refer to it that way consistently. However, "Nix **L**anguage" is an odd and awkward name. I don't think this is a good idea for multiple reasons:

1. Very few, if any, languages of any sort, computer or otherwise, are named in English with multiple words, none that I know of include "language" in the proper name.
2. "written in Nix" or "written in the Nix language" are conventional, clear, with obvious meaning. There is no reason to use unconventional or unusual wording.

fricklerhandwerk:
@edolstra @domenkozar @Mic92 @infinisil There seem to be no more objections, therefore I would merge this end of next week.

infinisil (approved):
Looking good!

CONTRIBUTING.md
```diff
@@ -183,3 +183,37 @@ Collect links at the end of each section, which are delimited by headings.
...
+
+  - Nix expression
+
+    Expression written in the Nix language.
+
+  - Nix file
+
+    File (`.nix`) containing a Nix expression.
+
+- Nixpkgs
```
infinisil, line 209: A slight ambiguity I just noticed: This doesn't say whether Nixpkgs should be capitalized when occurring within a sentence.

fricklerhandwerk: It says to capitalize as written, though.

infinisil: Ahh, I should've read the whole section 😅

PR #265, opened 2022-07-11
add outline to contribution guide

fricklerhandwerk:
we need an easily accessible place to guide contributors and their ideas. here is one way to do it.

@spacekookie @infinisil

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,16 @@
 
 nix.dev is a community effort to collect, create, and maintain world-class learning resources for Nix.
 
+It is supposed to fit between first impressions on the [Nix web site](https://nixos.org) and the reference manuals for [Nix][nix manual] ([source][nix manual src]), [Nixpkgs][nixpkgs manual] ([source][nixpkgs manual src]), and [NixOS][nixos manual] ([source][nixos manual src]).
+See the [vision](#vision) and [textual outline](#outline) for an idea of a desired end state.
+
+[nix manual]: https://nixos.org/manual/nix
+[nix manual src]: https://github.com/NixOS/nix/tree/master/doc/manual
+[nixpkgs manual]: https://nixos.org/manual/nixpkgs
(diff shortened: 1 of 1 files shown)


bew (commented):
Looks good to me :+1: 

(didn't approve, as I'm a bit new around here, maybe others will have comments)

fricklerhandwerk:
In addition to my usability tests, recent scans through Discourse revealed that beginners consistently ask the following questions:

- How is Nix different from Docker?
- What are flakes good for, should you learn to use them?
- When should you use home-manager?
- How do you pick a specific package version?
- How to set up Vim?

Answers to these questions should be one of the first documentation items to find. We mostly have them already, partially, somewhere.

fricklerhandwerk:
I had at least three conversations with Nix beginners who found this kind of overview helpful. Therefore I would like to get something out there at all that we can work with and pass around.

@domenkozar I really liked your idea of linking the outline directly to articles xor pre-made issues to measure demand. But I don't want to get into these details yet, because it would require setting up a Mermaid renderer and all kinds of things to really make good use of it.

What I'm most concerned about right now is display size. Maybe we would need multiple diagrams to navigate between.

I've recently seen the talk [Let’s Tell a Story: Scenario-Based Documentation](https://www.youtube.com/watch?v=3hneCmLYWjc), where the speaker shows examples of a complex task they were documenting, and on every large step there was a roadmap diagram with the current step highlighted.

For example, the command-line oriented section could be introduced by expanding only the relevant block:

```mermaid
flowchart TB
  subgraph first-steps[run any program]
    direction BT
    run[ad hoc environments]
    findpkgs[find packages]
    scripts[reproducible scripts]
    gc[garbage collection]
  end
  install[install Nix] --> first-steps
  first-steps --> lang[Nix language] & imperative[imperative package management]
  lang --> declarative[declarative environments]
  declarative --> configuration[declarative configuration management] & pkgs[declarative package management]
  configuration --> dev[software development lifecycle]
  pkgs --> dev
```

Eventually we may even auto-generate this from contents, but this seems like a lot of setup for not a lot of material at the moment.

PR #266, opened 2022-07-14
No longer needed, shell.nix works on Darwin

zupo:
Refs https://github.com/nix-dot-dev/nix.dev/issues/250


zupo:
Don't merge yet, please, might have found one problem.

PR #267, opened 2022-07-15
Nix language tutorial

fricklerhandwerk:
based on @tazjin's [tazjin/nix-1p][1] and @zimbatm's NixCon 2019 talk [Reading The Nix Language][2]

[1]: https://github.com/tazjin/nix-1p
[2]: https://youtu.be/hbJkMl631FE?t=1572

As discussed in https://github.com/NixOS/nix.dev/issues/288

To render a preview:

```
git clone https://github.com/fricklerhandwerk/nix.dev
cd nix.dev
git checkout nix-language-tutorial
nix-shell
./live
```


source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
+
+Examples:
+
+    # write a Nix expression to a file
+    echo 123 > nix_path_file
+
+    # set `NIX_PATH` to point a name to that file
+    env NIX_PATH=name_in_nix_path=nix_path_file nix repl
+
+    nix repl> import <nix_path_file>
```
jneem, line 120: `import <name_in_nix_path>`?

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
+    123
+
+Nix simply reads and parses the `NIX_PATH` environment variable.
+Entries are separated by colons (`:`).
+
+Note: The value of a named path depends on external system state. Therefore it is recommended not to use them.
```
jneem, line 126: As a new nix user, I find comments like this a bit scary because I see examples (especially ones involving `<nixpkgs>`) all over the place. Are these examples doing it wrong? Or should the "recommended not to use them" advice be qualified?

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
+- *declarative*
+
+  It has no notion of executing sequential steps.
+  Dependencies between operations are established only through data.
+  Everything in Nix is an expression that results in a single value.
+  Every Nix file (`.nix`) contains a single expression.
+
+- *purely functional*
+
+  Pure means: Nix does not change the value of declarations during computation – there are no variables, only names for immutable values.
```
infinisil, line 312: > there are no variables

Sounds very weird, because there are, but they aren't variable. Maybe reword to "There are no mutable variables"

infinisil (changes requested):

source/tutorials/nix-language.md
fricklerhandwerk:
```suggestion
    nix repl> import <name_in_nix_path>
```

source/tutorials/nix-language.md
fricklerhandwerk: > Are these examples doing it wrong?

IMHO: Yes. At best it's a sloppy short-hand. See the [pinning tutorial](https://github.com/NixOS/nix.dev/blob/21ed44040211f92edc179f1017ab96a74a4ebdd7/source/tutorials/towards-reproducibility-pinning-nixpkgs.md?plain=1#L5-L18) for reasoning.

```suggestion
Note:
The value of a named path depends on external system state.
That means, the same Nix expression can produce different results, which may be very confusing.
Therefore {ref}`we recommend to avoid named paths<Towards reproducibility: pinning Nixpkgs>`.
```

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
+Recursive sets allow elements to refer to attributes within the set.
+They are constructed with the `rec` keyword.
+
+    nix-repl> rec { a = 1; b = a + a; }
+    { a = 1; b = 2; }
+
+Note: This construct
+- allows infinite recursion, such as `rec { x = x; }.x`
+- can quickly become hard to understand
+- does not lend itself to static analysis.
```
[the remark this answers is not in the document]
fricklerhandwerk:
```suggestion
```

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
```
fricklerhandwerk, line 160:
```suggestion
- can quickly become hard to understand.
```

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,324 @@
...
+Nix functions can explicitly take an attribute set as argument.
+
+    {a, b}: a + b
+
+This is equivalent to
+
+    x: x.a + a.b
+
+The argument defines the exact attributes that have to be in that set.
+Leaving out or passing additional attributes is an error.
```
fricklerhandwerk, line 226:
```suggestion
While this is similar to

    x: x.a + x.b

only keyword arguments define the exact attributes that have to be in that set.
Leaving out or passing additional attributes is an error.
```

Bu we may as well leave out the similarity, as it introduces more detail than strictly necessary. I thought it may help readers transition from what may be structure of the argument in a function declaration.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,324 @@
...
+Applying that to another argument yields the inner body `x + y`, which can now be fully evaluated.
+
+    nix-repl> let
+              f = x: y: x + y;
+              in
+              f 1 2
+    3
+
+### Keyword arguments
+
```
fricklerhandwerk, line 216:
```suggestion

Also known as “attribute set argument”.
```

@infinisil This sounds more correct, but I haven't seen that one yet. Is it common? I would just pose it as the main way to call it, but it sounds clunky to me.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,324 @@
...
+              in
+              f { a = 1; b = 2; c = 3; }
+    3
+
+### Named keyword arguments
+
+Also known as “@ syntax” or “‘at’ syntax”.
+
+    {a, b, ...}@args: a + b + args.x
+
```
fricklerhandwerk, line 281:
```suggestion
    {a, b, ...}@args: a + b + args.x

or

    args@{a, b, ...}: a + b + args.x
```

source/tutorials/nix-language.md
fricklerhandwerk: "variables that are not variable" is an oxymoron. Therefore there are no variables in the Nix language.

infinisil: Even if variables never change during execution, they are still called variables in the sense that their value can vary between executions. See https://en.wikipedia.org/wiki/Variable_(mathematics) or https://www.merriam-webster.com/dictionary/variable

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
+    nix-repl> let x = 1; in { a = x; b = x + x; }
+    { a = 1; b = 2; }
+
+## Functions
+
+The Nix language is Turing-complete.
+It gives you the power to tame complexity and the freedom to create overwhelming amounts of it.
+
+Therefore, it is recommended to use functions sparingly and instead think more in terms of data, because data is usually easier to understand.
```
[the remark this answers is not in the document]
fricklerhandwerk: Okay, this is really opinionated, and we may want to discuss this on the docs team meeting, but here's my stance:

Independently, @edolstra and @ndmitchell aptly referred to what people tend create in Turing-complete configuration languages as "monstrosities". In my opinion, that is what people tend to create in *any* programming language if they don't care enough about long-term maintenance (which often has good reasons) and still scale up the code base (which is the curse of success). Both Eelco and Neil opted for restricting the language for Buck v2 and Nix respectively, by introducing a data-oriented layer at the very top. I argue we could instead encourage people to write code more carefully from the start. That is the entire premise of the article - JSON with functions where it helps.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
+# Reading the Nix language
+
+The Nix language is used to declare packages and configurations for the Nix package manager.
+
+You will quickly encounter Nix language expressions that may look very complicated.
+Yet, the language has only few basic concepts that can be combined arbitrarily.
+Use this guide to learn reading code in Nix language and understand its structure.
+
+Imagine the Nix language as JSON with functions.
```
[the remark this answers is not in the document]
fricklerhandwerk: This is a deliberate oversimplification to give . @edolstra just recently said [the very same thing](https://youtu.be/h8hWX_aGGDc?t=955) and I only wrote "functions" not to scare off people who are not familiar with lambda calculus.

To what extent is it misleading? What would you suggest instead as a bird's-eye-view on the topic?

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,325 @@
...
+Its purpose is to define structured data, and functions help with conveniently producing more complex data.
+All of the language constructs can be composed arbitrarily.
+
+### Attribute sets
+
+The primary data structure is the attribute set, a collection of key-value pairs.
```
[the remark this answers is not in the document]
fricklerhandwerk: > Calling it the primary data structure feels weird

I wanted to say it's the most important one, because everything revolves around attrsets. See [JSON+lambdas](https://github.com/NixOS/nix.dev/pull/267/files#r932169165). 

> Also data structure is a bit off

Could you clarify why, just so I understand?

(In general I agree that using the most precise terms is more correct. Yet, I don't think its also more better, as the goal of these tutorials is to pick up developers, who are not necessarily all too familiar with PLT, from where they are). 

> better distinction between values and expressions. This document here mixes these two concepts fairly arbitrarily.

Could you please provide concrete suggestions?

domenkozar:
See https://github.com/NixOS/nix.dev/pull/286 for my WIP

fricklerhandwerk:
- @fricklerhandwerk every attempt to write an introduction so far degenerated to what the manual should look like
    - idea: focus on only on the parts people will see most of the time
    - @infinisil omit obscure details such as dynamic attribute names
    - @domenkozar people will build simple configs, we should focus on most common work loads
      - @fricklerhandwerk trying to find out what the important 80% are, cannot do it on my own
        - @infinisil maybe just look at common Nix files and explain their components?
            - @fricklerhandwerk the thesis and manual already do this, but it's bumpy
                - the [tutorial-style part](https://nixos.org/manual/nix/stable/expressions/simple-expression.html) is derived from the thesis
                    - a long time ago was the best we had
                        - only one of my ten study subjects found it at all, and had severe trouble
                    - it should just go away
            - @domenkozar we should instead use concrete examples to test learning success
    - @infinisil saying Nix is JSON+functions is not enough
        - still need things like attribute acccess operators
    - @Mic92 it took me a while to understand that NixOS modules are functions
        - what should people learn from this, write packages or NixOS modules?
            - @fricklerhandwerk only *read* Nix code without fear
            - @infinisil there should be a specific tutorial for writing packages (and modules)
        - wonder if an online REPL would be a good way to help people learn
            - used one when learning programming with Ruby
            - @fricklerhandwerk if it shows to be helpful with actual learners we should invest the effort
            - definitely helped me with understanding how NixOS works
    - @Mic92 introduce unusual constructs that are used commonly in Nix
      - @fricklerhandwerk let us collect some
          - function application without parantheses
          - attribute set arguments
          - `import`
          - `inherit`
          - derivations
              - @infinisil this would introduce a lot of complexity and necessarily pull in `nixpkgs`
                  - @fricklerhandwerk no, we just have to say that it is one of the two side effects without going into detail
                      - so far we have a huge gap between pure language tutorials and what to do with them, so we need to introduce at least the principle of side effects
                  - @domenkozar just show the practically usable bits, no one actually writes a `derivation` directly
          - string contexts
          - space-separated lists
              - with space-separated function application
                - @fricklerhandwerk the [reference-style part](https://nixos.org/manual/nix/stable/expressions/expression-language.html) has all of these things
                    - it is already good in terms of contents, but fairly hard to read due to lacking structure and dense wording
    - @olaf start from the use case instead of listing all possibilities
        - attribute sets are a core feature, the material we build with, and
            - explain how to use them
        - @fricklerhandwerk agree, should start from expected knowledge (such as JSON) and build up sophistication guided by need
        - @Mic92 best books I learned from gave me exercises to test what I learned
            - @infinisil the draft has examples
                - @fricklerhandwerk I would like them to be interactive eventually
                    - @domenkozar we can add that later

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,799 @@
...
+pkgs.mkShell {
+  # ...
+}
+```
+
+Explanation:
+
+This expression is a function that takes an attribute set as an argument.
+If the argument has the attribute `pkgs`, it will be used in the function body.
+Otherwise, the default value of importing from the search path `<nixpkgs>` and calling the resulting function with an empty attribute set will be used.
```
bew, line 684: ⚠️ You didn't explain `import` yet, what it does, and how to use it

fricklerhandwerk: There will be a section on import, this PR is still a draft.

source/tutorials/nix-language.md
matta: I think a focus on whether Nix does or does not have "actual variables" is a confusing distraction when describing what pure functional programming means.

Colloquial use of the word "variable", at least among English speakers, applies in an intuitive way to functional programming. Nix has these in function arguments and let bindings, both of which allow a program to cause one name to take on different values. This is the core concept behind what a "variable" is.

See also https://nixos.org/manual/nix/stable/expressions/language-constructs.html#let-expressions which says "A let-expression allows you to define local variables for an expression."

See also https://nixos.org/guides/nix-pills/basics-of-language.html#idm140737320512656 which shows you how to use a let expression to shadow variables, effectively making one name refer to multiple values by introducing a new scope.

There is also https://en.wikipedia.org/wiki/Variable_(computer_science) which says "In pure functional programming, variables are bound to expressions and keep a single value during their entire lifetime due to the requirements of referential transparency."

I would suggest that the description of "pure" avoid the problem entirely by not focusing on a "variables that are not variable" oxymoron. Instead, focus on the idea that functions always return the same when passed the same arguments, and that computed values cannot be modified.  Or something like that...

fricklerhandwerk: I will remove this section entirely. I had been influenced by the manual, [which says](https://web.archive.org/web/20220716000213/https://nixos.org/manual/nix/stable/expressions/expression-language.html)

> for instance, there is no variable assignment

The manual should be the reference on what the Nix language is and does, so I moved my attempt to describe it more clearly there: https://github.com/NixOS/nix/pull/6907

The [NixOS Wiki on Nix language replicates this](https://web.archive.org/web/20220812152834/https://nixos.wiki/wiki/Overview_of_the_Nix_Language), and that redundancy should IMHO go away, and instead we should always refer to the manual. Same for this tutorial.

fricklerhandwerk:
@domenkozar @infinisil this is ready for review.

domenkozar:
From the CI:

```
Warning, treated as error:
/home/runner/work/nix.dev/nix.dev/source/tutorials/nix-language.md:301:broken link: https://nixos.org/manual/nix/stable/expressions/language-values.html#lists 
```

fricklerhandwerk:
This is fixed already, just re-run CI.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1600 @@
...
+  f = {a ? 0, b ? 0}: a + b
+in
+f { } # empty attribute set
+```
+
+    0
+
+### Additional attributes
+
+Additional attributes are allowed with an ellipsis (`...`):
```
djacu, line 1054: Just read through this and this section was the only thing to give me pause. I already knew this but I think bringing a little more attention to the use of ellipsis would be good.

```suggestion
Additional attributes are allowed with an ellipsis (`...`).

Unlike the previous counter-example, passing additional arguments will not be an error.

Example:
```

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1600 @@
...
+
+    {a, b, ...}: a + b
+
+Example:
```
djacu, line 1058: Just read through this and this section was the only thing to give me pause. I already knew this but I think bringing a little more attention to the use of ellipsis would be good.

```suggestion
Unlike the previous counter-example, passing additional arguments will not be an error.

Example:
```

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+```nix
+1 + 2
+```
+
+    3
+
+#### Interactive evaluation
+
+Use [`nix repl`][nix-repl] to evaluate Nix expressions interactively (by typing them on the command line):
```
infinisil, line 107: You need to mention `:p` for strict evaluation somewhere, it's needed for many examples

fricklerhandwerk: Not strictly needed, but yes, sometimes otherwise the output will not be the same.

fricklerhandwerk:
> It seems like anchor changes from lists to list

A general solution to changing anchors: https://github.com/NixOS/nix/pull/7006

fricklerhandwerk:
- @fricklerhandwerk how to proceed with reviews?
    - it's a very long article
    - already tested successfully with two absolute beginners
    - just merge and fix forward?
- @infinisil would still like to review it in full
    - @domenkozar me too, will try to find time next week
    - is there a rendered version
        - @domenkozar we had netlify, but it appears to have broken when moving to the NixOS GitHub org
            - will fix this
            - if you move the repo to a different org you have to start over with Cloudflare
            - unfortunately nix.dev will be down until it is set up again (a few minutes)
- @fricklerhandwerk we can add more notes in `<details>` tag to adjust level of detail for readers
    - @infinisil have to strike a balance what is learning material and what should go to reference
    - can also split out details into separate articles if necessary

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+It *does not* explain all Nix language features in detail.
+See the [Nix manual][manual-language] for a full language reference.
+
+[manual-language]: https://nixos.org/manual/nix/stable/language/index.html
+
+### What do you need?
+
+- Familiarity with software development
+- Familiarity with Unix shell, to read command line examples <!-- TODO: link to yet-to-be instructions on "how to read command line examples" -->
+- [](./install-nix), to run the examples
```
zuzuleinen, line 61: There is a missing text here

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+:::{important}
+A build task in Nix is called *derivation*.
+:::
+
+The Nix language has only few basic constructs which can be combined arbitrarily:
+
+- primitive data types
+
+  as basic building blocks
```
infinisil, line 192: I guess by this you mean "Expressions representing their values directly", so e.g. `null`, `1`, `"str"`, `{ x = 10; }` but not `1 + 1`, `let x = 10; in x`, etc.

However, this category then also includes functions, e.g. `x: x` or `{ f = x: x; }`, which also evaluate to their values directly. But functions are in the next category already

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+- compound data types and functions
+
+  to produce and transform complex data
```
infinisil, line 196: I'd say this category should be everything else, maybe call it "Expressions that aren't fully evaluated", including things like `(x: x) 1`, `let x = 10; in x`, etc. Evaluating all of these gives a value that's different from the expression

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+- name assignment
+
+  to manipulate complex data as units
```
infinisil, line 200: And then this category falls away. Notably with this category there would be another ambiguity, because functions also introduce names (function arguments).

infinisil (changes requested):
Only reviewed the first two top-level sections for now, I'll do the rest later. I hope you don't mind the many comments, let me know if you think I'm too pedantic :)

source/tutorials/nix-language.md
fricklerhandwerk: Grammatically, you are probably right.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+The Nix language is used to declare packages and configurations for [Nix][nix-manual].
+
+It is a domain-specific, purely functional, lazily evaluated, dynamically typed programming language.
+
+Notable uses of the Nix language are:
+
+- [Nixpkgs][nixpkgs]
+
+  The largest, most up-to-date software distribution in the world, and written in Nix language.
```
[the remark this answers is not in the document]
fricklerhandwerk: I tried to leave out articles where they appeared not strictly necessary. This is up for debate of course, but if we can have fewer words that would be great.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+## Overview
+
+Using the Nix language in practice entails multiple things:
+
+- language: syntax and semantics
+- function libraries: `builtins` and `pkgs.lib`
+- developer tools: testing, debugging, linting and formatting, ...
+- generic build mechanisms: `stdenv`, trivial builders, ...
+- composition and configuration mechanisms: `override`, `overrideAttrs`, overlays, `callPackage`, ...
+- ecosystem-specific packaging mechanisms: `buildGoModule`, `buildPythonApplication`, ...
```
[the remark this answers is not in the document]
fricklerhandwerk: Yes. Why would one capitalize this? They are not sentences, not even close.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+  The Nix language is largely unsurprizing to experienced functional programmers.
+
+  Read through the guide to get accustomed to the syntax and language features.
+
+We recommend running all examples and playing with them to validate your assumptions and test what you have learned.
+
+### How to run the examples?
+
+- Every valid piece of Nix language code is a *Nix expression*.
+- Nix expressions can contain other Nix expressions, that is, they can be nested.
```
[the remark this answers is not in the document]
fricklerhandwerk: That was indeed a fix after running it against beginners, clarifying "Nix file = Nix expression". I originally took that note from @tazjin's nix-1p, because I would have found that information very valuable early on. I can clearly remember wondering what kind of thing was in a Nix file, and someone telling me that it evaluates to exactly one thing would have helped a lot.

But, as you say, it made people wonder if expressions can be nested, and indeed they can, so that appeared like a reasonable addition.

So while I agree that it may be too much information at this point, I'm inclined to listen beginners here rather than experts.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+- Evaluating a Nix expression produces a single value.
```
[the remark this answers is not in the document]
fricklerhandwerk: Good catch!

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: Ah yeah makes sense

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+- Every *Nix file* (`.nix`) contains a single Nix expression.
+
+:::{note}
+To *evaluate* means to transform an expression according to the language rules until no further simplification is possible.
+:::
+
+All examples in this guide are valid Nix expressions.
+Each one is followed by the expected evaluation result.
```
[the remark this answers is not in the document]
fricklerhandwerk: Right, that seems to be an artifact before counter-examples came into play. Do you have a suggestion to fix this?

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+```console
+nix repl
+```
+
+    Welcome to Nix 2.5.1. Type :? for help.
+
+```console
+nix-repl> 1 + 2
+```
```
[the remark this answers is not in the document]
fricklerhandwerk: This will be distiguished visually with https://github.com/NixOS/nix.dev/pull/320

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+cat file.nix
+```
+
+    1 + 2
+
+The second command runs `nix-instantiate` with the `--eval` option on `file.nix`, which reads the file and evaluates the contained Nix expression.
+The resulting value is printed as output.
+
+`--eval` is required to evaluate the file and do nothing else.
+If `--eval` is omitted, `nix-instantiate` expects the expression in the given file to evaluate to a special data type called *Derivation*, which we will come back to at the end of this guide in [](build-tasks).
```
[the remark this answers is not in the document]
fricklerhandwerk: I'd argue it is special, because it's the whole point of Nix. It deserves special attention. Even if we should decide not to call it special here, I would like to emphasize that it is a value of a certain data type as opposed to the other three things a {store d,d,`d`}erivation could be.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+```
+
+    3
+:::
+
+[nix-instantiate]: https://nixos.org/manual/nix/stable/command-ref/nix-instantiate.html
+
+## Reading the Nix language without fear
+
+You will quickly encounter Nix language expressions that may look very complicated.
```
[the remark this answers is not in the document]
fricklerhandwerk: I didn't want to judge if they *actually* appear complicated to the reader, but I'm pretty sure one will encounter such things.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+As with any programming language, the required amount of Nix language code closely matches the complexity of the problem it is supposed to solve, and reflects how well the problem – and its solution – is understood.
+
+Building software is a complex undertaking, and Nix both exposes and allows managing this complexity with the Nix language.
+
+The purpose of the Nix language is to create *build tasks*: precise descriptions of how contents of existing files are used to derive new files.
```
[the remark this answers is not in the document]
fricklerhandwerk: This would be missing the fact that Nix makes that complexity visible in the first place. (Thanks to @AlexeiDrake for pointing this out, the two paragraphs are almost verbatim his.)

It would also be missing the purpose of the language, which motivates why one would even bother using or having it in the first place, and also builds an arch (of dramatic suspense 🙄) to the end of the tutorial, where we finally get to those build tasks we got promised.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+- name assignment
+
+  to manipulate complex data as units
+
+In addition it allows three side effects:
+
+  - reading files as Nix expressions
+
+    to enable code reuse
```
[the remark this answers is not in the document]
fricklerhandwerk: I agree and I was not sure whether to include this, but it motivates to show `import` and gives it a natural place to live, which was otherwise hard to find.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+  - reading files as *build inputs*
+
+    to capture what build tasks will operate on
```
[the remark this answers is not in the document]
fricklerhandwerk: Sure it reads files, otherwise Nix couldn't compute a store path.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+  - writing files as *build tasks*
+
+    to keep them for later execution, the *build*
```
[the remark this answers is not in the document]
fricklerhandwerk: This is all true, but then again how do we otherwise structure this introduction and still show the relevant mechanisms?

Given the [purpose of the language](https://github.com/notifications?query=backport#discussion_r967533619) it would be pointless to evaluate it without capturing the store derivations. Nix language is not a general-purpose language, there is no point using it outside of that problem domain.

source/tutorials/nix-language.md
fricklerhandwerk: Yes, functions are also primitives, but as with the other things you pointed out this is purely for helping orient the reader by providing consistent structure in both overview and sections.

source/tutorials/nix-language.md
fricklerhandwerk: I understand what you mean but I really doubt it would help the cause here. For me that kind of wording crosses the boundary of being so technically correct as to confuse unsuspecting beginners.

Sure, being not technically correct will confuse unsuspecting PLT experts. But you're not the intended audience. I'm already taking some liberty with that entire narrative, comparing Nix to JSON, building complex data and all that. These things are just to support that narrative, to create a mental model that is deliberately simplified to focus on the essentials.

I'd love to make it technically correct and still focused and readable, but I'm out of fantasy right now and would love to see a full suggestion how to get there. Feel free to make a PR against this PR, or propose such a significant change after merging.

fricklerhandwerk:
@infinisil Thanks for the review so far.

I would like to quote @thufschmitt here:

> Once agreed on a general direction – merge first, bikeshed later.

I love your rigorous feedback regarding being correct about PLT terminology. I'm mostly deliberately and only sometimes involuntarily a bit loose on some concepts to form a coherent narrative. If you can propose something that is more precise and doesn't break the overall structure it would be just great. Changing pieces in isolation would break that symmetry between overview and chapters, but we absolutely want to be predictable for readers and keep a pattern.

So I'd prefer we either merge it as is and work on a PR to get things right later; or you make a larger suggestion (or standalone PR) against this PR, which, if I understand correctly, may entail reordering sections or completely rewriting some passages to match the new structure.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+
+Example:
+
+```nix
+x: y: x + y
+```
+
+    <LAMBDA>
+
+The above function takes one argument and returns a function `y: x + y` with `x` set to the value of the argument.
```
matta, line 942: I am unfamiliar with the phrase "chaining arguments". The first thing I do when faced with this is try to research further, but I found it hard to find any references for what is meant by this phrase, whether it is a phrase I should be using in communication with other people, etc. I did web searches to try to figure out if this phrase is specific to Nix, functional programming, etc., and came up short. I did find results that refer to chaining *functions* but not arguments. This suggests to me that perhaps a tutorial should not use this phrase (unless it is used elsewhere in Nix reference material).

If the goal is to introduce Nix using familiar and clear concepts, perhaps title this section "Handling multiple arguments" and then explain in the section content that Nix has no such thing but a similar thing can be accomplished by (insert common functional programming term here).

In particular I don't quite know how to understand `x: y: x + y` at first glance. Beginning by explaining that the expression is equivalent to `x: (y: y + x)` would have helped me. It is obvious in retrospect, but I am still learning! :-) For some reason the parentheses and placing the `x` last also helps me understand. Or, phrasing the sentence "The above function takes one argument `x` and returns a function `y: y + x` with `x` set to the value of the argument.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+- [Nix manual: Nix language][manual-language] - Nix language reference
+- [Nix manual: Built-in Functions][nix-builtins] - Nix language built-in functions
+- [Nixpkgs manual: Functions reference][nixpkgs-functions] - Nixpkgs function library
+- [Nixpkgs manual: Fetchers][nixpkgs-fetchers] - Nixpkgs fetcher library
+
+## Next steps
+
+To get things done:
+
+- [](declarative-reproducible-envs) – create reproducible shell environments from a Nix file
```
matta, line 1541: FYI `[](declarative-reproducible-envs)` renders as the empty string in the document.

matta (commented):
@fricklerhandwerk this is great! At last I found time to read this through from start to finish and I feel like I understand the Nix language better now. The detailed explanations you give of how to read more complex Nix expressions at the end were *very* satisfying.

A thing I wish for now is a "Practical guide to debugging Nix" tutorial. ;-)

As I am just a random person on the Internet, please feel free to commit without changes. All of my comments are intended to be "for your information" only.

domenkozar:
@fricklerhandwerk if you rebase on top of master, we should (hopefully) get a preview link.

source/tutorials/nix-language.md
fricklerhandwerk: > chaining arguments

Yes, you're right. I kind of made that up. I thought about calling it currying, on the one hand this is already quite technical jargon, and on the other hand it's not even right, as currying is defined as transforming higher-arity functions to nested unary functions.

> "Handling multiple arguments" and then explain in the section content that Nix has no such thing

That's what I'll do.

> Beginning by explaining that the expression is equivalent to x: (y: y + x) would have helped me.

I'll definitely keep arguments in the same order, but I'll add the parenthesized version, it's a good idea.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1602 @@
...
+Example:
+
+```nix
+(x: x + 1) 1
+```
+
+    2
+
+:::{note}
+Since function and operand are separated by white space, sometimes parantheses (`( )`) are required to distinguish expressions.
```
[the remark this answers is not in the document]
fricklerhandwerk: Since the goal of the tutorial is to read the language, not to write it, I'm convinced it doesn't matter much where semicolons go. But I see what you mean, and will think about how to address these issues.

To answer your immediate question: Semicolons are used for three things.
- delimit attributes in a attribute sets
- delimit `let` declarations
- separate source and sink in `with`

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: (The link you gave doesn't work)

Derivations should be described as the _output_ of a Nix evaluation (it's a somewhat special type of Nix value), not a side effect. So Nix evaluation happens, a value gets produced, which is then used by other parts of Nix to turn them into something `nix-instantiate` (without `--eval`) creates the derivation in the store, and coupling that with `nix-build` also builds it. But the Nix language is not involved past evaluation.

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: Ahh you're right, my mistake. (though I don't think it's a relevant side effect)

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: I agree with @matta that not having a `the` here feels very unnatural

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: I'm not talking about "Evaluating a Nix expression produces a single value.", that makes a lot of sense and I can see how that helps beginners. But the fact that expressions can be nested feels out of place here. If we need to point out more explicitly that expressions can contain other expressions, this might be better further down somewhere.

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: How about "This guide contains examples of Nix expressions, each followed by the expected evaluation result"

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: Fair enough, I mainly think "data type" is not right. How about

```suggestion
If `--eval` is omitted, `nix-instantiate` expects the expression in the given file to evaluate to a special value called *Derivation*, which we will come back to at the end of this guide in [](build-tasks).
```

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: The _may_ at the beginning should already take care of this, the user may encounter expressions that look complicated [to them]. "looking complicated" is already subjective. Anyways not a big concern, the original sentence is just a bit off imo

source/tutorials/nix-language.md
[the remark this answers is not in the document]
infinisil: Side effects are kind of a bad thing in purely functional languages, so you'd want to avoid them whenever possible. So I really don't think something that's not even an actual side effect should be advertised as one. There has to be a better place to mention `import`

infinisil:
> > Once agreed on a general direction – merge first, bikeshed later.

While there's some parts of my review which are bike-shedding, the thing about advertising `import` as a side-effect is definitely not imo. That's just wrong and should be corrected before being merged. Yes it might require restructuring the tutorial, but that's what reviews are for. If I have time and motivation for this I'm gonna help out myself (I don't right now).

As a side note, I have a feeling that the Nix community in general tends to label everything "bike-shedding" even when it's not. Having discussions about details is not bike-shedding when those details are important.

source/tutorials/nix-language.md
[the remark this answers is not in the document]
fricklerhandwerk: Yeah you're right. I'll have to take some time to think about restructuring according to the larger points in the review comments.

source/tutorials/nix-language.md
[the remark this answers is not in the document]
fricklerhandwerk: It is a relevant side effect because files may change between evaluations as opposed to say fetched resources that have a predeclared hash.

source/tutorials/nix-language.md
fricklerhandwerk: > I guess by this you mean "Expressions representing their values directly"

Actually no, I'm not even thinking much in terms of expressions and values in this context, but instead want to address what I suppose are semantics: what do you *do* with these language objects?

In that sense (which is borrowed from Abelson & Sussman's SICP) functions are not primitives, even if the expressions declaring them do not evaluate further. They are means of combination. Bindings in function parameters are a bit more tricky, but intuitively it's not strictly a means of abstraction.

Maybe I'm relying on too ancient sources here, please enlighten me.

Ericson2314:
Yeah `import`, `builtins.readFile`, etc. are cached in ways that only makes sense if we assume the underlying data doesn't change. And pure eval also affects what can be read in the first place. So at the very least, any impurity here is an unimportant detail.

fricklerhandwerk:
@infinisil Sorry, I absolutely didn't want to sound dismissive of your comments. Proliferating the note on bike-shedding is of course driven by some frustrating experiences in the past. There is a recurrent pattern I notice where people (including myself) appear to jump between levels of detail. But that doesn't have anything to do with your work specifically. I agree on the side effects topic and will think it over. It has to be correct in the big picture.

source/tutorials/nix-language.md
infinisil: I think this might be important from what you linked:

> In programming, we deal with two kinds of elements: functions and data. (Later we will discover that they are really not so distinct.)

Because in functional programming languages, functions are first-class citizens and can be passed around like any other data. This then makes the distinction between data and functions tricky, because functions can be both. In the end I think it's really just a matter of convention about how terms are defined, but this leads to a fuzzy categorization. But I get your idea of making this distinction for having a better tutorial flow.

Taking that into account though, I think it would be good to categorize it into "simple expressions" and "function declaration and application"

infinisil:
I'm not sure how much you'll appreciate this, but I did a little exploration of how I'd organize such a tutorial. I focused on it being very incremental in terms of complexity and not having any prior knowledge. I like the categorization you did in https://github.com/NixOS/nix.dev/pull/267#discussion_r971337777, so I'm leaning onto that here. Note that I also listed some things yet missing from this tutorial (notably `or`, `if then else`, `elemAt`, `++` and more). I don't think we need to go in detail about all the operators, but pointing out that they exist would be good, and I think it would flow very nicely to introduce operators along the types they operate on.

- Simple expressions (everything except functions declaration and function application)
  - Primitive data types and their operations: `null`, booleans (and `!`, `&&`, `||`, `->`, `if then else`), ints and floats (and `+`, `-` (incuding as negation), `*`, `/` and all the comparison operators), literal strings (_without_ antiquotation, but with multiline, also `+` for string concatenation), paths (absolute, relative, and `+` specifically for path concatenation), mention that these evaluate to themselves
  - Basic variables: `let in` along with `inherit` syntax (don't mention attribute sets yet)
  - Composed data types and their operations: attribute sets (and `.` access, including `or`, `//` and `?` also `rec` and `with` and `inherit`), lists (and `++`, `{builtins,lib}.elemAt`), mention that these can contain _any_ Nix expression, and that they evaluate to themselves too, mention equality comparison, that it works on both primitive and composed data types, `==` and `!=`. Here the note about it being similar to JSON up to this point (ignoring operators and variables) can be put
  - String antiquotation: Include basic escaping, how it interacts with paths, mention `toString`
- Declaring and applying functions: Single and multiple arguments (currying), attribute set arguments, ellipsis, with `@`, attribute, defaults, how to call these
- Builtins and library functions
  - Importing Nix expressions from files: `import`, mention that it's common to apply arguments to the imported expression
  - Tracing and errors: `builtins.trace`, `throw`, `assert`, also mention/link to `lib.debug`
  - Link to `builtins` and `lib` manuals
  - Derivation primitive, mention and [link to](https://nixos.org/manual/nixpkgs/stable/#idm140737320134080) `stdenv.mkDerivation`
  - Fetchers, builtins but also mention nixpkgs fetchers
- Impurities (especially relevant for pure evaluation/flakes): `<NIX_PATH>`, `builtins.currentSystem`, fetchers without hashes

fricklerhandwerk:
@infinisil Thank you, I appreciate your input and its kind a lot!

Its great that you have written out in detail the orthogonal direction of what I intended to achieve, because so far I did not motivate my approach in depth. It helps me question, re-evaluate, and defend or give up the things I got convinced about in the process. Thank you.

The strategy you describe would probably be suitable for people who have barely any programming experience at all, and this is deliberately not the audience we want to address right now. I set out to explicitly follow the philosophy behind [tazjin/nix-1p](https://github.com/tazjin/nix-1p): you know how programming languages work in general, here are just the  things unique to Nix that you will see every day. The diff to such a brief rundown is really just the level of detail and care, because we can't expect readers to fill the gaps, left by lecture notes such as `nix-1p`, on their own.

Specifically, this is why I would not even deal with booleans and conditionals, they are boring and unsurprising. They work like in any other language. I would also not deal with `->` because, while surprising, its rarely used in the wild. You'll deal with it when you're deep enough down the rabbit hole. Operators are mentioned alongside `builtins`, that should be enough. Read the reference manual.

One primary reason to go down that route is to save people's time. nix.dev motto is "getting things done". I see no reason teaching people basic programming language concepts here. The overview in terms of "elements of programming" is just to draw an outline, make a reminder of what it's all about, and encourage people not to feel overwhelmed early on.

This of course bears a risk of losing people when they are lacking prerequisites that we implicitly assume. We try to avoid that by stating them (if ever so vaguely), and at least set expectations for everyone involved.

I think your proposed outline would be very suitable for the natural follow-up on actually writing Nix language code. This is where I see `elemAt`, `if ... then ... else`, when to put parentheses around things, what to watch out for when merging attribute sets, and how to find and use common library functions.

There should also be a companion piece that ports [Nix Language Quirks](https://nixos.wiki/wiki/Nix_Language_Quirks) to nix.dev.

None of this is relevant for understanding tutorials here on nix.dev or reading existing code, though.

A few details on your notes:

> Basic variables: `let in` along with `inherit` syntax (don't mention attribute sets yet)

TIL that `inherit` also works for `let ... in`, i.e. the following is valid

```nix
let a = 1; in let inherit a; in a
```

    1

I would absolutely definitely not include this here though, because I haven't consciously seen this in the wild in 4 years of using Nix.

> Single and multiple arguments (currying)

According to [Wikipedia's definition of currying](https://en.m.wikipedia.org/wiki/Currying), Nix doesn't have it. There *are* no multivariate functions, there is nothing to curry or uncurry. I stumbled over this multiple times when writing and rewriting the respective section. I've seen the term being used colloquially, and probably what we'd mean here is "curried functions" as used in the Wikipedia article (as that doesn't refer to the process of currying, but the form of representation), but I doubt it will make things clearer. I still think it may be worthwhile to link to that article for people who aren't familiar with that concept at all, if we find a way to express this gently. Maybe in the detailed explanation...

fricklerhandwerk:
Also I thought about "simple expressions" vs "functions" again, and I'm still not convinced this is a suitable dichotomy. After taking a look at the [introduction to Racket](https://docs.racket-lang.org/guide/syntax-overview.html) for a change of perspective, which use the word "simple" way too often for my taste, I'm not even sure what "simple" is supposed to mean. What makes `(def name value)` "simpler" than `(fun arg)` or `(if cond then else)`? Not to be dismissive, but that specific Racket guide fails to guide me as to what is most important to understand about Racket except "well, everything".

What would "simple" mean in our context here?

I agree that expressions denoting primitive values evaluate to themselves, okay. (Although that's not true for relative paths.) As said, I would like to structure the guide according to what each language constructs *means* when dealing with Nix language code, and its relative importance in expressing things we care about in the Nix ecosystem. The guide is to help people understand what we're trying to express. A purely technical distinction at the PLT level simply doesn't seem fit for that purpose.

fricklerhandwerk:
- @infinisil if then else should be mentioned, as it is different in every language
    - @fricklerhandwerk if then else is not even something you see regularly, it's usually wrapped
        - @infinisil let me open some random file in Nixpkgs... well, it's not that many instances indeed
            - it's still worth having it mentioned
        - sure, we should include it, just don't know where to put it
        - @infinisil can add a section on operators and accommodate if-then-else there
        - good idea. can we merge it with this structure already, and add the new section later?
- (lively review of structure)
    - No need to move up `let in`
    - Make `rec` less prominent
    - Side effect -> Impurity
    - guide -> tutorial
- @infinisil can live with that structure. not how I would write a tutorial, but it seems to get the job done.
- @Ericson2341: Is there another tutorial somewhere? Could we point out impurities better there?
    - @fricklerhandwerk a treatise on purity vs. impurity would be something that may actually grow best on top of the Nix Pills!
        - proposal for Nix Pills subtitle: Building the Nix ecosystem from first principles
        - idea:
            - reference = manuals
            - explanation = Nix Pills (grown to have more structure and cover more ground)
            - tutorials, guides ~ still have to figure that out
                - would like to talk again with @jonringer and @nrdxp on what we should focus on each
                - nix.dev says "getting things done", so probably that should be guides
                - doesn't matter much where things live or how they are called, as long as they are easy to find and meet user's needs
- @Ericson2341: Should the tutorial support random access?
  - @fricklerhandwerk: No, this would be for the reference manual
- Is `import` an impurity?
    - No, it's not relevant for reading the code anyway
    - @fricklerhandwerk the only thing I insist on is the distinction between stuff we see immediately in the code and stuff that is on the outside, such as environment variables and files on the network.
        - these may not be side effects, but they are also not under our full control
      - @LucPerkins how about "ambiguity"?
          - @fricklerhandwerk is that a common term? @matta already pointed out an instance where I made up a term, and that was confusing to readers

matta:
> The strategy you describe would probably be suitable for people who have barely any programming experience at all, and this is deliberately not the audience we want to address right now.

One or both of you may be interested in looking at a book: "COMMON LISP: A Gentle Introduction to Symbolic Computation." There is a PDF at the author's website here: https://www.cs.cmu.edu/~dst/LispBook/book.pdf

...not as an example of what this tutorial should be, but more of an example of just how much it takes to thoroughly explain things clearly.

The book is one of the best examples I have seen of slowly and thoroughly introducing a programming language to a reader who may not yet know how to program or even what programming really is.  To give you a quick sense of how thoroughly the book treats the subject:

1. It takes 30 pages before the idea of a list is introduced...on a book teaching LISP!
2. It takes 80 pages before function definitions are shown in lisp syntax (parentheses, etc.). Before that it is all box flow diagrams.
3. It takes 100 pages before the idea of using the LISP repl is introduced.

I think it is quite reasonable to expect a short tutorial to take some short cuts.  :-)

fricklerhandwerk:
Case in point. And TBH I'd rather not spend years of my (or anyone's) life teaching someone the Nix language, of all things.

infinisil:
Something important I'd like to mention is that calling file paths and fetchers an impurity is not really accurate:
- Paths are considered pure _if_ they are in the same "project scope" or "external", e.g. the same Git repository. In practice probably more than 99% of paths follow that pattern, and nobody considers these an impurity.
  - Also paths themselves aren't an impurity, it's reading those paths. Notably you can convert a path to a string without reading it using `toString`, though that's not very common.
  - Notably, absolute paths should _always_ be considered an impurity.
  - Relative paths can also be impure, but only if they escape the project scope, e.g. with `../../../home/myUser/some/file`
- Similarly, builtin eval-time fetchers like `builtins.fetchurl` are only considered an impurity if they don't provide a result hash. If they have a hash, Nix and the Nix community considers it pure. This is comparable to build-time fetchers (fixed-output derivations).
- In general, flakes (or more specifically pure evaluation mode) should be the gold standard for what is an impurity or not. We shouldn't call something an impurity if pure evaluation allows it

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+```nix
+let
+  f = {a, b, ...}: a + b;
+in
+f { a = 1; b = 2; c = 3; }
+```
+
+    3
+
+### Named attribute set argument
```
tazjin, line 1166: Note that while formals definitions can refer to each other (e.g. `{ a, b ? a * 2 }: ...`) they can not refer to the named pattern (e.g. `{ a, b ? args.a }@args: ...` is invalid)

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+
+It precludes build tasks from referring to files which are not explicitly specified as build inputs.
+
+Nix supports other types of impure expressions, such as [search paths](search-path) or the constant [`builtins.currentSystem`](https://nixos.org/manual/nix/stable/language/builtin-constants.html#builtins-currentSystem).
+We do not cover those here more detail, as they do not matter for how the Nix language works in principle, and because they are discouraged for the very reason of breaking reproducibility.
+:::
+
+### Paths
+
+Whenever a file system path is rendered to a character string with [antiquotation](antiquotation), the contents of that file are copied to a special location in the file system, the *Nix store*, as a side effect.
```
tazjin, line 1371: The link syntax is broken here

tazjin (commented):
Not all the way through yet, but food has appeared, so I'm leaving these comments for now :)

infinisil:
@tazjin Maybe try to not go into too much detail. We discussed this in the documentation team a bunch already, the goal of this tutorial is to get beginners to be able to read typical Nix code. The goal is not to document every feature and particularity of Nix (the reference can be a place for that, or we can have another Nix tutorial more like I described [here](https://github.com/NixOS/nix.dev/pull/267#issuecomment-1252815161)), so things like dynamic keys, details about scoping and laziness should probably not be in scope.

Also, @fricklerhandwerk would like to get this PR merged sooner rather than later, mainly to pin the general direction of this tutorial. Minor adjustments can easily be made at a later point.

So I encourage you to especially review the _structure_ and _correctness_ of this PR and less details :smile:

matta:
> Something important I'd like to mention is that calling file paths and fetchers an impurity is not really accurate:

@infinisil, I think there a difference between language level purity and build level purity that could be distinguished here.  

Personally, I think language level purity and build level purity are completely different and separable concepts.

The current `nix-language.md` says "Purity is the key to reproducible builds." in a section that I think conflates these two kinds of "purities" without explanation.  Language level purity is not necessary for reproducible builds, though it may be a useful programmer aid.  Reproducible builds are achievable with shell scripts!  :-)

infinisil:
Yeah, there's even three notions of purity:
- Eval time, enforced with `--pure-eval`
- Build time, mostly enforced with the build sandbox
- Runtime, mostly done by patched compilation and co. in nixpkgs

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+
+- [Assigning names and accessing values](names-values)
+- Declaring and calling [functions](functions)
+- [Built-in and library functions](libraries)
+- [Side effects](side-effects) to obtain build inputs
+- [Derivations](derivations) that describe build tasks
+
+:::{important}
+This guide *does not* explain all Nix language features in detail.
+See the [Nix manual][manual-language] for a full language reference.
```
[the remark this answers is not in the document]
fricklerhandwerk: I generously and deliberately apply wishful thinking w.r.t. the manual. We (as a community) will fix forward once someone gets around to it.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+[^1]: Details: Nix manual - attribute naming rules <!-- TODO: create and link manual section -->
+[^2]: Details: [Nix manual - lists][manual-lists]
+
+[manual-lists]: https://nixos.org/manual/nix/stable/language/values.html#list
+
+(rec-attrset)=
+#### Recursive attribute set `rec { ... }`
+
+You will sometimes see attribute sets declared with `rec` prepended.
+This allows access to attributes from within the set.
```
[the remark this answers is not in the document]
fricklerhandwerk: I don't go into dynamic attributes because they are rare enough in practically relevant code not to matter here. If you just learn about Nix and want to go through some tutorials here you won't dig into Nixpkgs internals or NixOS module implementations where those become more prominent.

source/tutorials/nix-language.md
fricklerhandwerk: If you find bugs, please leave suggestions I can merge immediately. Otherwise I'll have to dig through each comment individually to figure out what exactly was wrong and how to fix it.

fricklerhandwerk:
@infinisil this kind of thing should go to the Nix Pill as well. It's obscure enough not to matter in practice except for dark corners of very sophisticated use cases - implementers of which would have to acquire familiarity at the level of breadth and depth of the Pills - and yet actually enlightening once you know enough concepts to connect the dots.

Concerning impurities how they're used here you're both right, thanks for pointing out that oversight.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+    Welcome to Nix 2.5.1. Type :? for help.
+
+```console
+nix-repl> 1 + 2
+```
+
+    3
+
+:::{note}
+The Nix language by default uses lazy evaluation, and will only compute values when needed.
```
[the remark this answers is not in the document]
fricklerhandwerk: I think this is a close-enough characterization for an introduction. Such details can be left for the reference manual.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+echo 1 + 2 > default.nix
+
+nix-instantiate --eval
+```
+
+    3
+:::
+
+:::{note}
+The Nix language by default uses lazy evaluation, and will only compute values when needed.
```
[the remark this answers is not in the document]
fricklerhandwerk: Yes, intentional.
a) I expect people to scroll over this introductory boilerplate because they'd pick the `nix-repl` approach.
b) The invocation here is different and I think motivating it place is meaningful.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+  one = 1;
+  two = one + 1;
+  three = two + 1;
+}
+```
+
+    { one = 1; three = 3; two = 2; }
+
+:::{note}
+Attribute sets are unordered.
```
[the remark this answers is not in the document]
fricklerhandwerk: Out of curiosity, how is stable output relevant in Nix's domain?

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+let
+  attrset = { a.b.c = 1; };
+in
+attrset
+```
+
+    { a = { b = { c = 1; }; }; }
+
+(with)=
+### `with ...; ...`
```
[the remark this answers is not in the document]
fricklerhandwerk: We will not note this here, because it will not matter much for beginners.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1679 @@
...
+
+Example:
+
+
+```nix
+../.
+```
+
+    /current
+
```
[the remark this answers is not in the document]
fricklerhandwerk: This is true and so rarely seen that it's unimportant here.

source/tutorials/nix-language.md
fricklerhandwerk: Important only for writing, not reading code.

fricklerhandwerk:
@infinisil @domenkozar @matta I addressed all comments, except anything regarding CSS -this can (and should be, at some point) be done in a separate pull request.

Would be great if we could merge this to unleash it onto the world, so I can focus on something else.

fricklerhandwerk:
@domenkozar where can we see the preview?

domenkozar:
For some reason it doesn't work, probably because the PR is older than the pages installation. We could fix it by opening a new PR, what do you think?

fricklerhandwerk:
Shall we instead merge and fix forward? Your editors will have an easier time working on language that way.

source/tutorials/nix-language.md
```diff
@@ -0,0 +1,1766 @@
...
+### Attribute set `{ ... }`
+
+An attribute set is a collection of name-value-pairs, where names must be unique.
+
+The following example shows all primitive data types, lists, and attribute sets.
+
+:::{note}
+If you are familiar with JSON, imagine the Nix language as *JSON with functions*.
+
+Nix language data types *without functions* work just like their counterparts in JSON and look very similar.
```
sternenseemann, line 329: This is not true. It's fine to say *think about* Nix as JSON plus functions, but this is *only* an analogy. Specifically:

* Strings in C++ Nix are byte sequences and not Unicode-encoded in some shape or form as in JSON.
* Integers in Nix are 64 bit integers and not represented as floating point numbers as they are in JSON.
* Floats overall behave weirdly in Nix to the point that you should warn about them, i.e. they get truncated when serializing them in many situations.
* Due to the nature of Nix's strings, not every legal Nix attribute name is a legal JSON object key.

yannham: I think the formulation "imagine the Nix language as" is somehow similar to saying "think about Nix as", isn't it?

sternenseemann: Yes, but the claim in the second sentence is stronger than that and wrong, as I've pointed out. “Nix language data types *without functions*” does *not* work just like their counterparts in JSON.

tazjin:
> Why do we call it "antiquotation"?

Do we though? I've never actually heard anyone say that, and usually people talk about interpolation instead. The term really only seems to show up in documentation, in the implementations of Nix, rnix, and Tvix (probably also hnix) they're always referred to as "string interpolation" and "Interpolation parts".

Maybe we should get rid of the "antiquotation" term completely.

source/tutorials/nix-language.md
sternenseemann: I feel that it is at least implied. With tutorials as the first contact point, one should be careful to avoid creating wrong impressions in readers – even by accident. FWIW my point is not that these differences *must* be explained, but rather that we should be careful to word it in a way that we don't imply they don't exist.

---

I'm also not sure these edges are that far out.

The integer thing is a relatively well known JSON pitfall – Nix is not thorny here as it behaves like you'd expect.

```console
> echo 34343499993845656749 | jq
34343499993845658000
```

The float situation can be discovered quite easily. E.g.:

```console
nix-repl> 1.000004 
1
nix-repl> builtins.toJSON 1.000004
"1"
```

String indexing, I'd say is a very basic topic that should be covered in a tutorial, since the internal representation is apparent in this, it is relevant. Indexing into non-ASCII strings will unexpectedly  e.g. `builtins.substring 0 1 "καλημερα"` will return in some garbage being returned. E.g. “Nix strings are similar to C strings” is a statement like the one about JSON data, but more accurate.

fricklerhandwerk:
@mkaito thank you for this extremely valuable insight. I will make the appropriate updates ASAP.

@tazjin Agreed, we should rename "antiquotation" to "string interpolation" in the Nix manual. Sometimes it's the small changes that make a big difference. What are "interpolation parts" though? Is it the same thing?

tazjin:
@fricklerhandwerk Interpolation parts are the individual parts of a string with interpolations, e.g. in `"one${two}three${four}"` the `two` and `four` are interpolation parts.

PR #295, opened 2022-08-02
remove "Software Nix has influenced"

fricklerhandwerk:
the glossary is empty, and influences are irrelevant for the objective
of learning Nix

### DIFF
--- a/source/influences.md
+++ b/source/influences.md
@@ -1,7 +1,0 @@
-# Software Nix has influenced
-
-- https://github.com/alexanderGugel/ied
-- https://www.habitat.sh/
-- https://www.gnu.org/software/guix/
-- https://github.com/andrewchambers/hermes
-- https://bobbuildtool.dev/
--- a/source/reference/index.md
+++ b/source/reference/index.md
@@ -13,5 +13,4 @@ NixOS Manual <https://nixos.org/manual/nixos/stable/>
 pinning-nixpkgs.md
 glossary.md
 ../recommended-reading.md
-../influences.md
 ```


domenkozar:
Why do you think we should remove glossary? It's useful to be able to reference it here for the most basic terms that a beginner might encounter for the first time.

For influences, it's a sign of a health community to show that others are building work on top of what we're doing.

fricklerhandwerk:
> Why do you think we should remove glossary?

Because it only contains TODOs.

You said you would prefer not to present skeletons, and I agree in part: have a scaffold and give it some contents to get the feeling for the whole thing. Having two terms in the Glossary is probably fine, but they should at least be defined.

(Having a [visual] outline would also be good if some parts are filled - right now the textual one does the same thing.)

Another point is that I think the glossary should be part of the manual(s), as those are (supposed to be) the reference. A non-manual Glossary should at best contain the more softly defined jargon from the ecosystem. And even then we should choose where to maintain a single source of truth - here or the Wiki?

> For influences, it's a sign of a health community to show that others are building work on top of what we're doing.

Yes, agreed. Question is whether that should be curated here, and my intuition is we should sharpen the focus on beginner and problem solving and guides, and move these softer issues to the Wiki.

One side effect is reducing the length of the main navigation side bar to depict that very clear focus. Just trying to ease overall navigation within the ecosystem.

fricklerhandwerk:
@NixOS/marketing-team This information should be maintained by marketing.

infinisil (approved):

henrik-ch (approved):

Issue #281, opened 2022-06-29
teaching Nix 3 CLI and Flakes


fricklerhandwerk:

domenkozar:
![2023-07-17-120427_830x571_escrotum](https://github.com/NixOS/nix.dev/assets/126339/27ab855f-3da6-48eb-b4dd-d673a140de31)
![2023-07-17-120337_917x440_escrotum](https://github.com/NixOS/nix.dev/assets/126339/7d917e94-9c4d-4271-8277-b31e2975300f)

Ericson2314:
I left a comment on the tweet: https://twitter.com/ericson2314_/status/1680630061353824256 We should not promote unstable features, or "unstable" doesn't mean anything. Period. I wrote an RFC for how and when to stabilize Flakes, https://github.com/NixOS/rfcs/pull/136. Pro-Flakes people have largely ignored the RFC.

If any wants Flakes to be promoted in official teaching material, they should want Flakes to be stable, and they should therefore engage with the RFC or counter-propose something else.

PR #307, opened 2022-08-17
Bring anti-patterns more inline with nixpkgs contributing guide

SuperSandro2000:


SuperSandro2000:

SuperSandro2000:
recheck

source/anti-patterns/language.md
```diff
@@ -42,6 +42,10 @@ in {
 }
 ```
 
+:::{note}
+It is fine to use `rec` in nixpkgs' stdenv.mkDerivation function especially for reusing version in a fetcher for src since that is a fixed output derivation anyway.
```
fricklerhandwerk, line 46: I can't follow the reasoning here. What makes the fixed output derivation special in this case?

source/anti-patterns/language.md
```diff
@@ -31,7 +31,7 @@ There are a couple of pitfalls:
 - combining with overriding logic such as `overrideAttrs` function in nixpkgs has a surprising behaviour
   of not overriding every reference.
 
-A better way is to use simpler `let .. in`:
+A better way is to use simpler but more verbose `let .. in`:
```
fricklerhandwerk, line 34:
```suggestion
A better way is to use `let ... in`:
```

I'm not sure in what sense `let ... in` is simpler, and I don't think it's worthwhile pointing out it requires more characters to express.

source/anti-patterns/language.md
```diff
@@ -70,23 +74,9 @@ with (import <nixpkgs> {});
 # try this instead:
 let
   pkgs = import <nixpkgs> {};
-  inherit (pkgs) curl jq;
 in ...
 ```
 
-```nix
```
fricklerhandwerk, line 77: Since I can't make suggestions on deleted lines (WTF GitHub?) and lines not modified:

l. 68:

> We recommend the following:

l. 80:

> Use `with` only for tight scopes which leave no ambiguity:
> 
> ```
> buildInputs = with pkgs; [ curl jq ];
> ```
> ```
> meta.maintainers = with lib.maintainers; [ edolstra ];
> ```

source/anti-patterns/language.md
```diff
@@ -100,7 +90,7 @@ In most cases, `$NIX_PATH` is set to the latest channel when Nix is installed, a
 
 :::{note}
 [Channels](https://nixos.wiki/wiki/Nix_channels) are a way of distributing Nix software.
-They are being phased out, but still used by default.
+They are being phased out by flakes, but still used by default.
```
fricklerhandwerk, line 93: We should first explain visibly what flakes are and how to use them before telling people "channels are all over the place, but forget that. use flakes, but you'll have to figure out what they even do on your own."

So we would only refer to flakes once we have an authoritative (and hopefully satisfactory) definition. There are  RFC49 (closed!) and Eelco's blog posts, and I'm fine with linking those, but something with Nix™ on them would be better.

fricklerhandwerk, line 93: We should first explain visibly what flakes are and how to use them before telling people "channels are all over the place, but forget that. use flakes, but you'll have to figure out what they even do on your own."

So we would only refer to flakes once we have an authoritative (and hopefully satisfactory) definition. There are  RFC49 (closed!) and Eelco's blog posts, and I'm fine with linking those, but something with Nix™ on them would be better.

I'm not even sure if the statement was true to begin with. What about a bit more cautiously saying

```suggestion
They are used by default, but there are plans to replace them with a more robust system.
```

source/anti-patterns/language.md
```diff
@@ -116,7 +106,7 @@ But that is still problematic, unless:
 2. you control the shell environment via your source code, setting `$NIX_PATH` via `nix-shell` or NixOS options.
 
 :::{note}
-We recommend to avoid using search paths and to disable channels by permanently setting `NIX_PATH=` to be empty.
+We recommend to avoid using extra search paths and to switch to flakes and disable channels by permanently setting `NIX_PATH=` to be empty.
```
fricklerhandwerk, line 109: Again, switching to flakes should be supported by authoritative material on how to do that, otherwise it will just confuse people.

```suggestion
We recommend to avoid using search paths and disable channels by permanently setting `NIX_PATH=` to be empty.
```

The docs team decided "[not to drag people into unstable features](https://github.com/NixOS/nix.dev/issues/281#issuecomment-1198128399)" and instead build up Nix 3 and Flakes on a separate branch.

source/anti-patterns/language.md
infinisil, line 46: Should be discouraged a bit more, it's not great but the best we have in some cases.

source/anti-patterns/language.md
infinisil, line 93: Let's not recommend flakes yet, but now we have this page giving some flakes context: https://nix.dev/concepts/flakes#flakes

source/anti-patterns/language.md
infinisil: Agreed, we shouldn't recommend flakes

infinisil (changes requested):
Other than these bits, it looks okay, we reviewed this in todays docs team meeting

infinisil:
@SuperSandro2000 Can you follow up on the feedback?

PR #315, opened 2022-09-01
update installation instructions

fricklerhandwerk:
Darwin no longer needs special incantations


domenkozar:
Seems like something went wrong with rebasing?

PR #320, opened 2022-09-07
add highlighting for shell input/output

sbruder:
This implements highlighting for shell input/output. To use it, two adjacent code blocks must have `shell-session` and none as their highlighting language.

As for how to highlight it, I copied my approach from the Summer of Nix homepage (which itself is based on the German Ubuntu Wiki) and inverted the text/background colours. I’m not sure if that is the best way to handle this here, because the output also gets highlighted and might look weird with inverted colours.

Also, I noticed that Sphinx does not check if the files in `html_static_path` have changed on a rebuild, so they will not be replaced unless a full build is triggered.

cc @fricklerhandwerk

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -1,0 +1,4 @@
+div.highlight-shell-session + div.highlight-none pre {
+  background-color: rgba(var(--pst-color-preformatted-text),1) !important;
+  color: rgba(var(--pst-color-preformatted-background),1) !important;
+}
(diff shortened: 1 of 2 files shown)


source/_static/css/custom.css
```diff
@@ -0,0 +1,4 @@
+div.highlight-shell-session + div.highlight-default pre {
```
fricklerhandwerk, line 1: Tested this with the following

````
```shell-session
foo bar
```

    baz
````

and it did not show a difference, because the CSS class in that case is `hightlight-none`.

```suggestion
div.highlight-shell-session + div.highlight-none pre {
```

fricklerhandwerk (approved):

PR #325, opened 2022-09-16
add tutorial for nix-shell in shebang

rapenne-s:
This explains how to use `nix-shell` in a shebang.

I'm open to changes to better fit nix.dev style :+1:

### DIFF
--- a/source/tutorials/nix-shell-in-shebang.md
+++ b/source/tutorials/nix-shell-in-shebang.md
@@ -68,7 +68,7 @@ Make it executable using
  chmod o+x nixpkgs-releases.sh
  ```
  
-and run it with
+Run the script:
 
 ```console
 ./nixpkgs-releases.sh
--- a/source/tutorials/nix-shell-in-shebang.md
+++ b/source/tutorials/nix-shell-in-shebang.md
@@ -1,0 +1,33 @@
+# How to use nix-shel in a shebang
+
+The [nix
+shell](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html)
+is feature provided by the [Nix package
+manager](https://nixos.org/guides/how-nix-works.html). It allows users
+to enter into a new shell environment containing a set of packages given
+on the command line.
+
+You can nix-shell as an interpreter](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#use-as-a--interpreter)
+in the shebang, not only to define the interpreter, but also a list the
+dependencies required for the script file.
+
+Imagine a simple shell script downloading an XML file, running a linter
+on it and sending an email of the result. It requires `curl`, `xmllint`
+and `mail`.
+
+```shell
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p curl libxml2 mailutils
+
+curl http://localhost.example/report.xml | xmllint -
+
+if [ $? -eq 0 ]; then
+    echo "The report is valid" | mail someone@localhost.example
+else
+    echo "The report is invalid" | mail someone@localhost.example
+fi
+```
+
+By using `nix-shell` as the shebang, you don't have any prerequisites
+steps like typing `nix-shell` or installing packages before using the
+program.
--- a/source/tutorials/nix-shell-in-shebang.md
+++ b/source/tutorials/nix-shell-in-shebang.md
@@ -10,8 +10,10 @@ In this tutorial, you will learn how to use the [nix shell] to create reproducib
 
 - A working [Nix installation](install-nix)
 - Familiarity with [Bash]
+- Familiarity with the concept of [shebang]
 
 [Bash]: https://www.gnu.org/software/bash/
+[shebang]: https://en.m.wikipedia.org/wiki/Shebang_(Unix)
 
 # A trivial script with non-trivial dependencies
 
@@ -28,11 +30,6 @@ It also requires the `bash` interpreter.
 If any of these dependencies are not present on the system running the script, it will fail partially or altogether.
 
 We will use `nix-shell` interpreter directives to declare all dependencies explicitly, and produce a script that will always run on any machine that supports Nix and the required packages from Nixpkgs.
-
-The characters `#!` at the beginning of a script are known as a [shebang].
-This is a mechanism to tell which program should be used to interpret the following code.
-
-[shebang]: https://en.m.wikipedia.org/wiki/Shebang_(Unix)
 
 We will use the [nix-shell as an interpreter] in the shebang, not only to define the interpreter, but to also list the packages needed to provide the commands we saw in the previous section: `bash`, `python3Packages.xmljson`, `jq` and `curl`.
 
--- a/source/tutorials/nix-shell-in-shebang.md
+++ b/source/tutorials/nix-shell-in-shebang.md
@@ -1,36 +1,38 @@
 # Reproducible interpreted scripts
 
 In this tutorial, you will learn how to use the [nix shell] to create reproducible interpreted scripts.
-We will write a script requesting latest releases of NixOS/nixpkgs GitHub repository, convert it from XML to JSON and pretty-print the result.
 
 [nix shell]: https://nixos.org/manual/nix/stable/command-ref/nix-shell.html
 
 # Steps
 
-## Getting Nix
-
-The first step is to [install the Nix package manager].
-
-[install the Nix package manager]: https://nix.dev/tutorials/install-nix
-
 ## Requirements
 
-In order to write our script, we must thinkg beforehand about which programs we will need.
+- A working [Nix installation](install-nix)
+- Familiarity with [Bash]
 
-1. get the content of the url `https://github.com/NixOS/nixpkgs/releases.atom`
-2. convert the XML into a JSON structure
-3. display the JSON into the terminal so it's human readable
+[Bash]: https://www.gnu.org/software/bash/
 
-For the first step, we will need the program curl, provided by the nix package `curl`.
-For the second step, we will need the program `xmljson`, provided by the nix package `python3Packages.xmljson`. As it's a python program, it's bundled at the python3 packages level.
-For the third step, we will need the program `jq`, provided by the nix package `jq`.
+# A trivial script with non-trivial dependencies
 
-Finally, the whole program will run in a `bash` shell, provided by the nix package `bash`.
+Take the following script, which fetches the content XML of a URL, converts it to JSON, and formats it for better readability:
 
-## Shebang theory
+```bash
+#! /bin/bash
 
-The shebang is the name of the first line of an interpreted script, and it looks like `#!/bin/some_path`.
-This is a mechanism to tell which binary should be used when running an executable script.
+curl https://github.com/NixOS/nixpkgs/releases.atom | xml2json | jq .
+```
+
+It requires the programs `curl`, `xml2json`, and `jq`.
+It also requires the `bash` interpreter.
+If any of these dependencies are not present on the system running the script, it will fail partially or altogether.
+
+We will use `nix-shell` interpreter directives to declare all dependencies explicitly, and produce a script that will always run on any machine that supports Nix and the required packages from Nixpkgs.
+
+The characters `#!` at the beginning of a script are known as a [shebang].
+This is a mechanism to tell which program should be used to interpret the following code.
+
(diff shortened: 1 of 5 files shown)


source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,33 @@
...
+The [nix
+shell](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html)
+is feature provided by the [Nix package
+manager](https://nixos.org/guides/how-nix-works.html). It allows users
+to enter into a new shell environment containing a set of packages given
+on the command line.
+
+You can nix-shell as an interpreter](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#use-as-a--interpreter)
+in the shebang, not only to define the interpreter, but also a list the
+dependencies required for the script file.
```
fricklerhandwerk, line 12: Depending on whether you want this to be a Tutorial or a Guide:
- for a Tutorial, the motivation for should come right after the page title, and followed by sections with a sequence of instructions (and detailed explanations where necessary)
- for a Guide, there should be only a very brief description of the feature, such as

```suggestion
Nix can [wrap interpreted scripts](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#use-as-a--interpreter) with a shebang to make them reproducible.
It allows you to precisely define all dependencies of a script, including the exact version of the interpreter.

Such a script will run with a working [Nix installation](install-nix) as its only prerequisite.
```

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,33 @@
...
+if [ $? -eq 0 ]; then
+    echo "The report is valid" | mail someone@localhost.example
+else
+    echo "The report is invalid" | mail someone@localhost.example
+fi
+```
+
+By using `nix-shell` as the shebang, you don't have any prerequisites
+steps like typing `nix-shell` or installing packages before using the
+program.
```
fricklerhandwerk, line 33: Moved to the top and reworded. 

```suggestion
```

To avoid ending this article abruptly, we may want to add related material, such as tutorials/guides for including scripts in nix derivations as "Next steps".

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,33 @@
...
+to enter into a new shell environment containing a set of packages given
+on the command line.
+
+You can nix-shell as an interpreter](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#use-as-a--interpreter)
+in the shebang, not only to define the interpreter, but also a list the
+dependencies required for the script file.
+
+Imagine a simple shell script downloading an XML file, running a linter
+on it and sending an email of the result. It requires `curl`, `xmllint`
+and `mail`.
```
fricklerhandwerk, line 16:
```suggestion
The following shell script example downloads an XML file, runs a linter on it, and sends an email depending on the result.
It requires `curl`, `xmllint`, and `mail`.
```

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,33 @@
+# How to use nix-shel in a shebang
+
+The [nix
+shell](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html)
+is feature provided by the [Nix package
+manager](https://nixos.org/guides/how-nix-works.html). It allows users
+to enter into a new shell environment containing a set of packages given
+on the command line.
```
fricklerhandwerk, line 8: These lines are really hard to read in source. Please make use of [our markdown conventions](https://github.com/NixOS/nix.dev/blob/master/CONTRIBUTING.md?plain=1#L173-L185).

fricklerhandwerk (changes requested):
Note that there already is a [section on shellbang in the article on ad hoc environments](https://github.com/NixOS/nix.dev/blob/master/source/tutorials/ad-hoc-developer-environments.md?plain=1#L131-L150). I would like to move it out into a separate article, because I noticed that people seem not to find it on their own (including you, apparently, but also me when I had needed something like that in the past). 

Also the Python example there was very confusing to every beginner I discussed it with.

I would like you to clarify more strongly if you want this to be a [Guide](https://diataxis.fr/how-to-guides/) (clear-cut recipe for people who know that they want this feature and why/when to use it, but need a summary or reminder) or a [Tutorial](https://diataxis.fr/tutorials/) (motivating introduction to a concept, which teaches a skill).

I think for shellbang we may actually want a tutorial to introduce beginners, as this still [appears to be a fairly obscure feature](https://news.ycombinator.com/item?id=26474134) although it is so powerful. It is hard to find the feature if you don't know the right search keywords. Googling "nix shell shebang" produces good results, and some very nice third-party posts the spirit of which we should definitely incorporate at some point. Googling "nix wrap shell script" leads to seemingly random stuff, and only one result is relevant, but will probably look shady to most people. We want people to find the manual or nix.dev (and tell them nix.dev is legit).

Here is a rundown of what is out there after a quick search:
- https://gist.github.com/travisbhartwell/f972aab227306edfcfea
- https://nixos.wiki/wiki/Nix-shell_shebang
- https://notes.yukiisbo.red/posts/2021/07/Spice_up_with_Nix_Scripts.html
- https://t-ravis.com/post/nix/nix-make/

source/tutorials/nix-shell-in-shebang.md
rapenne-s: I'll make it a tutorial, that's a nice entry to nix IMO

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,33 @@
...
+
+You can nix-shell as an interpreter](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#use-as-a--interpreter)
+in the shebang, not only to define the interpreter, but also a list the
+dependencies required for the script file.
+
+Imagine a simple shell script downloading an XML file, running a linter
+on it and sending an email of the result. It requires `curl`, `xmllint`
+and `mail`.
+
```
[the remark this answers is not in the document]
rapenne-s: Thanks for pointing nix-index

rapenne-s:

source/tutorials/nix-shell-in-shebang.md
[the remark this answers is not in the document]
rapenne-s: fixed by revamping the content entirely

rapenne-s:
I've not been able to use `nix-index` without locking up my laptop twice due to a lack of memory, it seems 9 GB wasn't enough. I think finding which package is providing a binary should be another tutorial.

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,58 @@
+# Reproducible interpreted scripts
+
+In this tutorial, you will learn how to use the [nix shell] to create reproducible interpreted scripts.
+We will write a script requesting latest releases of NixOS/nixpkgs GitHub repository, convert it from XML to JSON and pretty-print the result.
```
fricklerhandwerk, line 4: We should present the script as given, explain what it does, and then work along that.

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,58 @@
...
+- `-i` to indicates which program should be used to interpret the file, we want `bash`
+- `-p` to indicates a list of packages that should be provided in the current shell, we need `jq python3Packages.xmljson curl`
+- `-I` to indicates special attributes to `nix` such as `nixpkgs` to use a specific pinned version instead of using "the current version available now".
+
+Create a file named `nixpkgs-releases.sh` with the following content:
+
+```shell
+#!/usr/bin/env nix-shell 
+#! nix-shell -i bash -p curl jq python3Packages.xmljson
+#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/2a601aafdc5605a5133a2ca506a34a3a73377247.tar.gz
```
fricklerhandwerk, line 53: As in the other comment, these should be two steps. Since I proposed to delete the original introductory part which maps commands to packages, it should come up here as an explanation.

The other PR about scripts has the same problem with how to find the right package: https://github.com/NixOS/nix.dev/pull/317#issuecomment-1266750934

Probably we should have a dedicated article on that, so it's fine to just present the mapping as given. We'll fix forward then.

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,58 @@
...
+
+curl https://github.com/NixOS/nixpkgs/releases.atom | xml2json | jq .
+```
+
+Make it executable using `chmod o+x nixpkgs-releases.sh` and run it using `./nixpkgs-releases.sh`.
```
fricklerhandwerk, line 58:
````suggestion
Make it executable using

 ```console
 chmod o+x nixpkgs-releases.sh
 ```
 
and run it with

```console
./nixpkgs-releases.sh
```
````

**Review by fricklerhandwerk — CHANGES_REQUESTED** [review]
In this review I focused on structure. Please also add a References section that shows the relevant reference documentation in an overview. You started the NixOS Wiki article on searching packages, that may be a good entry point for people to solve that problem on their own.

There should also be a section on next steps - the natural successor would be setting up declarative environments, and depending on who gets merged first, the article on wrapping scripts in Nix derivations.

File: source/tutorials/nix-shell-in-shebang.md (hunk 7 above)
Comment (rapenne-s) [reply source/tutorials/nix-shell-in-shebang.md:54]: what do you mean by two steps, we should explain each line in a separate paragraph?

Comment (fricklerhandwerk) [reply source/tutorials/nix-shell-in-shebang.md:54]: Yes. There should be a paragraph explaining how the dependencies come into being (`xml` comes from `pytho3Packages.xmljson` etc.) and present the script only with those. (Usually here I would ask to show how to come up with the exact package names, but as said we'll defer that to a dedicated guide.)

And then one to make note of the fact that the script will only reliably run on any other machine if the dependencies are precisely determined by a revision of Nixpkgs, show how to come up with the `-I` line, and then present the full script. (Eventually, finding Nixpkgs commit hashes should also be a dedicated guide we'd link to.)

File: source/tutorials/nix-shell-in-shebang.md [hunk 9]
```diff
@@ -10,8 +10,10 @@ In this tutorial, you will learn how to use the [nix shell] to create reproducib
 
 - A working [Nix installation](install-nix)
 - Familiarity with [Bash]
+- Familiarity with the concept of [shebang]
```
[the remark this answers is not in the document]
rapenne-s: let's try with another version

rapenne-s:
Ping :)

fricklerhandwerk (changes requested):
Almost perfect now, thanks for fixing up and reminding me.

While this is still missing the steps to figure out mappings of binaries and package names, and how to find recent commit hashes, that should be an independent guide we can link to later. Even if the latter will mostly be made obsolete by flakes, until then people will have that issue.

n8henrie (changes requested):
Consider adding the `--pure` flag, which ensures that the packages being used are from the env specified; I happen to have `jq`, `bash`, and `curl` in my `PATH`, so mine runs without error if I omit several of those, but might fail on another system.

For example, running `printf '%b\n' "${PATH//:/\\n}" | wc -l` as the script in question, my PATH is about 3 times as long without the `--pure`.

Further, with `--pure`, the script fails on my system due to absence of `cacert`; consider adding `pkgs.cacert` to the list of dependencies.

fricklerhandwerk:
@n8henrie Thanks for the review. For the sake of clarity in this tutorial maybe we could add this as an additional step, to add the script "more pure"? Although to relieve @rapenne-s from this long-standing open PR I suggest to merge it after fixing the build error. @n8henrie would you be interested in doing the follow-up as you described?

n8henrie:
Sure! Would love to help. Hope to work on this within the next week 🤞

n8henrie:
"Within the next week" was definitely optimistic, but it's on the todo list.

n8henrie:
It also may be worth mentioning the use if `nix-locate` as a handy and efficient way to determine which `-p packages` must be added when converting an existing script to `nix-shell --pure`.

fricklerhandwerk:
Please rebase, the broken link was fixed in #438.

fricklerhandwerk (approved):
Awesome! Thanks a lot for pulling through with this.

PR #262, opened 2022-07-08
assign copyright to NixOS Foundation

fricklerhandwerk:
based on Nix documentation team [discussion on 2022-07-07][notes]

[notes]: https://discourse.nixos.org/t/2022-07-07-documentation-team-meeting-notes-4/20198

PR #332, opened 2022-10-04
nix-language: Don't imply strict eval is a thing

roberth:
This phrases it a bit more accurately, while also using the word `compute` as a correct but general synonym for non-fp novices to grab on to. That said, the last sentence isn't as important to me as the prior edits.

### DIFF
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -115,10 +115,11 @@ nix-repl> 1 + 2
 ```
 
 :::{note}
-The Nix language by default uses lazy evaluation, and will only compute values when needed.
+The Nix language uses lazy evaluation, and `nix repl` by default only computes values when needed.
 
-Some examples show results of strict evaluation for clarity.
+Some examples show a fully evaluated data structure for clarity.
 If your output does not match the example, try prepending `:p` to the input expression.
+This will compute more of the returned value before showing it.
 
 Example:
 
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -119,7 +119,6 @@ The Nix language uses lazy evaluation, and `nix repl` by default only computes v
 
 Some examples show a fully evaluated data structure for clarity.
 If your output does not match the example, try prepending `:p` to the input expression.
-This will compute more of the returned value before showing it.
 
 Example:
 
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -175,7 +175,7 @@ $ nix-instantiate --eval
 :::
 
 :::{note}
-The Nix language uses lazy evaluation, and `nix-instantiate` by default does not compute everything.
+The Nix language uses lazy evaluation, and `nix-instantiate` by default only computes values when needed.
 
 Some examples show a fully evaluated data structure for clarity.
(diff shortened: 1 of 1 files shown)


roberth:
What's up with https://github.com/NixOS/nix.dev/pull/262?

source/tutorials/nix-language.md
```diff
@@ -113,10 +113,11 @@ nix-repl> 1 + 2
     3
 
 :::{note}
-The Nix language by default uses lazy evaluation, and will only compute values when needed.
+The Nix language uses lazy evaluation, and `nix repl` by default only computes values when needed.
 
-Some examples show results of strict evaluation for clarity.
+Some examples show a fully evaluated data structure for clarity.
 If your output does not match the example, try prepending `:p` to the input expression.
+This will compute more of the returned value before showing it.
```
fricklerhandwerk, line 120:
```suggestion
```
Are you saying there is a cut-off for strictness? Since this should not really bother people here, I'd rather not even hint at it.

PS: Its funny how many obscure details I learned about the Nix language (which I neither wanted to be true nor learn about them, simply to sleep better at night) just by generously generalizing.

fricklerhandwerk (changes requested):
Thanks for the clarification. I like the more straightforward wording. Please add these changes to the `nix-instantiate` section as well.

source/tutorials/nix-language.md
roberth: Technically I think `nix repl` computes up to WHNF, then prints what it knows about the value after doing so. That's definitely too much info to put there though.
I'm only stating what's obvious from the behavior of `:p` though, because that's all the understanding we need at this point, but I decided to state it, because I think the functionality of `:p` is important enough that we should not leave it to the reader's inference. Readers are lazy too ;)

> a cut-off for strictness

That's kind of what WHNF is. Btw, I don't think WHNF is discussed in much Nix documentation at all, and iirc it has not been formally defined for Nix anyway, and it's heavy jargon. I wish we had a nicer term for the minimal amount of "successful" result from a lazy computation, but I don't think such a term exists.
OTOH I don't think we can get away with avoiding "WHNF" entirely. I'd love to be able to point to a Nix-specific definition of it, because it does come up in discussions around the module system for example.

source/tutorials/nix-language.md
```diff
@@ -181,9 +182,9 @@ nix-instantiate --eval
 :::
 
 :::{note}
-The Nix language by default uses lazy evaluation, and will only compute values when needed.
+The Nix language uses lazy evaluation, and `nix-instantiate` by default does not compute everything.
```
fricklerhandwerk, line 185: Why is that different from the other section?

source/tutorials/nix-language.md
fricklerhandwerk: I like how you phrased the thing simply but correctly, but it also has this looming threat of being a rabbit hole. "What do you mean by 'more', *how much* more, why not everything?" I'd rather ignore this completely at this point. I understand the irresistable tendency to include more and more inconvenient truths into documentation, but that should really go to the Nix manual most people seem to be afraid of even touching.

fricklerhandwerk (approved):

fricklerhandwerk:
@roberth I guess you have to do the CLA dance, than I can merge the change.

fricklerhandwerk (approved):

das-g (approved):
LGTM

PR #274, opened 2022-07-25
NixOS configuration - testing, debugging, and organization

olafklingt:
The aim of this PR is to contribute a guide to the topic of `NixOS configuration testing debugging and organization`.
It revises https://nix.dev/tutorials/integration-testing-using-virtual-machines and utilizes additional information from nix wiki and manuals.
It might happen that the topic of configuration file organization is split into a separate guide.
This initial commit has the function to show that I am working on the topic.

PR #335, opened 2022-10-05
tutorial: NixOS configuration file organization

olafklingt:
split from PR #274 .
This PR is a draft and I don't expect it to be ready for review for a bit!


olafklingt:
I still think a tutorial or rather receipt, on how one would organize ones NixOS configuration files, is interesting. But it will have nothing to do with the content of this draft. So i close it.

PR #343, opened 2022-10-17
Nix language tutorial: fix typos, add notes

fricklerhandwerk:
Thanks to @jherland

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1491,6 +1491,8 @@ It is an error if the file system path does not exist.
 
 </details>
 
+For directories the same thing happens: The entire directory (including nested files and directories) is copied to the Nix store, and the evaluated string becomes the Nix store path of the directory.
+
 ### Fetchers
 
 Files to be used as build inputs do not have to come from the file system.
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1650,7 +1650,8 @@ Explanation:
   It returns an attribute set.
 - The argument must at least have the attributes `config` and `pkgs`, and may have more attributes.
 - The returned attribute set contains the attributes `imports` and `environment`.
-  `imports` is a list with one element: a path to a file next to this Nix file, called `hardware-configuration.nix`.
+- `imports` is a list with one element: a path to a file next to this Nix file, called `hardware-configuration.nix`.
+  When evaluated, the `./hardware-configuration.nix` will be copied into the Nix store, converting it into a Nix store path.
 
   :::{note}
   `imports` is not the impure built-in `import`, but a regular attribute name!
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1043,7 +1043,7 @@ x: x + 1
 
 The `<LAMBDA>` indicates the resulting value is an anonymous function.
 
-We can assign functions a name as to any other value.
+As with any other value, functions can be assigned to a name.
 
 Example:
 
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -852,7 +852,7 @@ Example:
 
     /absolute/path
 
-Paths are relative when they contain at least one slash (`/`) but to not start with one.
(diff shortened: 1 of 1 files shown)


source/tutorials/nix-language.md
```diff
@@ -905,6 +905,10 @@ Example:
 
     /current
 
+:::{note}
+Before you start combining file system paths and antiquotation, you probably want to read about the [path-related impurities below](path-impurities).
```
fricklerhandwerk, line 909: @jherland I'd like to keep the structure linear. Was there something specific that putting impure path behavior in front would have helped you with? Also, for reading the language we don't really care how things work once you use them, it's really just to understand what things are.

jherland:

source/tutorials/nix-language.md
jherland: Understood. I only stumbled into this by getting ahead of myself: I read the tutorial with a `nix repl` session open side-by-side, and after looking at antiquotation and then paths, I got curious about the difference between paths and strings. Specifically I remember thinking "so are paths and strings the same (like in Python 2), or are they separate (like with pathlib in Python 3)?" and next I tried to "string-ify" a path with antiquotation and all hell broke lose: 😉 

```nix
nix-repl> /dev/null
/dev/null

nix-repl> "${/dev/null}"
error (ignored): error: end of string reached
error: file '/dev/null' has an unsupported type

nix-repl> ./.
/home/jherland/nix-language-tutorial

nix-repl> "${./.}"
"/nix/store/0qg0178kca2qgdk83lz7f8zcjlw3sz95-nix-language-tutorial"

nix-repl> ../.
/home/jherland

nix-repl> "${../.}"
<HANGS AND MUST BE SIGKILLED!>
```

Most of this is explained later in the impurities section (incl. my proposed changes). But especially the last part was not a nice experience, and one that I can quickly see someone else running into as well.

I'm not sure what could be done in the tutorial to prevent this. As you say, the tutorial is about the language, and not really about the behavior, but we also do not want to discourage experimentation in `nix repl`...

fricklerhandwerk: Okay, I agree this is a pitfall. I'll make a proposal to reword to carry less of your specific context and be more generally informative.

source/tutorials/nix-language.md
```diff
@@ -1043,7 +1047,7 @@ x: x + 1
 
 The `<LAMBDA>` indicates the resulting value is an anonymous function.
 
-We can assign functions a name as to any other value.
+As with any other value, we can assign functions to a name.
```
fricklerhandwerk, line 1050:
```suggestion
As with any other value, functions can be assigned to a name.
```

source/tutorials/nix-language.md
```diff
@@ -1489,8 +1494,16 @@ The file is copied into the Nix store directory `/nix/store` as a side effect of
 
 It is an error if the file system path does not exist.
 
+:::{note}
+Note that the Nix language assumes that all input files will remain _unchanged_ during the course of the Nix expression evaluation.
+If you for example antiquote a file path during a `nix repl` session, and then later in the same session, after having changed the file contents, evaluate the antiquotation with the file path again, then Nix will still return the first store path.
+It will _not_ reread the file contents to produce a different Nix store path.
+:::
+
```
fricklerhandwerk, line 1502:
```suggestion
```

Sorry for the back and forth, but I thought about it again and came to the conclusion that this is too much of an implementation detail. This belongs to the Nix language reference. Would be great if you added your note [somewhere around here](https://github.com/NixOS/nix/blob/a324e9a5c84a144b824303064220463977c63c73/doc/manual/src/language/values.md?plain=1#L146-L151).

fricklerhandwerk (commented):
Thanks for the quick response, and sorry for the additional noise. Almost ready now. 👍

source/tutorials/nix-language.md
jherland: https://github.com/NixOS/nix/pull/7191 👍🏻

fricklerhandwerk:
@jherland thanks again for the fixups!

PR #353, opened 2022-10-22
Move contributing from tutorials to top-level

yukiisbored:

PR #360, opened 2022-10-22
Test how we can use CloudFlare Pages's redirects

zupo:
This commit needs to be merged to master to verify Cloudflare correctly picks it up. I will revert it after this is verified.

Needed for https://github.com/NixOS/nix.dev/pull/353


Makefile
```diff
@@ -51,6 +51,10 @@ clean:
 .PHONY: html
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+
+	# support for Cloudflare Pages' redirects
+	@cp _redirects build/html/_redirects
```
lucperkins, line 56: Instead of this, you can add the file directly to `source/_static` and Sphinx automatically adds it to the build output

lucperkins (changes requested):

Makefile
zupo: Oh, sweet, lemme try that.

zupo: Sadly, does not seem to work. At least on my local the file gets put into `_static/` output folder, not into root. 

Screenshot 2022-10-22 at 15 49 20

lucperkins: Never mind

zupo:
https://developers.cloudflare.com/pages/platform/redirects/ recommends changing the build command to something like `jekyll build && cp _redirects _site/_redirects` and we might have to do that. 

I asked @domenkozar to give me access to Cloudflare Pages.

zupo:
Turns out, the Cloudflare Pages build failure was random, re-running it fixed it. 

Also, no need to merge this, Cloudflare Pages preview is enough. I've verified the redirects works on a preview page.

```
➜  nix.dev git:(master) ✗ curl --head https://f80dea2a.nix-dot-dev.pages.dev/foo
HTTP/2 302 
```

Issue #345, opened 2022-10-21
Issue on page /tutorials/nix-language.html


Ten0:
Helloo!

Just heard about nix.dev today at NixCon, great resource, thanks a lot!

So I basically don't know anything about Nix (which is a nice test run for this doc I guess) and I started going through it.

I noticed that this page mentions `pkgs.lib` at https://nix.dev/tutorials/nix-language#pkgs-lib
before it is actually shown how `pkgs.lib` gets there: https://nix.dev/tutorials/nix-language#derivations
later in the page, which makes this first section a bit difficult to read.

I think it might be nice to add a "detailed explanation" section there that mentions that this is explained in section X below - or something else along those lines.

Thanks again!

domenkozar:
cc @fricklerhandwerk

fricklerhandwerk:
Hi @Ten0, thanks for the feedback. You're absolutely right, beginners taking the tutorials make for the best tests we can get.

Trying to understand what exact problem you had there. `pkgs.lib` are pure Nix language functions, they are not derivations. Since the concept of derivations is not relevant for that, I don't see a point of referring to them. Could you go back to that state of mind where you read the `pkgs.lib` section and try to identify what exactly confused you and what would have helped you avoid that confusion? Would it for instance help noting that those library functions are just for transforming pure values, while actually building things is done somewhere else?

Ten0:
So as I was reading this page I learnt about, in order:
1. [Search paths](https://nix.dev/tutorials/nix-language#search-path), and specifically `<nixpkgs/lib>`.
2. [pkgs.lib](https://nix.dev/tutorials/nix-language#pkgs-lib) "The `nixpkgs` repository contains an attribute set called `lib`", these functions are usually accessed through `pkgs.lib`.
3. [How we get `<nixpkgs>` into `pkgs`](https://nix.dev/tutorials/nix-language#derivations): `let pkgs = import <nixpkgs> {}; in "${pkgs.nix}`

So as I read 2, I wonder: "I thought we were importing nixpkgs through `<nixpkgs>` at some point, and suddenly it becomes a ~builtin `pkgs`? Or is that imported somehow? Where/when? Hmm something to remember to check for..."

And only as I read 3 (and the "Worked examples section that comes right after"), I get this information that it's just a convention to import `<nixpkgs>` in `pkgs` at the beginning of the files.

[...]

PR #386, opened 2022-11-08
Nix language tutorial: add explanation on how `pkgs.lib` is obtained

fricklerhandwerk:
Closes https://github.com/NixOS/nix.dev/issues/345

@Ten0 please take a look.

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1415,7 +1415,8 @@ Parentheses are required to separate function declaration from function applicat
 (pkgs-lib)=
 ### `pkgs.lib`
 
-The [`nixpkgs`][nixpkgs] repository contains an attribute set called [`lib`][nixpkgs-lib], which provides a large number of useful functions implemented in the Nix language.
+The [`nixpkgs`][nixpkgs] repository contains an attribute set called [`lib`][nixpkgs-lib], which provides a large number of useful functions.
+They are implemented in the Nix language, as opposed to [`builtins`](builtins), which are part of the language itself.
 
 :::{note}
 The Nixpkgs manual lists all [Nixpkgs library functions][nixpkgs-functions].
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1308,11 +1308,13 @@ We recommend to at least skim them to familiarise yourself with what is availabl
 
 [operators]: https://nixos.org/manual/nix/stable/language/operators.html
 
+(builtins)=
 ### `builtins`
 
 Also known as “primitive operations” or “primops”.
 
 Nix comes with many functions that are built into the language.
+They are implemented in C++ as part of the Nix language interpreter.
 
 :::{note}
 The Nix manual lists all [Built-in Functions][nix-builtins], and shows how to use them.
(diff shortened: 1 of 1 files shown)


Ten0 (approved):
Very clear, thanks ! :)

source/tutorials/nix-language.md
```diff
@@ -1414,22 +1412,117 @@ Parentheses are required to separate function declaration from function applicat
 
 </details>
 
+(pkgs-lib)=
 ### `pkgs.lib`
 
-The [`nixpkgs`][nixpkgs] repository contains an attribute set called [`lib`][nixpkgs-lib], which provides a large number of useful functions.
+The [`nixpkgs`][nixpkgs] repository contains an attribute set called [`lib`][nixpkgs-lib], which provides a large number of useful functions implemented in the Nix language.
```
Ten0, line 1418: I'm not sure if this adds clarity: it looks like this may be interpreted either as "they are implemented using the language" or "they are part of the language"... I think if we don't add anything here it's pretty clear that builtins are the exception and these are regular nix functions that were just explained.

fricklerhandwerk: That change was to address [this comment](https://github.com/NixOS/nix.dev/pull/267#issuecomment-1270076332). I didn't want to make a separate sentence, but probably it would be sensible here.

```suggestion
The [`nixpkgs`][nixpkgs] repository contains an attribute set called [`lib`][nixpkgs-lib], which provides a large number of useful functions.
They are implemented in the Nix language.
```

Ten0: How about "They are implemented in the Nix language", to get rid of that potential confusion ?

source/tutorials/nix-language.md
```diff
@@ -1414,22 +1412,117 @@ Parentheses are required to separate function declaration from function applicat
...
+let
+  nixpkgs = fetchTarball https://github.com/NixOS/nixpkgs/archive/3590f02e7d5760e52072c1a729ee2250b5560746.tar.gz;
+  pkgs = import nixpkgs {};
+in
+  pkgs.lib.strings.toUpper "always pin your sources"
+```
+
+    ALWAYS PIN YOUR SOURCES
+
+See [](pinning-nixpkgs) for details.
```
yukiisbored, line 1468: Missing text for link

source/tutorials/nix-language.md
```diff
@@ -1414,22 +1412,117 @@ Parentheses are required to separate function declaration from function applicat
...
+
+Example:
+
+```nix
+{ pkgs, lib, ... }:
+# ... multiple uses of `pkgs`
+# ... multiple uses of `lib`
+```
+
+For historical reasons, some of them are equivalent to [`builtins`](builtins) of the same name.
```
yukiisbored, line 1523: I think it's better to explicitly state what "them" is in this case. I'm not sure what this refers to myself. My educated guess is functions in `pkgs.lib`.

yukiisbored (changes requested):
Couple of minor stuff

source/tutorials/nix-language.md
fricklerhandwerk: This will be filled by the template engine to use the article title.

Issue #385, opened 2022-11-07
Integrate deploy previews


yukiisbored:
As nix.dev use Markdown with Sphinx, it would be helpful for reviewers to be able to see the changes proposed by a pull request without having to locally fetch the branch.

Since nix.dev uses Cloudflare pages, I believe it should be possible to have deploy previews.

cc @domenkozar

domenkozar:
Duplicates #348

PR #390, opened 2022-11-08
Make code blocks consistent

yukiisbored:
Illustration of shell sessions or terminals should use shell-session.

Entire expressions should be encapsulated in code blocks, not inline code.

### DIFF
--- a/source/faq.md
+++ b/source/faq.md
@@ -30,7 +30,7 @@ nix = {
 
 Using `Nix`:
 
-```bash
+```shell-session
 $ echo "trusted-binary-caches = https://hydra.snabb.co" >> /etc/nix/nix.conf
 $ nix-build helpers/bench.nix --option extra-binary-caches https://hydra.snabb.co
 ```
@@ -43,7 +43,7 @@ The default timeout for that is 1 hour as of writing.
 
 To wipe all cache-lookup-caches:
 
-```bash
+```shell-session
 $ rm $HOME/.cache/nix/binary-cache-v*.sqlite*
 ```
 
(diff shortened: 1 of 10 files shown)


fricklerhandwerk (changes requested):
In general I strongly agree that this should be made consistent, but I have a few general objections:

If we prepend $ to every command line, one has to manually select only the relevant part. Each code block has a copy button which IIRC take the whole string. Even if it's improbably smart for shell-session, those who for whatever reason don't find or use it will still have that problem. I would rather keep it consistent without the prompt. You'll notice that then many things stop working as they were originally layed out, such as the Nix shell prompts. I have no answer for that yet.

For the input output pairs I and quite certain: no, please keep them as in the Nix language tutorial. It has advantages for reading the source, and once highlighting works properly output can be nicely distinguished in rendering. See [this comment on a similar issue](https://github.com/NixOS/nixos-summer/pull/41#discussion_r953222745) for details.

Also I've seen people discuss it in other contexts and would like to adopt it for us to avoid descriptions of examples within code comments. Prose should be prose, code should be code, and there should be as little code as possible, because it's always a cognitive burden to read unfamiliar code.

Suggestion: change it to be consistent according to my proposal, and leave it as it is where it's not easy or obvious what to do.

yukiisbored:
@fricklerhandwerk 

> If we prepend $ to every command line, one has to manually select only the relevant part. 

`shell-session` makes the prompt unselectable as it uses the `gp` CSS class which sets `user-select` to `none`.

Here's a GIF which demonstrates me trying to highlight and copy-paste code from various snippets:


> Each code block has a copy button which IIRC take the whole string. Even if it's improbably smart for shell-session, those who for whatever reason don't find or use it will still have that problem.

The copy button with `shell-session` copies only the first line without the prompt and as demonstrated before, `shell-session` makes prompt unselectable.

> I would rather keep it consistent without the prompt. You'll notice that then many things stop working as they were originally layed out, such as the Nix shell prompts. I have no answer for that yet.

nix-shell prompts works fine as the lexer which detects the prompt takes that into account: https://pygments.org/docs/terminal-sessions/

If anything, this goes to show how https://github.com/NixOS/nix.dev/issues/385 is important as we don't use GitHub Flavored Markdown to render them.

fricklerhandwerk (approved):
@yukiisbored you convinced me, I shut up. 🙂

PR #376, opened 2022-10-27
Rewrite "Set up development environment"

yukiisbored:
The previous version is specific to Python as it uses `buildPythonApplication` instead of `mkShell`.

This one is more general but maintains the same Python Flask web application example.

PR #393, opened 2022-11-11
Restructure nix.dev: Tutorials are not guides

yukiisbored:
This PR contains a major change in nix.dev where the tutorials section is renamed to guides. 

Reading the articles under the tutorials section, I feel like most of them read more like guides than tutorials.

For those who aren't familiar with the difference between the two: https://diataxis.fr/tutorials-how-to/#tutorials-how-to

TL;DR: Tutorials are learning-oriented while Guides are task-oriented.

There are still some articles under this new section which reads like Tutorials but IMO, I think they need to be either removed or rewritten to guides. 

I think an ideal "tutorial" in nix.dev should look like a "Tour of Nix" where the reader is guided to a series of tasks which I've somewhat explained here: https://github.com/NixOS/nix.dev/pull/376#issuecomment-1293707625


yukiisbored:
Nix Pills could be considered as a "Tour of Nix" but the problem with Nix pills is it tries to teach which violates the rule that [tutorials should not teach] and only learn by doing.

[tutorials should not teach]: https://diataxis.fr/tutorials/#don-t-try-to-teach

fricklerhandwerk:
The Nix pills are IMHO closest to the Explanation category. Building things from first principles mostly helps with deeper understanding, not with building skills or solving problems. They would need some restructuring to be more focused at that, though.

source/reference/pinning-nixpkgs.md
```diff
@@ -1,35 +0,0 @@
-(ref-pinning-nixpkgs)=
-
-# Pinning Nixpkgs
```
fricklerhandwerk, line 3: We should not remove that without replacement, yet. I don't know of any good resource where the multiple ways specifying inputs without flakes is documented in one place.

fricklerhandwerk (approved):
Thanks! I think in general this is right, but the Nix language thing is quite clearly a tutorial and it would probably be not wrong to keep a tutorial category for skills to teach. I'd rather remove the templates section because that's stub.

source/reference/pinning-nixpkgs.md
yukiisbored: I think https://nix.dev/tutorials/towards-reproducibility-pinning-nixpkgs is a better article on this.

Not sure if we should add a section on `$NIX_PATH` however since I feel like that's not something we want someone to do.

yukiisbored: Though at least explaining how `import <nixpkgs>` magically resolves to a `nixpkgs` might be a good idea. Just not sure where to put it.

I guess "towards reproducibility pinning nixpkgs" can have an intro showing what the problem with that first so the reader understands why it's an issue in the first place.

fricklerhandwerk: Yeah, just listing that same at the right spot is fine, and that article is probably a good place.

yukiisbored:
Gonna close this in favor of something smaller.

PR #395, opened 2022-11-17
[Docs] [Tiny PR] Remove duplicated text from Nix Language Docs

bee-san:
Screenshot 2022-11-17 at 12 52 42

Absolute path is repeated here and I was a bit tired and tried to find the difference 🙈  This PR removes the duplicated documentation line

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -830,8 +830,6 @@ Example:
 /absolute/path
 ```
 
-    /absolute/path
-
 Paths are relative when they contain at least one slash (`/`) but do not start with one.
 They evaluate to the path relative to the file containing the expression.
 


bee-san:

fricklerhandwerk:
Hi @bee-san. Unfortunately the CSS doesn't show the difference (yet), but the point was to show that the expression evaluates to itself. This is opposed to, for instance, relative paths.

Instead changing the CSS to make clearly visible that the above is a Nix expression and the below is its evaluation result would be a great improvement!

yukiisbored:
I believe this is intentional as examples/demonstrations in the Nix language tutorial are always in a form of input and output. I think this should be more evident to avoid confusion for readers who are unaware of this convention.

fricklerhandwerk:
@bee-san closing this due to https://github.com/NixOS/nix.dev/pull/405, which should obviate the issue.

PR #400, opened 2022-11-24
glossary: import terms from nixionary.org

zimbatm:
Nix is using a bunch of esoteric terms like "closure" and "realization"
that can turn off newcomers. I'm assuming that having a good glossary
can help with that. I always like a good glossary to speed-run a new
project.

<https://nixionary.org> was a little project I had for NixCon, to
collect more of those words for the glossary.

### DIFF
--- a/source/glossary.md
+++ b/source/glossary.md
@@ -1,8 +1,42 @@
 # Glossary
 
 ```{glossary}
-attribute name
-    Attribute sets are a collection of key/value pairs. The attribute name is the key.
+Attribute sets
+    Attribute sets represent a sorted collection of name/value pairs in the Nix
+    language.
+
+Channels
+    Represents two things:
+    1. Releases of nixpkgs built and pushed by Hydra. See
+       <https://status.nixos.org>.
+    2. GC roots that are managed by `nix-channel`.
+
+Closure
+    The closure is the tree of the dependencies of a derivation, and the
+    dependencies of the depencencies recursively.
+
+Derivation
+    **Build recipe** described with Nix. It declares all the inputs and steps
+    that will be used for the build.
+
+Flakes
+    Flakes it to Nix what NPM is to NodeJS. It's a dependency manager and
+    central entry point to projects.
+
+GC root
+    A reference to a store entry that is used to protect it and all its
+    dependencies from the garbage collector. The reference is represented as a
+    sylink to a `/nix/store` entry and stored under `/nix/var/nix/gcroots/`.
+
+Hydra
+    **CI** system that is designed specifically for nixpkgs. The main instance
+    lives at <https://hydra.nixos.org>.
+
+Import From Derivation (IFD)
+    The practice of reading the output of a built derivation during the Nix
+    evaluation phase.
+    It allows moving computation from eval to build time but also breaks the
+    eval/build split in the process.
 
 Nix
     Build system and package manager.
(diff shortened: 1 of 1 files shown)


yukiisbored:
@zimbatm Yes, currently @domenkozar still holds the copyright for now.

This will be reassigned to the foundation in the future.

source/glossary.md
```diff
@@ -1,35 +1,95 @@
...
+    A git repository containing a collection of 80000+ packages. See
+    <https://nixpkgs.org>.
 
     Read /nɪks ˈpækɪʤɪz/ ("Nix packages").
 
 NixOS
-    Linux distribution based on Nix and Nixpkgs.
+    The name represents two things at the same time:
+
+    1. The whole project and community.
```
fricklerhandwerk, line 74: That's not true and a reason for large amounts of confusion.

source/glossary.md
```diff
@@ -1,35 +1,95 @@
...
+    Per user GC roots. Those are typically set when using `nix-env` or
+    `nixos-rebuild`.
+
 Nix file
     File (`.nix`) containing a Nix expression.
 
 Nixpkgs
-    Software distribution built with Nix.
+    A git repository containing a collection of 80000+ packages. See
+    <https://nixpkgs.org>.
```
fricklerhandwerk, line 67: Adding a link is good, more descriptions, I think, less so. Each project should just describe itself authoritatively, as a single source of truth. That description may change over time, no point tracking these things here.

source/glossary.md
```diff
@@ -1,35 +1,95 @@
...
+
+Channels
+    Represents two things:
+    1. Releases of nixpkgs built and pushed by Hydra. See
+       <https://status.nixos.org>.
+    2. GC roots that are managed by `nix-channel`.
+
+Closure
+    The closure is the tree of the dependencies of a derivation, and the
+    dependencies of the depencencies recursively.
```
fricklerhandwerk, line 16: These should go to the Nix manual's glossary.

source/glossary.md
zimbatm: What is the name of the project and community then? The entry point domain is nixos.org.

I agree that it's confusing, that's why it's worth documenting it.

zimbatm:

fricklerhandwerk:
I'm trying to add definitions for a part of these terms to the Nix manual:

https://github.com/NixOS/nix/pull/7318
https://github.com/NixOS/nix/pull/7332

[Substituter](https://nixos.org/manual/nix/unstable/glossary.html?highlight=substituter#gloss-substituter) already exists in the Nix manual' glossary.

I'm not sure about the change overall. It's probably good to have multiple entry points for terms. But maybe it's enough to link to the terms' authoritative definitions (in their respective manuals) when they are used in texts.

In any case we should not re-make the definition here, and rather link to the existing one. Sometimes we don't have (a good) one, like it was the case for the Nix language in the Nix manual, but that should be addressed first then.

yukiisbored:
I don't think it's a bad idea to have another glossary in nix.dev since I imagine they rarely change.

Though, I'm not opposed to linking the definition on the Nix manual.

zimbatm:
I'm happy if you incorporate my work in any way that you think is useful. If you want to push all the definitions to the respective projects and then transform this glossary into a meta glossary, that's fine.

Since you seem to be opinionated about it @fricklerhandwerk, I think the easiest would be for you to cannibalize my PR. That way we avoid too much back and forth.

fricklerhandwerk:
@zimbatm thanks, I appreciate the addition in general. Sorry, the tone here is usually quite to the point, so my comment was not to dismiss your effort. This is a strategic decision for the documentation team to make, I just have random opinions which are surely most often wrong.

zimbatm:
No worries, it makes sense to be opinionated for documentation, to give it a consistent feel.

Please have a chat with the documentation team and let me know once you agree on the high-level direction.

zimbatm:
oops, deleted my fork by mistake

fricklerhandwerk:
@NixOS/documentation-team

infinisil:
Imo the nix.dev glossary should ideally be the combined glossaries of [Nix itself](https://nixos.org/manual/nix/stable/glossary.html), nixpkgs and NixOS, plus all of the general terms not mentioned anywhere else (like Nixpkgs, NixOS, etc.)

I'm not sure how feasible this is though

zimbatm:
Imagine all the people who could have benefited from that Glossary, even if imperfect. It would be better than the current situation. And you can still move around later?

PR #405, opened 2022-12-03
Improve page styling to improve readability

yukiisbored:
This PR improves readability by manipulating the page styles via CSS along with markup changes.

Sample using [`let ... in ...` in Nix language tutorial](https://nix.dev/tutorials/nix-language.html#let-in) as a test subject:

Screenshot showcasing the section readbility improved via styling changes

Without the changes:

Screenshot showcasing an example without the changes which reduces readibility


yukiisbored:
In the future, I think we should enforce the use of [`code-block`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) directives as it allows further manipulation.

source/tutorials/nix-language.md
```diff
@@ -292,7 +304,7 @@ Nix language data types *without functions* work just like their counterparts in
   <th>JSON</th>
 </tr>
 <tr>
-<td>
+<td style="padding-right: 1em">
```
fricklerhandwerk, line 307: Won't this make the table asymmetric? I think that change is unrelated.

fricklerhandwerk (commented):
@domenkozar @lucperkins please take a look.

I think this is the exactly right direction. It makes the whole setup more verbose, but that's a price I'll gladly pay for the greatly improved readability.

source/tutorials/nix-language.md
yukiisbored: It doesn't, the web browser will realign the rest.

image

But yes, I forgot to make a separate commit for this change.

yukiisbored:
I recently found out [sphinx-design] which essentially implements the [dropdown] interface and add specific markup for it.

It also includes other components which would be useful for us. 

[sphinx-design]: https://sphinx-design.readthedocs.io/en/sbt-theme/index.html
[dropdown]: https://sphinx-design.readthedocs.io/en/sbt-theme/dropdowns.html

yukiisbored:
I made more changes to this PR.

I integrated sphinx-design which adds new directives.

Raw HTML `<details>` are replaced with `dropdown` directive, now details look this:

Details with visible dropdown

Raw HTML `<table>` are replaced with `grid` directive, now the comparisons look like this:

Comparison between Nix and JSON using grid

Currently, it looks like this:

Comparison between Nix and JSON using table

And I also changed the "Install Nix" page to use tabs instead of dedicated sections for each operating system:

Install Nix page with tabs to choose different operating systems, Linux is shown

Install Nix page same as before but Windows (WSL2) is shown

Now the page is less daunting for new users :)

lucperkins (approved):

PR #413, opened 2022-12-10
More details on first worked example

henrik-ch:
Details on the shell environment worked example.


fricklerhandwerk:
I thought about adding something along these lines when I wrote this example, but then refrained from it. Because then why don't we have the same instructions for the other examples. It should be fine to (even implicitly) leave that as an exercise for the reader, which you apparently have worked through. :) What would be useful and unobtrusive though is adding links on the terms "shell environment" "package" and "configuration module" which lead to the respective manuals (or tutorials if no good manual page exists yet).

PR #416, opened 2022-12-10
You have to run nixos-rebuild switch as well

mmolhoek:
small addition, otherwise people reboot and nothing changed


fricklerhandwerk (approved):
Good catch, thanks!

PR #422, opened 2022-12-17
Make install command in documentation compatible with fish shell

doubledup:
With the [fish shell support](https://github.com/NixOS/nix/pull/7014) added earlier this year, Nix supports installations in fish shell out of the box. The installation instructions still use process substitution syntax that's specific to other shells (eg. bash & zsh), so users installing Nix from fish shell must translate the command before the installation will work.

This PR makes the install commands in the installation instructions compatible with fish shell.

Running the current install command in fish produces
```fish
$ sh <(curl -L https://nixos.org/nix/install)
...
fish: Invalid redirection target: 
sh <(curl -L https://nixos.org/nix/install)
   ^
```

Running the command with the correct fish syntax starts the installer (also tested with the `--daemon` flag passed to the installer)
```fish
$ curl -L https://nixos.org/nix/install | sh
...
Switching to the Multi-user Installer
Welcome to the Multi-User Nix Installation
...
```

### DIFF
--- a/source/tutorials/install-nix.md
+++ b/source/tutorials/install-nix.md
@@ -7,19 +7,11 @@
 Install Nix via the recommended [multi-user installation]:
 
 ```shell-session
-$ sh <(curl -L https://nixos.org/nix/install) --daemon
+$ curl -L https://nixos.org/nix/install | sh -s -- --daemon
 ```
 
 :::{note}
 For security you may want to [verify the installation script] using GPG signatures.
-:::
-
-:::{note}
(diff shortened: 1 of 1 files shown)


doubledup:

fricklerhandwerk:
As [commented on your other PR](https://github.com/NixOS/nix/pull/7474#issuecomment-1359255569) we could just as well use

```console
curl -L https://nixos.org/nix/install | sh
```

doubledup:
@fricklerhandwerk That's much better, thanks! I've updated the PR and the description.

fricklerhandwerk (approved):

PR #427, opened 2023-01-17
working on #PR 325 - nix shell and shellbang

henrik-ch:
I hope that I am not committing a faux pas here.

This is basically the work from @rapenne-s with the comments and suggested changes from @fricklerhandwerk  and @n8henrie. The reason that I started a new PR was that I wasn't sure how to continue inside the #325 itself.


n8henrie:
I have no issue with you assuming charge of this PR. I would consider expanding more on the `--pure` part, as I think this is one of the most important flags one can use to ensure portability on non-NixOS systems.

henrik-ch:
Thank you for giving feedback and guidance here @n8henrie.

The info that I have found on the --pure parameter is on this webpage:
https://nixos.org/manual/nix/stable/command-ref/nix-shell.html
the screenshot pure section specifically:
image
Is that what you were thinking about? 
In that case, I could either link to it, or bring parts of it over to the tutorial page. 
What do you think is better?

If you know of a different/better source than that man page linked about, I am of course very happy to hear about that too.

fricklerhandwerk:
> In that case, I could either link to it, or bring parts of it over to the tutorial page.
What do you think is better?

I wasn't asked, but still: Definitely link, to unambiguously keep a single source of truth.

henrik-ch:
> I wasn't asked, but still: Definitely link, to unambiguously keep a single source of truth. In case the manual doesn't do a good-enough job, let's improve the manual.

I tried to follow this advice. Please let me know if you see any further changes on this.

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,76 @@
...
+- `-p` lists packages that should be present in the interpreter's environment
+- `-I` explicitly sets the search path for packages
+- `--pure` clears the environment before start - more details in the [nix-shell man page]: https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#options
+
+Create a file named `nixpkgs-releases.sh` with the following content:
+
+```shell
+#!/usr/bin/env nix-shell 
+#! nix-shell -i bash --pure
+#! nix-shell -p cacert curl jq python3Packages.xmljson
```
n8henrie, line 53: Please also include `-p bash`; while implicit in this case (it is included by something else), this omission in the prior tutorial led to me wasting a fair chunk of my time! See also: https://discourse.nixos.org/t/nix-shell-shebang-default-nix-no-such-file-or-directory/24246/2?u=n8henrie

EDIT: (Adding `-p bash` was the main reason I was tagged to contribute to this PR :) thanks for picking up and running with it!)

source/tutorials/nix-shell-in-shebang.md
```diff
@@ -0,0 +1,76 @@
...
+It takes a command name as argument and will run the first executable by that name it finds in the directories listed in the environment variable `$PATH`.
+
+We use [`nix-shell` as a shebang interpreter].
+It takes the following parameters relevant for our use case:
+
+[`nix-shell` as a shebang interpreter]: https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#use-as-a--interpreter
+- `-i` tells which program to use for interpreting the rest of the file
+- `-p` lists packages that should be present in the interpreter's environment
+- `-I` explicitly sets the search path for packages
+- `--pure` clears the environment before start - more details in the [nix-shell man page]: https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#options
```
n8henrie, line 46: Purely for stylistic reasons, I would remove `--pure` here and make a separate section for it. I think it is the single most useful insurance for portability in `nix-shell` (and shockingly missing from `nix shell`, whose `-i` flag doesn't clear `PATH`!). If you feel like this is too much effort or undesirable I'm happy to make a separate PR with this modification once this is merged.

I was basically envisioning this tutorial and script minus `--pure`, and then discussion and an example of how esp on non-NixOS systems your `PATH` is still being search for these binaries (that might not be present on someone else's system). Purity is the sine qua non of Nix IMO and I think deserves special attention for this *really important* use case!

Thanks for considering.

n8henrie (changes requested):

source/tutorials/nix-shell-in-shebang.md
henrik-ch: Thank you for the feedback Nathan, this is included now.

source/tutorials/nix-shell-in-shebang.md
henrik-ch: I can see the value in the suggestion Nathan. For me it's a difficult section to write as my understanding isn't that advanced yet, so I'd rather pass on including it for now and make a separate revision/pull request for it. 

I have a similar thing that I don't understand but could also be good to include: With the I-flag, how can I figure out what the eligible values are to put in for the search path? I am not in a position to write that either as I don't know the answer.

n8henrie: Sounds good, I'll make a separate PR with an example.

Sorry, you're referring to `-I explicitly sets the search path for packages`?

That wording should maybe be changed? I'm not that advanced either, but my understanding is that sets the search path for looking up e.g. `nixpkgs`, not for "packages" in the traditional sense.

For example, the first two of these give me the same binary of nixpkgs because using my user's default search path for `nixpkgs` is implicit in the first and explicit in the second. In the third, I override `nixpkgs` with a path to a locally checked out repository using a different version of nixpkgs (current master, recently pulled). In the fourth invocation, I pinned a specific version of `nixpkgs`, which should help ensure complete reproducibility.

So the packages you get depend entirely on the version of `nixpkgs` you're setting in your channel.

(Flakes are relevant here, as they are always explicit about the version of `nixpkgs` being used.)

```console
$ nix-shell --pure -p nix --command 'type -a nix; nix --version'
nix is /nix/store/gzd5wpvq9b5q5l3hvwpihgrhcrrllgi0-nix-2.11.1/bin/nix
nix (Nix) 2.11.1
$ nix-shell --pure -p nix -I nixpkgs=~/.nix-defexpr/channels/nixpkgs --command 'type -a nix; nix --version'
nix is /nix/store/gzd5wpvq9b5q5l3hvwpihgrhcrrllgi0-nix-2.11.1/bin/nix
nix (Nix) 2.11.1
$ nix-shell --pure -p nix -I nixpkgs=~/git/nixpkgs --command 'type -a nix; nix --version'
nix is /nix/store/gmknmx6vrdj8bzzfmyk9qfcalhsrpc3a-nix-2.12.0/bin/nix
nix (Nix) 2.12.0
$ nix-shell --pure -p nix -I nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/92a047a6c4d46a222e9c323ea85882d0a7a13af8.tar.gz --command 'type -a nix; nix --version' 
nix is /nix/store/w1xw3nsq03m2h8c0g23ps62lcbkdlc83-nix-2.0.4/bin/nix
nix (Nix) 2.0.4
```

For the record, if you want a list of all packages in that revision of `nixpkgs`, something like is sloppy but seems to work (I'm sure others can tell you a better way.

```console
$ nix-instantiate --eval -E 'with (import <nixpkgs> {}); lib.attrNames (import "${pkgs.path}/pkgs/top-level/all-packages.nix" { \inherit lib config overlays; noSysDirs = true; } {} {} {})'
$ # alternatively, with new-style syntax and json output
$ nix eval --impure --json --expr 'with (import <nixpkgs> {}); lib.attrNames (import "${pkgs.path}/pkgs/top-level/all-packages.nix" { inherit lib config overlays; noSysDirs = true; } {} {} {})' | jq
```

Or just `nix eval nixpkgs#<TAB><TAB>`.

henrik-ch: Thank you for sharing the examples above - this is really instructive. 

Two follow-up questions. My standard shell is zsh instead of bash.

For some reason, for example two in zsh I get back:
`warning: Nix search path entry '~/.nix-defexpr/channels/nixpkgs/' does not exist, ignoring`
whereas if I enter bash, and run the very same command, it works. 
It might not be a nix-shell question strictly, but I would love to know why it behaves differently in zsh and bash.

In the fourth example, how did you work out that that specific archive tar.gz file would exist? Can you browse these on github, or is there some other way? 

Thanks a lot!

henrik-ch:
I think this may be ready to be included now. 

I was thinking of adding instructions on how to find versions for pinning nixpkgs, but I will omit that as I believe it would fit better under source/reference/pinning-nixpkgs.md

fricklerhandwerk:
I'll get back to reviewing this once #325 is merged.

henrik-ch:
As the original PR has now been merged, this one can be closed.

fricklerhandwerk:
#325 is merged, please rebase. I try to make time a review next week, but no promises.

henrik-ch:
Took some extra steps to get a sensible merge to current master, but I think this should be okay now.

fricklerhandwerk (changes requested):
This still needs to update the [provenance of commands](https://github.com/NixOS/nix.dev/pull/427/files#diff-082fa701747fc51e6261b12263201b4462ba86eb784dd23383786cd9595cd347R65), since you added `bash` and `cacert`. Probably you should say that adding the interpreter package is required to make sure the exact right version is used.

Now that https://github.com/NixOS/nixos-search/pull/610 is merged we can actually tell people to just use search.nixos.org to find what they need.

We could add right after that line:

```
Use [search.nixos.org](https://search.nixos.org/packages) to find packages providing the program you need.
```

source/tutorials/reproducible-scripts.md
```diff
@@ -2,7 +2,7 @@
 
 # Reproducible interpreted scripts
 
-In this tutorial, you will learn how to use Nix to create and run reproducible interpreted scripts.
+In this tutorial, you will learn how to use Nix to create and run reproducible interpreted scripts with [shebang].
```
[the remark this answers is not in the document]
henrik-ch: I am thinking of this more as for the "human eye SEO" - i.e. when a user lands on the page they can quickly determine that the tutorial will help them figure out how to write a nix shebang line on their bash script.

I am happy to take your lead and go with your decision if you prefer it in or out.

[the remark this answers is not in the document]
fricklerhandwerk: Right, makes sense. How about this?

```suggestion
In this tutorial, you will learn how to use Nix to create and run reproducible interpreted scripts, also known as [shebang] scripts.
```

source/tutorials/reproducible-scripts.md
```diff
@@ -42,26 +42,30 @@ We use [`nix-shell` as a shebang interpreter].
 It takes the following parameters relevant for our use case:
 
 - `-i` tells which program to use for interpreting the rest of the file
+- `--pure` excludes most environment variables when the script is run
 - `-p` lists packages that should be present in the interpreter's environment
 - `-I` explicitly sets [the search path] for packages
 
+More details on the options can be found in the [`nix-shell` reference documentation](https://nixos.org/manual/nix/stable/command-ref/nix-shell.html#options).
+
+
```
fricklerhandwerk, line 51:
```suggestion
```

source/tutorials/reproducible-scripts.md
```diff
@@ -42,26 +42,30 @@ We use [`nix-shell` as a shebang interpreter].
...
 #! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/2a601aafdc5605a5133a2ca506a34a3a73377247.tar.gz
 
 curl https://github.com/NixOS/nixpkgs/releases.atom | xml2json | jq .
 ```
 
-We specify `bash` as the interpreter.
+We specify `bash` as the interpreter with the -i option and we include the `bash` package in the environment with the -p option. The -p inclusion of `bash` is necessary as we are setting the --pure option.
 
-The command `xml2json` is provided by the package `python3Packages.xmljson`, while the commands `jq` and `curl` are provided by packages of the same name.
+The command `xml2json` is provided by the package `python3Packages.xmljson`, while the commands `jq`, `curl`, and `cacert` are provided by packages of the same name. Use [search.nixos.org](https://search.nixos.org/packages) to find packages providing the program you need.
```
fricklerhandwerk, line 68:
```suggestion
With the `-p` option we specify the packages required for the script to run.
The command `xml2json` is provided by the package `python3Packages.xmljson`, while `bash`, `jq`, and `curl` are provided by packages of the same name. `cacert` must be present for SSL authentication to work. Use [search.nixos.org](https://search.nixos.org/packages) to find packages providing the program you need.
```

fricklerhandwerk (approved):
Very good, just minor suggestions on wording. Thanks a lot!

PR #430, opened 2023-01-18
Clarify that name should be set in builtins.path

### LINKED ISSUE #428: Reproducibility referencing top-level directory with ./.: Unclear if it's the use of builtin.path with name that is the fix
In https://nix.dev/anti-patterns/language#reproducibility-referencing-top-level-directory-with we see:

If someone builds the project in a differently named folder, they will get a different hash of the src and everything that depends on it.

> 
> A better way is to use builtins.path:
> ```nix
>  { pkgs ? import <nixpkgs> {}
>  }:
> 
>  pkgs.stdenv.mkDerivation {
>    name = "foobar";
> 
>    src = builtins.path { path = ./.; name = "myproject"; };
> }
> ```

It is unclear if just using `builtins.path` is sufficient or if `name` is needed for the solution.

kjeremy:
Fixes #428

### DIFF
--- a/source/anti-patterns/language.md
+++ b/source/anti-patterns/language.md
@@ -186,7 +186,7 @@ control of in the source code.
 If someone builds the project in a differently named folder, they will get a different hash of the
 `src` and everything that depends on it.
 
-A better way is to use `builtins.path`:
+A better way is to use `builtins.path` with the `name` attribute set to something fixed. This will derive the human readable portion of the store path from `name` instead of the working directory:
 
 ```nix
  { pkgs ? import <nixpkgs> {}


kjeremy:

kjeremy:
The test failure looks unrelated.

source/anti-patterns/language.md
```diff
@@ -186,7 +186,7 @@ control of in the source code.
 If someone builds the project in a differently named folder, they will get a different hash of the
 `src` and everything that depends on it.
 
-A better way is to use `builtins.path`:
+A better way is to use `builtins.path` with the `name` attribute set to something fixed. This will derive the human readable portion of the store path from `name` instead of the working directory:
```
fricklerhandwerk, line 189:
```suggestion
A better way is to use `builtins.path` with the `name` attribute set to something fixed.
This will derive the human readable portion of the store path from `name` instead of the working directory:
```

fricklerhandwerk:
Awesome!

PR #436, opened 2023-01-30
Rename antiquotation to string interpolation

iFreilicht:
This is in accordance with the official naming, which was [changed in Release 2.13](https://nixos.org/manual/nix/stable/release-notes/rl-2.13.html).

PR #440, opened 2023-02-07
refine wording around string interpolation

fricklerhandwerk:
instead of a belated review of #436 by @iFreilicht


iFreilicht (approved):
I know I don't have the power to permit the merge, but I would've happily accepted these additions to my PR. Thank you!

PR #453, opened 2023-02-23
Update nix-language.md

drupol:
Fix markdown link

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -104,7 +104,7 @@ The following example is a Nix expression adding two numbers:
 
 #### Interactive evaluation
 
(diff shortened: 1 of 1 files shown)


drupol:

fricklerhandwerk:
@drupol Thanks for the catch. Apparently I introduced more errors while fixing one. The intention was to use `[\`nix repl\`]` everywhere so one doesn't have to add the link reference separately. Would be lovely if you could fix that instead.

drupol:
Hello,

I think I just fixed them all.

Let me know if it's ok.

fricklerhandwerk:
Yes, thanks so much!

PR #462, opened 2023-03-05
Quote URLs in nix language tutorial

munyari:
Called out as an anti-pattern https://nix.dev/anti-patterns/language#unquoted-urls

This change is to make the docs more self-consistent, and promote good habits in the reader.


munyari:

PR #471, opened 2023-03-15
add draft for Google Season of Docs project proposal

fricklerhandwerk:
@olaf since the project application requires a public document to link to, we better start now


fricklerhandwerk:
I'll merge this so we have something to point to under a stable URL. We will continue adding updates as we go.

PR #477, opened 2023-03-17
first commit new tutorial

henrik-ch:
Connecting Bash to Nix and mkDerivation.

Please note, this content is originally from https://www.zombiezen.com/blog/2023/03/connecting-bash-to-nix/.

Before we merge and publish we will make sure that @zombiezen is okay with it.

Please treat this as a first draft. 
I anticipate that there may be changes needed to structure, grammar and content.

PR #480, opened 2023-03-19
V bash tutorial zombiezen

henrik-ch:
As we talked about in #477, we start from a verbatim version with gist version of the blog post as published by @zombiezen.

Thank you both @zombiezen and @fricklerhandwerk for giving input to this!


henrik-ch:
The reason for the immediate modifications after the verbatim version was getting  the html rendering to work.

fricklerhandwerk:
Oh you could have force pushed over the other PR, but fine. :)

fricklerhandwerk:
Discussed in the Nix documentation team meeting:

- @fricklerhandwerk: motivation for group review: share thought process and general approach to reviews
- @zmitchell: tutorials shouldn't read like a blog post
    - @fricklerhandwerk: this is a verbatim copy to be able to trace changes
- @fricklerhandwerk: this article is essentially [Nix Pills 7](https://nixos.org/guides/nix-pills/working-derivation.html), but somewhat simpler
- @zmitchell: what is the tutorial about? what's the learning goal
    - @fricklerhandwerk: cannot identify any, this is clearly an explanation, so tutorial is the wrong label
        - can we salvage something by shifting focus of the article?
            - @khaled: probably not worthwhile, would not ever work at that low level
                - valuable information, but not for the purpose of using Nix
- decision: do not follow right now, but continue compiling a more complete survey of what's out there
  - once we have more overview, we can determine what to focus on, and will probably return to this in order to unify the explanation on derivations.

PR #495, opened 2023-03-25
Add doc structure proposal

zmitchell:
This PR adds a proposed documentation structure. Notably it does not change the structure of the `nix.dev` site, that would happen in a separate PR pending approval by the documentation team. This PR is meant to be a means of async discussion on the structure of the documentation.

The outline in `documentation-structure.md` is intended to be the skeleton of the documentation structure and specifically does not include a list of articles we intend to write/migrate. Articles would go directly under each of the leaf-node entries in the outline in `documentation-structure.md` once we have an approved documentation structure.

### DIFF
--- a/maintainers/working_groups/learning_journey/README.md
+++ b/maintainers/working_groups/learning_journey/README.md
@@ -28,42 +28,9 @@ Membership is open to anyone that wants to contribute.
 Meetings will be held every week to ensure that progress can be made between meetings.
 
 ## Immediate projects
-### Decide on documentation structure
-This project will decide on a high level structure for the documentation so that contributors know the audience they are speaking to and where their contributions fit into the Nix documentation landscape.
 
-The documentation will follow the [Diataxis](https://diataxis.fr) framework to the degree that it makes sense. This project will produce a proposal to be submitted to the Documentation Team.
-
-The first draft of this structure is as follows:
-- Tutorials
-    - Installation of Nix
-    - Walking a user through their first derivation
-    - Packaging an existing project for Nix for the first time
-- How-To
-    - Packaging for specific languages
-    - Packaging idioms
-        - `callPackage`, "import from derivation", "fixed output derivation", etc
-    - Day-to-day development and workflows
-        - Integration with IDEs/editors
-        - Nix in CI
-        - Nix and NixOS deployments
-        - Building containers
-    - Rollbacks
-    - Contributing in various ways
-- Explanation
-    - Concepts
-    - What is a derivation?
-    - Cross-compilation as a first-class citizen
-        - Where to put your dependencies (`buildInputs`, `nativeBuildInputs`, etc)
-    - The Nix Store and the database
-    - Channels, profiles, etc
-    - Overlays, overrides, `follows`
-- Reference
-    - Command reference
-    - How store paths are computed
-    - Build phases
-    - How derivations work
-    - Nix language reference
-    - Flake schema
+### Establish structure of documentation
+In order to unblocked contributions, the first priority is deciding on a documentation structure. We should be able to direct new contributors to the exact place to put the materials they want to contribute and understand beforehand whether a contribution overlaps with existing documentation.
 
 ### Set standards for documentation pages
 In order to provide a level of consistent quality for the documentation, we should decide on the bare minimum standards for documentation pages.
--- a/maintainers/working_groups/learning_journey/README.md
+++ b/maintainers/working_groups/learning_journey/README.md
@@ -30,7 +30,7 @@ Meetings will be held every week to ensure that progress can be made between mee
 ## Immediate projects
 
 ### Establish structure of documentation
-In order to unblocked contributions, the first priority is deciding on a documentation structure. We should be able to direct new contributors to the exact place to put the materials they want to contribute and understand beforehand whether a contribution overlaps with existing documentation.
+In order to unblock contributions, the first priority is deciding on a documentation structure. We should be able to direct new contributors to the exact place to put the materials they want to contribute and understand beforehand whether a contribution overlaps with existing documentation.
 
 ### Set standards for documentation pages
 In order to provide a level of consistent quality for the documentation, we should decide on the bare minimum standards for documentation pages.
--- a/maintainers/working_groups/learning_journey/README.md
+++ b/maintainers/working_groups/learning_journey/README.md
@@ -30,7 +30,7 @@ Meetings will be held every week to ensure that progress can be made between mee
 ## Immediate projects
 
 ### Establish structure of documentation
-In order to unblock contributions, the first priority is deciding on a documentation structure. We should be able to direct new contributors to the exact place to put the materials they want to contribute and understand beforehand whether a contribution overlaps with existing documentation.
+In order to unblock contributions, the first priority is deciding on a documentation structure. We should be able to direct new contributors to the exact place to put the materials they want to contribute and understand beforehand whether a contribution overlaps with existing documentation. The documentation will be organized using the [Diataxis](https://diataxis.fr) framework, so that tutorials, how-to guides, explanatory materials, and reference materials are clearly separated.
 
 ### Set standards for documentation pages
 In order to provide a level of consistent quality for the documentation, we should decide on the bare minimum standards for documentation pages.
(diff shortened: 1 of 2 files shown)


maintainers/working_groups/learning_journey/documentation-structure.md
```diff
@@ -0,0 +1,24 @@
...
+- Explanation
+    - Nix store
+    - Cross-compilation
+    - Packaging conventions
+- Reference
+    - Nix language reference
+    - Commands reference
+    - Derivations
+    - Nix built-ins
+    - Nixpkgs lib
```
henrik-ch, line 24:
```suggestion
    - Nixpkgs lib
    - Necessary unix concepts and familiarity before starting with nix
```

Feel free to rephrase as you see fit, I hope and think that it's clear what is meant.

fricklerhandwerk, line 24: These should just be links to the reference manuals. They are self-contained and should be under curation of the code maintainers. That's how we do it right now and there is little reason (or possibility or capacity, in fact) to change that.

maintainers/working_groups/learning_journey/documentation-structure.md
```diff
@@ -0,0 +1,24 @@
...
+        - CI
+        - Operations
+        - Containers
+        - Shells
+    - Package management
+    - Contributing
+- Explanation
+    - Nix store
+    - Cross-compilation
+    - Packaging conventions
```
fricklerhandwerk, line 18:
```suggestion
    - Packaging conventions
    - Module system
```

It's known as NixOS module system, but it's by far not specific to NixOS.

maintainers/working_groups/learning_journey/documentation-structure.md
```diff
@@ -0,0 +1,24 @@
+- Tutorials
+    - Nix
+    - NixOS
```
fricklerhandwerk, line 3: This needs much more work on breaking down which skills people need to use the eco system effectively. Priming the group to think in terms of projects seems counterproductive. Tutorials and guides are that part where I suggest we should think in terms of skills and use cases, not software components.

fricklerhandwerk (commented):
This looks like a good start, but I'm convinced we have to be much more deliberate in how we think about the different genres. They are created and used in very different ways, and we have to sharpen our mental models not to lump them together.

maintainers/working_groups/learning_journey/documentation-structure.md
zmitchell: I was just imagining a flat directory of tutorials in the "correct" sequence. I agree there's work to be done figuring out what needs to exist, but that's a content question (that we've already started working on) not a structure question.

zmitchell:
Could you provide an example? It's not clear to me whether you're suggesting changes or suggesting that we keep this in mind going forward.

maintainers/working_groups/learning_journey/README.md
```diff
@@ -28,42 +28,9 @@ Membership is open to anyone that wants to contribute.
 Meetings will be held every week to ensure that progress can be made between meetings.
 
 ## Immediate projects
-### Decide on documentation structure
-This project will decide on a high level structure for the documentation so that contributors know the audience they are speaking to and where their contributions fit into the Nix documentation landscape.
 
-The documentation will follow the [Diataxis](https://diataxis.fr) framework to the degree that it makes sense. This project will produce a proposal to be submitted to the Documentation Team.
```
asymmetric, line 34: Why remove the reference to Diataxis?

asymmetric (commented):
IIUC the only conroversial part is that the specific tutorial examples have been removed, and replced with software components? (I agree that it made more sense before).

Otherwise this seems mostly a rearranging of things from one file to another.

Or am I failing at reviewing this?

maintainers/working_groups/learning_journey/README.md
zmitchell: Just an oversight. I removed the whole section to remove the outline and then wrote a little bit about the project. I'll add it back in.

zmitchell:
The outline didn't really belong in the README, it was always going to get removed. We're moving it now so while discussing and finalizing it. If everyone gives it a thumbs up then yes it's just moving it from one file to another, if not it will be changed wrt review comments, etc.

zmitchell:
It may contain more information to include which tutorials we're considering, but determining which tutorials we need is a content question and is out of scope for this PR.

We started putting together ideas for the first tutorials in our last WG meeting.

maintainers/working_groups/learning_journey/documentation-structure.md
fricklerhandwerk: Very probably there is no flat sequence, but it's a dependency graph of skills.

I just wonder what additional information this particular list item provides. Yes, "tutorials" will contain tutorials about stuff concerning Nix-related tooling.

fricklerhandwerk:
>  suggesting that we keep this in mind going forward?
> (referring to https://github.com/NixOS/nix.dev/pull/495#pullrequestreview-1357982198)

I suggest we should be very aware of this all the time when designing the structure and content specifics.

Examples:

People always arrive with a use case, and will usually look for a guide. Yesterday I had:

> How do I make reproducible a one-file Python script that uses a particular version of a dependency?

The guide would essentially have two items:

1. Put it in a `nix-shell` shebang script
  - Learn here how to do that: https://nix.dev/tutorials/reproducible-scripts
2. Declare package dependencies for Python, pinning the correct Nixpkgs version
  - Learn here how to declare Python dependencies: TBD
  - Learn here how to find the Nixpkgs commit with your library version: TBD
  
Now we have three completely independent skills that are required for fulfilling that use case, and the Python dependencies thing will itself depend on being able to deal with the Nix language. But the tutorials don't need to know anything about the guides, they can live in isolation, or at best – optionally – refer to guides for example applications or exercises.

maintainers/working_groups/learning_journey/documentation-structure.md
fricklerhandwerk: Also covering that is clearly out of scope, but mentioning it in the right places as an "external dependency" is definitely worthwhile.

maintainers/working_groups/learning_journey/documentation-structure.md
fricklerhandwerk: Right, NixOS is both a specific way of using Nix as well as an advanced use case for Nix. In my outline from last year I lumped NixOS together with Home Manager and nix-darwin under "declarative configuration management" as they all work almost the same, but for slightly different scopes.

zmitchell: That's a good idea, I'll make an update to include a section along those lines.

zmitchell:
I've tweaked the tutorial section a little bit to separate the curated, walkthrough-style tutorials that take a new user through their first interactions with Nix. I think it's too early to predict the structure that the rest of the tutorials will take, but I think we can easily divide up this "walkthrough" section into at least a Nix and a NixOS/home-manager/etc section since these are targeting different use cases. The names of the not-walkthrough sections need bikeshedding since they're intended to be a grab-bag of sorts until a natural organization emerges on its own.

erooke:
Looks good to me. I think the trick with the tutorial organization is going to be following a less is more approach. Pick one compelling story from each of the major use cases (so two to three total) and focus on making them, to quote the diataxis, "meaningful and complete".

I guess what I'm saying is I don't think we should be splitting the tutorials up into "first steps" and "workflows". Tutorials should be monolithic and unambiguous start points focused on a compelling end result. Or at least that's my reading of the diataxis. I have more thoughts on the tutorials but they start to run into the content of the tutorials which is out of scope.

> we have to be much more deliberate in how we think about the different genres. They are created and used in very different ways, and we have to sharpen our mental models not to lump them together.

Agreed 100%. It is especially difficult for me ([and maybe others](https://diataxis.fr/tutorials-how-to/#tutorials-how-to))to not conflate tutorials and how-to's.

maintainers/working_groups/learning_journey/documentation-structure.md
henrik-ch: Yes, for me this goes in the same direction as what @fricklerhandwerk mentions here below about the reference manuals. 

> These should just be links to the reference manuals. They are self-contained and should be under curation of the code maintainers. That's how we do it right now and there is little reason (or possibility or capacity, in fact) to change that.

It's definitely out of scope for us to write this. 
What may be in scope for us is to link to resources that we think do a good job curating and delivering this information.  (in a similar manner and for similar reasons that we link to the Nix reference manuals)

PR #502, opened 2023-03-30
Add missing trailing newline in multi-line string example

niklasmohrin:
I was just checking out the [Nix language basics](https://nixos.org/guides/nix-language.html) and found the example of multi-line strings a bit odd as it exhibits different (and weirder) behavior than Python `"""` strings and is inconsistent with the second example. Checking with the NixOS Virtualbox image shows that in reality, the behavior is as expected and this seems to be a typo :)

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1033,7 +1033,7 @@ string
 
 ```{code-block}
 :class: value
-"multi\nline\nstring"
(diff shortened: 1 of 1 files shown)


niklasmohrin:
I don't think I have any copyright here due to minority, if possible I would prefer not to "sign" your document. If not, please feel free to reproduce the fix yourself in another PR

PR #506, opened 2023-04-01
Update nix-language.md

lxnn:
I added information about what `import` does if you give it the path to a directory rather than a file.

I think, without this information, the meaning of `import <nixpkgs>` in the next section isn't clear. I was a little bit confused by it.

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1545,7 +1545,7 @@ builtins.toString
 Most built-in functions are only accessible through `builtins`.
 A notable exception is `import`, which is also available at the top level.
 
-`import` takes a path to a Nix file, reads it to evaluate the contained Nix expression, and returns the resulting value.
+`import` takes a path to a Nix file, reads it to evaluate the contained Nix expression, and returns the resulting value. If the path is a directory, the file `default.nix` in that directory is read.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
It's much more complicated than that: https://nixos.org/manual/nix/unstable/language/builtins.html#builtins-import

But this is not the place to unwrap that complexity. The question is why you got confused: did you know otherwise beforehand, or did it mislead you into believing wrong things?

infinisil:
I don't think this has to be complicated. This is a small change pointing out something non-obvious but often-used.

source/tutorials/nix-language.md
```diff
@@ -1545,7 +1545,7 @@ builtins.toString
 Most built-in functions are only accessible through `builtins`.
 A notable exception is `import`, which is also available at the top level.
 
-`import` takes a path to a Nix file, reads it to evaluate the contained Nix expression, and returns the resulting value.
+`import` takes a path to a Nix file, reads it to evaluate the contained Nix expression, and returns the resulting value. If the path is a directory, the file `default.nix` in that directory is read.
```
infinisil, line 1548:
```suggestion
`import` takes a path to a Nix file, reads it to evaluate the contained Nix expression, and returns the resulting value.
If the path points to a directory, the file `default.nix` in that directory is used instead.
```

infinisil (changes requested):

PR #515, opened 2023-04-07
Switch from poetry to nixpkgs

zmitchell:
This PR removes the `poetry` based workflow in favor of a nixpkgs based workflow. In doing so the Python version will be updated from 3.7 to 3.10. This also necessitates upgrading `sphinx` to v5.1.1. The version of `sphinx-book-theme` in nixpkgs is v0.3.3, which is not compatible with `sphinx` v5.x. The latest version of `sphinx-book-theme` is v1.0.1, which is compatible with `sphinx` v5.x, but it's not in nixpkgs.


flake.nix
```diff
@@ -1,61 +1,36 @@
...
-            old: {
-              buildInputs = (old.buildInputs or [ ]) ++ [ self.flit-core ];
-            }
-          );
+      in {
+        defaultPackage = pkgs.stdenv.mkDerivation {
+          name = "nix-dev";
+          src = self;
+          buildInputs = with pkgs.python310Packages; [
+            livereload
```
yukiisbored, line 18: `livereload` is only used for local development as it is a tool which invokes a build command every time a change is detected.

flake.nix
```diff
@@ -1,61 +1,36 @@
...
+            myst-parser
+            sphinx
+            sphinx-book-theme
+            sphinx-copybutton
+            sphinx-design
+            black
```
yukiisbored, line 24: `black` is also a development only tool as it is a Python linter.

flake.nix
```diff
@@ -1,61 +1,36 @@
...
-          );
+      in {
+        defaultPackage = pkgs.stdenv.mkDerivation {
+          name = "nix-dev";
+          src = self;
+          buildInputs = with pkgs.python310Packages; [
+            livereload
+            myst-parser
+            sphinx
+            sphinx-book-theme
```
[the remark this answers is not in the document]
fricklerhandwerk: Yes, something so seemingly simple as changing the version of a package is quite involved and requires a good understanding of the Nix language. That's because Nixpkgs' notion of a package currently is "function from packages (i.e., dependencies) to a (Nix language level) derivation with additional attributes". There is no way to declare dependencies and parameters, such as the version, distinctly in these function arguments, and some packages lump them together, some only allow passing dependencies. See this [Discourse thread for a discussion](https://discourse.nixos.org/t/proposal-move-values-from-derivation-attributes-to-function-arguments/20697/8).

`override` lets you re-evaluate the package function with new arguments (too coarse), `overrideAttrs` let's you re-evaluate the inner `mkDerivation` call with new arguments (too involved: fields were traditionally interdependent using `rec`, compounded by source fetching often needing extra infrastructure). Only somewhat recently we got a new interface thanks to @roberth that allows more granular use of `overrideAttrs`, but that's still dark magic underneath and by far not as simple for the user as setting configuration parameters on NixOS modules.

I have high hopes in the Nixpkgs architecture team around @infinisil to develop a more consistent, convenient, and easier to understand package paradigm, such as modules.

flake.nix
zmitchell: Are you suggesting I do something differently?

yukiisbored: @zmitchell Yeah, I'm just saying it's not necessary to be inside `buildInputs` since it is not a dependency required for building the website.

zmitchell: Ok, that part I get. What I don't get is where `black` and `livereload` are supposed to go instead. The `live` script relies on `livereload` being available for the call to `nix-shell --run`, and since I'm new to Nix I'm not sure the best way to achieve the same thing if those Python modules aren't in `buildInputs`.

fricklerhandwerk:
Discussed in the Nix documentation team meeting:

- @zmitchell: How to run the `./live` script with flakes?

flake.nix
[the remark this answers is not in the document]
FRidh: Python packages should always use `.override` and/or `.overridePythonAttrs`. If you use `.overrideAttrs` or `.overrideDerivation` instead of `.overridePythonAttrs` you go one respectively two layers "too deep" bypassing the `buildPython*` and/or `stdenv.mkDerivation` routines.

The Python section in the Nixpkgs manual gives plenty of examples on how to override a package.

flake.nix
```diff
@@ -1,61 +1,36 @@
...
-          sphinx-design = super.sphinx-design.overridePythonAttrs (
-            old: {
-              buildInputs = (old.buildInputs or [ ]) ++ [ self.flit-core ];
-            }
-          );
+      in {
+        defaultPackage = pkgs.stdenv.mkDerivation {
+          name = "nix-dev";
+          src = self;
+          buildInputs = with pkgs.python310Packages; [
```
FRidh, line 17:
```suggestion
          nativeBuildInputs = with pkgs.python310Packages; [
```

flake.nix
```diff
@@ -1,61 +1,36 @@
...
+            sphinx-design
+            black
+          ];
+          buildPhase = ''
+            make html
+          '';
+          installPhase = ''
+            mkdir -p $out
+            cp -R build/html/* $out/
+          '';
```
FRidh, line 32: See also the `sphinxHook` in the Nixpkgs manual, maybe it is of interest. I don't think it provides too much value here though.

FRidh (changes requested):

flake.nix
zmitchell: Can you explain the motivation here? I gather that (native|propagated)BuildInputs and buildInputs are rooted in cross compilation, but this distinction is lost on me in the context of Python.

FRidh: Simply correctness; native build inputs are your tools. Yes, outside cross-compilation it is less valuable however for Python it for sure still matters. Remember, there is a lot of native code used by Python modules.

Note `buildPython*` makes the distinction (it sets `strictDeps = true;`).

flake.nix
FRidh: A good approach is to have the main derivation describing only the tools and libraries it needs to build, and then have a separate derivation corresponding to a development shell where you add additional tools.

flake.nix
zmitchell: Let me rephrase the question to hone in on the part that confuses me: in general and in the Python specific case, when should I put packages in buildInputs, nativeBuildInputs, or propagatedBuildInputs?

FRidh: - `nativeBuildInputs`: these are build-time tools, that is, programs that run *during* the build (hence they need to be native)
- `buildInputs`: these are headers/libraries which one needs during build time, e.g. to link against, but it is not executed. These are thus your run-time dependencies.
- `propagatedBuildInputs`: these are `buildInputs` but that are propagated, that is, whoever adds your derivation in a `buildInputs`, will automatically get all these `propagatedBuildInputs` as well. And this is done recursively.

The Nixpkgs manual describes this in more detail.

In `buildPython*` we use `propagatedBuildInputs` for our run-time dependencies, which in a way is in line with what I wrote above, but it has its issues. It's a historic artifact and unfortunately a complete pain to fix.

flake.nix
```diff
@@ -3,82 +3,55 @@
...
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs {
           inherit system;
-          overlays = [ poetry2nix.overlay ];
+          overlays = [
+            # Add sphinx-sitemap from an overlay until
+            # it becomes available from nixpkgs-unstable
+            (import ./overlay.nix)
+          ];
```
alejandrosame, line 16: sphinx-sitemap PR: https://github.com/NixOS/nixpkgs/pull/260274

alejandrosame:
PR to add `sphinx-sitemap` to nixpkgs: https://github.com/NixOS/nixpkgs/pull/260274

Issue #498, opened 2023-03-30
Learning Journey tutorial - "nix-shell -p"


zmitchell:
The Learning Journey WG identified that the first tutorial in the learning journey should be a tutorial walking the user through creating ephemeral shell environments with `nix-shell -p`. This is the tracking issue for discussion on this tutorial and related work.

- [x] Compile a survey of `nix-shell` tutorials (@erooke)
- [x] Determine which ground we want to cover in the tutorial
    - Do we want to cover just `nix-shell -p`, or do we want to end with creating a `shell.nix`?
- [ ] Prepare an outline of the tutorial
- [ ] Prepare a draft of the tutorial
- [ ] Submit the draft as a PR

## Tutorial contents
- Install Python via `nix-shell -p python310`
- Illustrate that the version of Python outside `nix-shell` differs from the version of Python inside `nix-shell`
- Illustrate that once you leave `nix-shell` you no longer have access to the Python 3.10 that you had inside the shell
- Demonstrate how to run a single command inside an ephemeral shell with `nix-shell -p python310 --run python3 --version` 
- Explain that using `nix-shell -p <pkgs>` is useful for experimentation

## How
- Where to put the result: A new file in https://github.com/NixOS/nix.dev/tree/master/source/tutorials
- Person to contact for help:
  - Technical: @infinisil
  - Editorial: @zmitchell

erooke:
Here is a dump of all the notable resources I found on nix-shell last week along with a brief summary trying to categorize it according to the diataxis.

## Unofficial
- [ghedam](http://ghedam.at/15978/an-introduction-to-nix-shell)
  Introduces `nix-shell -p`, shows how it modifies the env, builds a `shell.nix` with a `shellHook`. Includes examples for python and rust. Pins dependencies and attempts to create a cross platform `shell.nix`. Somewhere between a tutorial and a how to guide.
- [inria](https://nix-tutorial.gitlabpages.inria.fr/nix-tutorial/first-package.html#diving-into-environments-nix-shell
) 
  Uses `nix-shell --pure` to test compiling a package in a derivation. Moves on to defining a dev environment inside a `shell.nix`. Tutorial.
  

- [Xebia functional (Video)](https://youtu.be/eW8KU6h_ZNo?t=448)
  Introduction to nix in general. Installs nix, uses `nix-env` to install a package, uses `nix-shell` to run a package  not installed. Tutorial.

[...]

PR #525, opened 2023-04-20
Useful programs in ad-hoc-developer-environments

erooke:
As per #498 this moves the tutorial for ad-hoc-developer-environments to use gh and jq to list out the issues on nix.dev instead of cowsay and lolcat.

As an example it displays the last 5 issues numbers, title, and author on our own repo.

### DIFF
--- a/source/tutorials/ad-hoc-developer-environments.md
+++ b/source/tutorials/ad-hoc-developer-environments.md
@@ -12,46 +12,74 @@ You can also share the command invoking such a shell with others, and it will wo
 
 Once you {ref}`install Nix <install-nix>`, you can use it to create new *shell environments* with programs that you want to use.
 
-In this section you will run two exotic programs called `cowsay` and `lolcat` that you will probably not have installed on your machine:
+In this section you will run two programs called `gh` and `jq` that you may not have installed on your machine:
 
 ```shell-session
-$ cowsay no can do
-The program ‘cowsay’ is currently not installed.
+$ gh api --method GET /repos/nixos/nix.dev/issues
+The program ‘gh’ is currently not installed.
 
-$ echo no chance | lolcat
-The program ‘lolcat’ is currently not installed.
+$ echo "{}" | jq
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
I know I'm blocking your progress here, and I'm genuinely sorry for being both slow with my intermittent availability and annoying with my critical feedback:

I strongly agree with the general direction that the sample programs should lead to something practical and be chosen such that they can be continued to be used in follow-up tutorials.

But this particular change obscures the lesson with lots of visual noise and inherent but needless complexity. This tutorial was traditionally a vehicle to show off, and the last changes had the purpose to emphasize the essential skill of using `nix-shell -p` productively. Specifically, the lesson idea is to teach how to:
1. Make programs available in a temporary environment
2. Find package names for the programs one is interested in
3. Increase reproducibility

The whole thing is intended to produce at least a slight surprise effect in order to continue the tradition of showing off: pick literally any ever-so-obscure program and it will work.

Everything in that document is serving these purposes. The examples are chosen deliberately simple to illustrate the principle. Intricate program invocations distract from that. I'm all in favor to transition to more practical use cases, and that's the secondary objective of the package search example. The primary objective is to make clear the distinction of package name and program name.

I'm all in favor of transitioning the narrative towards a more practical use case, and suggest to rather use *an additional exercise* for that purpose, since the tutorial currently lacks those completely. Exercises also allow space to introduce side concepts without taking away attention from the core lesson.

infinisil:
@fricklerhandwerk Your answers can be very intimidating! I'd encourage you to keep it as short as possible. I think it can be summarised with

> The main point of the tutorial is to teach `nix-shell -p`. Any non-trivial programs would distract from that.

In particular, the `gh` CLI even requires authentication with a GitHub account, so it's much more involved than it seems.

I generally agree with this, so let's close this. Thanks for the PR anyways @erooke, I think this taught us an idea regarding how to write tutorials :smile:

PR #530, opened 2023-05-06
Update nix-language.md

alper:
Probably have to do a rename cross-site, but a page this long can't be considered "basics".

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1,6 +1,6 @@
 (reading-nix-language)=
 
-# Nix language basics
+# Nix language primer
 
 The Nix language is used to declare packages and configurations to be built by [Nix][nix-manual].
 


fricklerhandwerk:
I'm not sure how this is an improvement. The new title doesn't really say something fundamentally different than the existing one. You're arguing the amount of material is too much for fundamentals. Smaller units are indeed better for learning. Therefore, what would you cut and move into a different piece without producing holes or leaving the learner with partial knowledge that cannot be used for anything practical?

Because the main purpose of this tutorial, as it says like three times in the text, is to enable learners to read und understand the structure of examples and typical code in the wild. If you can find a smaller subset of the language fulfilling that goal, I'm all for adopting the corresponding change to the tutorial.

alper:
The length is fine but I expect a page called "basics" to be less than 25% of that length.

zmitchell:
I agree that the page is long, but I don't think "primer" is an improvement. "Basics" and "primer" mean essentially the same thing to me. If you're really looking to change the name I think "Nix language guide" would work just fine.

fricklerhandwerk:
In that case I'd like to keep "basics" simply to have a distinctive name from all the guides out there.

PR #532, opened 2023-05-06
Fix link to renamed anchor

zmitchell:
This PR fixes a link that was broken when an anchor in the `cachix/install-nix-action` README was changed.


zmitchell:
To be clear, this link works perfectly fine. It's just `linkcheck` that fails on it.

PR #540, opened 2023-05-11
Add more templates

CobaltCause:
Alternative to https://github.com/NixOS/nix.dev/pull/404 in the spirit of https://github.com/NixOS/nix.dev/pull/404#issuecomment-1335090971. Primarily motivated by [this message][0] and [this message][1] in [#nix:nixos.org](https://matrix.to/#/#nix:nixos.org). I'm not sure whether this is the right place to put this kind of thing, but I suppose I'll find out by posting this PR :sweat_smile:

I've been maintaining these two project templates for quite a while with what I believe are best practices for both the ecosystem each is for and their integration with Nix. It took a considerable amount of time (and is still ongoing) to gain the necessary experience, read the relevant documentation, and discern the current ecosystem direction to reach the point these are currently at[^1].

I've allegedly saved a handful of people a substantial amount of headache by having these available for use/reference, so increasing their discoverability will likely be very helpful for others. I hadn't considered trying to document these templates anywhere myself before since that kinda feels like self advertising, but the messages linked above suggest some actual desire from others for this to happen, so here we are.

[0]: https://matrix.to/#/!KqkRjyTEzAGRiZFBYT:nixos.org/$CiHJLK8NGecC7Lrml10anZeQTUF2MjIPv-TM3Sk0zh4?via=nixos.org&via=matrix.org&via=tchncs.de
[1]: https://matrix.to/#/!KqkRjyTEzAGRiZFBYT:nixos.org/$k2JVzNaw3veTUGRJurTDFd1MKx7n2p4OeFZzrbvxqoc?via=nixos.org&via=matrix.org&via=tchncs.de

[^1]: Especially for Python...

### DIFF
--- a/source/templates/index.md
+++ b/source/templates/index.md
@@ -1,6 +1,14 @@
 # Templates
 
-## Getting started Nix template
+## Generic
 
-[Getting started Nix template](https://github.com/nix-dot-dev/getting-started-nix-template)
-gives you best practices learned for setting up any kind of development project using Nix.
+### [Getting started Nix template][getting-started-nix-template]
+
+Provides an example baseline development environment using
+[`devenv.sh`][devenv].
+
+[getting-started-nix-template]: https://github.com/nix-dot-dev/getting-started-nix-template
+
+<!-- General links -->
+
+[devenv]: https://devenv.sh
--- a/source/templates/index.md
+++ b/source/templates/index.md
@@ -9,6 +9,33 @@ Provides an example baseline development environment using
 
 [getting-started-nix-template]: https://github.com/nix-dot-dev/getting-started-nix-template
 
+## Python
+
+### [Nix + Python Quickstart][nix-python-quickstart]
+
+A [`cargo-generate`][cargo-generate] project template for use with Flakes,
+[`direnv`][direnv], and [`nix-direnv`][nix-direnv]. Generated projects are
+fully-featured but not tailored for any specific use-case, including things such
+as a development environment, a package expression, and continuous integration
+tooling.
+
+[nix-python-quickstart]: https://or.computer.surgery/charles/nix-python-quickstart
+
+## Rust
+
+### [Nix + Rust Quickstart][nix-rust-quickstart]
+
+A [`cargo-generate`][cargo-generate] project template for use with Flakes,
+[`direnv`][direnv], and [`nix-direnv`][nix-direnv]. Generated projects are
+fully-featured but not tailored for any specific use-case, including things such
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
I don't and won't have the capacity to maintain any of this if it were adopted, so take my feedback with a large grain of salt.

1. I'm strongly inclined to vouch for those templates being maintained by nix.dev maintainers. This is now the official place for learning and onboarding Nix (not yet visible as such, but we'll get there), and in my opinion we should have control over what is presented and be very deliberate about the why and how.

  The alternative would be linking to third-party materials kitchen-sink style, but then probably [awesome-nix](https://github.com/nix-community/awesome-nix) would be a better place – also for the existing template.

2. Those templates look really involved to me. Since teaching essential skills and a firm understanding of basic principles in the Nix ecosystem is currently the documentation team's main objective, stacking tooling on top of each other that way feels off-target.

   Yes, that's what Nix is really good for, and just providing people with a ton of scaffolding surely has its uses. But in my opinion it's not very valuable didactically, especially to those who are still struggling to grasp the various layers of abstraction involved.
   
@NixOS/documentation-team what do you think? I can of course merge it if someone wants to take responsibility for it.

CobaltCause:
It's alright, I understand, and I appreciate the consideration!

> But in my opinion it's not very valuable didactically, especially to those who are still struggling to grasp the various layers of abstraction involved.

I think this is a point of confusion on my part; I didn't realize this was the intended target audience of the templates section. I thought it was only supposed to be "a collection of templates that provide a good starting point for various circumstances", not necessarily targeted at a given skill level.

Thanks again for taking a look at this; I'm curious to see what others have to say as well. In the meantime, I'll try submitting these to awesome-nix like you suggest and maybe https://nixos.wiki since that's also unofficial.

fricklerhandwerk:
Adding it to NixOS Wiki is always an option, and probably more likely to be found in frenetic internet searches across a thousand tabs. There is a large population of users that operates this way.

zmitchell:
I don't think it's the documentation team's responsibility to maintain starter templates. In order to do that we would need some infrastructure to know the instant the templates break due to changes in whatever language they're supporting. Fixing that requires people with knowledge of each language, etc. You can see how this gets out of hand pretty quickly. We don't have the bandwidth to do this right now.

I _do_ think it's our responsibility to point people to where templates can be found. Maybe this is something the Nix team/community can maintain and we can point at because I do think it's very valuable to be able to tell a new user "just do this to get started, don't worry about the details if you don't want to."

PR #545, opened 2023-05-12
FAQ: what are flakes

fricklerhandwerk:
this is indeed a frequently asked question.

the answer is an attempt to clarify concepts and link to the relevant reference documentation without getting into too much technical detail.

### DIFF
--- a/source/recipes/faq.md
+++ b/source/recipes/faq.md
@@ -2,6 +2,44 @@
 
 ## Nix
 
+### What are flakes?
+
+What is usually referred to as "flakes" is:
+- A policy for managing dependencies between {term}`Nix expression`s.
+- An [experimental feature] in Nix, implementing that policy and supporting functionality.
+
+[experimental feature]: https://nixos.org/manual/nix/stable/contributing/experimental-features.html
+
+Technically, a [flake](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#description) is a file system tree that contains a {term}`Nix file` named `flake.nix` in its root directory.
+
+Flakes add the following behavior to Nix:
+
+1. A `flake.nix` file offers a uniform [schema](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html#flake-format), where:
+   - Other flakes can be referenced as dependencies providing {term}`Nix language` code or other files.
+   - The values produced by the {term}`Nix expression` in `flake.nix` are structured according to pre-defined use cases.
+
+1. References to other flakes can be specified using a dedicated [URL-like syntax](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html#flake-references).
+   A [flake registry] allows using symbolic identifiers for further brevity.
+   References can be automatically locked to their current specific version and later updated programmatically.
+
+   [flake registry]: https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-registry.html
+
+1. A [new command line interface], implemented as a separate experimental feature, leverages flakes by accepting flake references in order to build, run, or deploy software defined as a flake.
+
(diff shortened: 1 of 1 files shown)


zmitchell (changes requested):
This is an explanation and thus no longer a recipe. This and #546 should probably live in a single file under the "concepts" section.

fricklerhandwerk:
I don't fully agree but also don't disagree that it's not recipe material. The answer doesn't really explain anything, it just states the difference to classic Nix. I suggest parking it here regardless under FAQ, since it is supposed to live next to the other flakes answers.

We could still later split up the recipe answers and leave a general FAQ in a different section, which could be filled with other actual questions people have, such as how Nix relates to Docker.

asymmetric:
Not familiar with the proposed structure nix.dev will have, but I'm in favor of putting this under "concepts" or something else. Agree it doesn't make sense under "recipes".

source/recipes/faq.md
```diff
@@ -2,6 +2,44 @@
...
 
+### What are flakes?
+
+What is usually referred to as "flakes" is:
+- A policy for managing dependencies between {term}`Nix expression`s.
+- An [experimental feature] in Nix, implementing that policy and supporting functionality.
+
+[experimental feature]: https://nixos.org/manual/nix/stable/contributing/experimental-features.html
+
+Technically, a [flake](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#description) is a file system tree that contains a {term}`Nix file` named `flake.nix` in its root directory.
```
infinisil, line 13: I hate to have to point this out, but it's technically not a Nix file, because not the entire Nix language is supported, see https://github.com/NixOS/nix/issues/4945

fricklerhandwerk: Thank you. I love how you manage to break things.

Suggestions on how to stick to the truth without neither misleading or confusing people, nor going over board with details?

"a file named `flake.nix`" does not point out the distinction because it leaves open which language it's in (and one may assume the technically wrong thing)

Maybe one could add that bit to where we talk about the schema?

infinisil: Yeah that sounds good. Alternatively how about adding a footnote at "Nix file" and then link to that issue in it.

fricklerhandwerk: No, I think we should not pollute documentation of stable interfaces with that stuff. Once stabilising flakes can be deemed on the finish line, we can still think about how and where to talk about these things.

zmitchell:
Answers to questions like "what even _is_ XYZ?" strike me as conceptual, understanding oriented materials, so it makes sense to me to put that in Concepts. I don't want this in the recipes section because we both agree it's not recipe material, meaning that at some point down the line we'll end up moving it. So, let's not make work for our future selves and put it somewhere else.

fricklerhandwerk:
Discussed in Nix docs team meeting:

- @fricklerhandwerk: What if we rename "FAQ" to "How-to guides" instead and create a concept section, put flakes in there?
- @infinisil: Make "What is Flakes" link to the section in the concepts
- @fricklerhandwerk: What about e.g. "Why are flakes controversial"? It's an FAQ
- @zmitchell: Also move that to concepts
- @roberth: Would be good to have a compiled glossary, because people don't know where to look and some concepts cross project boundaries
  - @infinisil: Needs tooling
    - @fricklerhandwerk: Doubt it's worthwhile right now
- @fricklerhandwerk: Will create a concepts header with just the flakes section in it for now
  - @roberth: Should the concepts be migrated to glossary entries?
  - @zmitchell: Glossary entries should be short, concepts longer
  - @roberth: Seems more like a task for the individual projects
- @fricklerhandwerk: Need to teach people how to understand the ecosystem, how parts interact with each other. Should probably have a diagram or something in the concepts header/glossary
- decision: make a concepts section and add flakes in there as a first entry

fricklerhandwerk:
Merging since the general direction was reviewed and agreed upon by the docs team. Details can be fixed forward.

PR #560, opened 2023-05-16
First version of the document survey table template

henrik-ch:
Please let me know what you think. 
An especially open point is the types of tags we would use, which this template suggestion doesn't define yet. 

I appreciate any feedback on the other details and columns as well of course - thanks a lot.
By the way, a tip for beginners - this edited md page doesn't generate onto the website itself, so the easiest way to view and review this change is to navigate to the "Files changed" tab and  "Display the rich diff" - button that looks like a blank page.

### DIFF
--- a/maintainers/working_groups/learning_journey/documentation-survey.md
+++ b/maintainers/working_groups/learning_journey/documentation-survey.md
@@ -1,15 +1,20 @@
 # Documentation survey
 
-## Table template
-
-The first row defines possible values and the second row gives an example for the first entry.
 
 
-|                            URL                            |        Content Type        |                                    Summary                                    | Video length/read estimate | Year Updated | Nix-Commands                  | Author    | Tags                    |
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
What I'd care about in terms of metadata is
- summary (what do I get from dealing with it?)
- video length or estimated reading time (how long does it take?)
- author (who is the source of that information?)
- year of production - maybe month if it's younger than one year, but not that important (how up to date is it)

The additional information, such as which commands are used or which concepts are explained may be valuable, but before we have a quite sophisticated overview it may be enough to encode it softly in the summaries.

We can still add links to and from these aspects once we have them sorted out on our side. Let's not overthink it for now.

henrik-ch:
I think that the new version is more close to what you suggested @fricklerhandwerk.

I suggest limiting the summary to 80 words or less. The table might still get unwieldy, but hopefully still okay enough to work with.

henrik-ch:
Updated as we talked about in the meeting today. 
If we are happy with this version, then I can start updating the information for each URL accordingly.

PR #561, opened 2023-05-16
update documentation project outline, add details

fricklerhandwerk:
as discussed with @zmitchell and @henrik-ch, in order to brief project participants @JacklineYim and @roberth.


fricklerhandwerk:
Ah thanks. That's what happens if you do too many things at once quickly.

PR #575, opened 2023-05-29
Change project's self-description

### LINKED ISSUE #536: Update nix.dev site tagline
## Description
Today the tagline on nix.dev is "An opinionated guide for developers getting things done using the Nix ecosystem."

## Problem
Now that nix.dev is an official project, the tagline should reflect the position of the site within the project.

## Approach
Change the site description in the `source/index.md` front matter to reflect the status of `nix.dev` as an official project and the hub for official Nix documentation.

asymmetric:
Feel free to propose better wording.

Closes #536

### DIFF
--- a/source/_templates/about.html
+++ b/source/_templates/about.html
@@ -7,4 +7,4 @@
 
 <iframe src="https://ghbtns.com/github-btn.html?user=nix-dot-dev&repo=nix.dev&type=star&count=true&size=large" frameborder="0" scrolling="0" width="170" height="30" title="GitHub"></iframe>
 
-<p>The official resource for Nix documentation.</p>
+<p>Official documentation for the Nix platform.</p>
--- a/source/_templates/about.html
+++ b/source/_templates/about.html
@@ -7,4 +7,4 @@
 
 <iframe src="https://ghbtns.com/github-btn.html?user=nix-dot-dev&repo=nix.dev&type=star&count=true&size=large" frameborder="0" scrolling="0" width="170" height="30" title="GitHub"></iframe>
 
-<p>Official documentation for the Nix platform.</p>
+<p>Official documentation for getting things done with Nix</p>
--- a/source/_templates/about.html
+++ b/source/_templates/about.html
@@ -7,4 +7,4 @@
 
 <iframe src="https://ghbtns.com/github-btn.html?user=nix-dot-dev&repo=nix.dev&type=star&count=true&size=large" frameborder="0" scrolling="0" width="170" height="30" title="GitHub"></iframe>
 
-<p>Official documentation for getting things done with Nix</p>
+<p>Official documentation for getting things done with Nix.</p>
--- a/source/_templates/about.html
+++ b/source/_templates/about.html
@@ -7,4 +7,4 @@
 
 <iframe src="https://ghbtns.com/github-btn.html?user=nix-dot-dev&repo=nix.dev&type=star&count=true&size=large" frameborder="0" scrolling="0" width="170" height="30" title="GitHub"></iframe>
 
-<p>An opinionated guide for developers getting things done using <a href="https://nixos.org">the Nix ecosystem</a>.</p>
+<p>The official resource for Nix documentation.</p>
--- a/source/reference/glossary.md
+++ b/source/reference/glossary.md
@@ -45,10 +45,6 @@ NixOS
     - [NixOS source code](https://github.com/NixOS/nixpkgs/tree/master/nixos)
     :::
 
-Nix platform
-    The set of components that combine to create the official Nix ecosystem:
-    Nix, Nixpkgs, NixOS.
-
 reproducible
     Reproducibility would guarantee exactly the same results no matter
     **when** or **on what machine** you run the command.
--- a/source/reference/glossary.md
+++ b/source/reference/glossary.md
@@ -45,6 +45,10 @@ NixOS
     - [NixOS source code](https://github.com/NixOS/nixpkgs/tree/master/nixos)
     :::
 
+Nix platform
+    The set of components that combine to create the official Nix ecosystem:
+    Nix, Nixpkgs, NixOS.
+
 reproducible
     Reproducibility would guarantee exactly the same results no matter
     **when** or **on what machine** you run the command.
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 # [nix.dev](https://nix.dev)
 
-The official resource for Nix documentation.
+Official documentation for the Nix platform.
 
 ## Contributing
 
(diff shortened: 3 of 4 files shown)


fricklerhandwerk:
No objections from my side. Is this blocked by anything we should take care of to avoid confusion or wrong impressions?

fricklerhandwerk (approved):

asymmetric:
I think it's mostly about: is nix.dev only for Nix documentation? Or does/will it cover nixpkgs/nixos as well?

And is documentation broad enough, to include tutorials, howtos, etc, or should we mention those explicitly?

fricklerhandwerk:
> is nix.dev only for Nix documentation?

We can resolve the ambiguity by referring to the "Nix ecosystem" encompassing all official projects (or a new term that emerged in the UX workshop last week, that was supposed to distinguish from all related but not official projects: Nix platform - @roberth @infinisil what do you think about that by the way?)

For example: "Official documentation for the Nix ecosystem"

asymmetric:
I like Nix platform, it's going in the right direction IMO by pointing to the fact that it's an integrated set of components.

Ecosystem is broader and could include unofficial things like HM.

fricklerhandwerk:
If we decide to use "Nix platform", we should update the [glossary](https://github.com/NixOS/nix.dev/blob/dec25a814e7f9e9aee13ebb58cddc91461c2693d/source/reference/glossary.md) right within this PR. Since it's completely new, people will otherwise (rightfully) get confused.

asymmetric:
@fricklerhandwerk what do you think of the definition?

source/reference/glossary.md
```diff
@@ -25,6 +25,10 @@ NixOS
 
     Read /nɪks oʊ ɛs/ ("Niks Oh Es").
 
+Nix platform
+    The set of components that combine to create the official Nix ecosystem:
+    Nix, Nixpkgs, NixOS.
+
```
fricklerhandwerk, line 31:
```suggestion
Nix platform
    The officially supported software components that form the basis of the {term}`Nix ecosystem`:
    Nix, Nixpkgs, NixOS.

Nix ecosystem
    Products and services based on the {term}`Nix platform`.

```

Should also define Nix ecosystem there while we're at it.

fricklerhandwerk (approved):
I like it, but it's still very fresh. It's great we're finally pushing this issue (thanks a lot for taking care!), but before introducing a new term  let's wait for what others think, especially @NixOS/documentation-team, @NixOS/marketing-team, and possibly @NixOS/nix-team. I don't want us to make such a far-reaching move if it doesn't have broad support.

06kellyjac:
I find "nix platform" a bit confusing to be honest.

Nix ecosystem is something I'm used to, and makes sense as a collection of tightly related projects, but with the proposed descriptions I'm not sure what the "platform" is and how it differs from the "ecosystem"

If I had to call anything the "nix platform" I instantly think of the nix CLI/tooling but that is quite different to what we're going for here.

Also think along the lines of Platform as you'd find in PaaS (Platform as a Service) it tends to be more of a product offering like Elastic Beanstalk or Heroku

thufschmitt:
I'm personally fine with the term. I think it's useful (it is indeed different from “Nix ecosystem”), and “platform” is not a bad term to describe it. Now this is largely out of scope for the Nix team, so I'll let more relevant people voice their opinion :wink:

06kellyjac:
I don't want to say it's a bad term, it's just based on what's written here I personally don't really understand it, there might be a bit of extra detail that can be added to the description to make the distinction clear.

Maybe I need to just let the term stew in my head for a while kind of like when there's a UI change you start off disliking but end up liking? :shrug:

zmitchell:
My preference is to keep "Nix ecosystem". The term "Nix platform" isn't widely used, and therefore people in the _ecosystem_ won't have a shared understanding of what it is.

infinisil (approved):
I think it's accurate to use "Nix Platform" to describe the officially supported Nix, NixOS and Nixpkgs projects. It's almost the literal platform the ecosystem is built upon.

We might want to take a look at https://github.com/orgs/NixOS/repositories at some point to figure out if it's more than just these three projects, in particular these stand out:
- https://github.com/NixOS/nixos-hardware (arguably part of NixOS though)
- https://github.com/NixOS/nixops (not really maintained anymore though)
- https://github.com/NixOS/hydra

06kellyjac:
Ok. The "nix platform" being official projects but everything else sitting as "nix ecosystem" makes sense. If that nuance could be added to the description/definition it would help if I was reading this

06kellyjac:
> it is explained, in the glossary, perhaps you've missed that?

Yeah I read the original glossary. The suggested change is pretty close to what I'm looking for actually yes. I'll add some comments there

zmitchell:
I don't have a strong opinion on this, if other people think "Nix platform" makes sense, please go ahead

source/reference/glossary.md
fricklerhandwerk:
```suggestion
Nix platform
    The officially supported software components that form the basis of the {term}`Nix ecosystem`:
    Nix, Nixpkgs, NixOS.

Nix ecosystem
    All products and services based on the {term}`Nix platform`.

```

I think we should keep "the" as they are definite. How about "all" for the ecosystem?

asymmetric:
@zmitchell as discussed during the meeting (see notes), @fricklerhandwerk proposed to wait until the foundation has capacity to weigh in on this.

vcunat:
In my mind there's a slight clash with Nix being "cross-platform".  (But certainly not a big deal to me.  It's not like I have a term that I consider significantly better.)

source/_templates/about.html
```diff
@@ -2,7 +2,7 @@ <h1><a href="https://nix.dev">nix.dev</a></h1>
 
 <iframe src="https://ghbtns.com/github-btn.html?user=nix-dot-dev&repo=nix.dev&type=star&count=true&size=large" frameborder="0" scrolling="0" width="170" height="30" title="GitHub"></iframe>
 
-<p>An opinionated guide for developers getting things done using <a href="https://nixos.org">the Nix ecosystem</a>.</p>
+<p>Official documentation for the Nix platform.</p>
```
fricklerhandwerk, line 5: Here's an alternative that wouldn't deviate from the original gist, and also avoids going down the naming rabbit hole.

```suggestion
<p>Official documentation for getting things done with Nix</p>
```

roberth: The Nix platform is cool and all, and maybe the documentation is a largely _about_ the Nix platform, but this documentation is not _for_ the Nix platform, but rather _for_ getting your own things done (where learning is also one of the things I suppose; not just guides but also tutorials). So I like this suggestion better.

asymmetric:
Rebased to fix conflicts.

fricklerhandwerk (approved):
Thanks @asymmetric for keeping at it!

PR #585, opened 2023-06-06
Add Nix snowflake logo

yukiisbored:
Screenshot:
image


source/_templates/about.html
```diff
@@ -1,11 +1,10 @@
-<h1><a href="https://nix.dev">nix.dev</a></h1>
+<div class="logo">
+  <a href="/">
+    <img src="/_static/img/nix.svg" alt="Nix snowflake logo">
```
fricklerhandwerk, line 3:
```suggestion
    <img src="/_static/img/nix.svg" alt="">
```

One can consider this a decorative element, which then should have an empty `alt` tag. Otherwise screen readers will say "Nix snowflake logo nix dot dev", which is not helpful at all.

fricklerhandwerk (approved):
Great, thanks a lot. I renamed the logo file name to be just `nix.svg` because IMHO there is not much more to it.

PR #83, opened 2020-07-20
Add nix-community direnv section

fzakaria:
Add section for direnv that links readers to the nix-community
implementation as it's:
1 - faster due to caching of the environment
2 - creates a gcroots

Open question:
 Why isn't this implementation upstreamed to the default direnv
 implementation ?

PR #597, opened 2023-06-18
rebase of fzakaria pull request 83

henrik-ch:
This code is a slight format adaption and rebase from the PR #83  from @fzakaria .

### DIFF
--- a/source/tutorials/first-steps/declarative-and-reproducible-developer-environments.md
+++ b/source/tutorials/first-steps/declarative-and-reproducible-developer-environments.md
@@ -136,6 +136,16 @@ direnv: using nix
 hello
 ```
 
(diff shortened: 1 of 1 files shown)


fzakaria:
Ty !

fricklerhandwerk:
As mentioned in https://github.com/NixOS/nix.dev/pull/83#issuecomment-1594741412 I think this should just be the default instead of adding more options. @henrik-ch could you fix it up accordingly?

henrik-ch:
I made a new commit - it reduced the scope of the change a fair bit. 
I hope that this is what you were looking for @fricklerhandwerk .

fricklerhandwerk (approved):

PR #608, opened 2023-06-21
add redirects for changed directory structure

fricklerhandwerk:
remove unnecessary nesting

@nixos/documentation-team


roberth:
I don't see how `first-steps` is "incompatible" with `nixos`. This seems to be another bogus categorization.
Order is good, but this is not order.
Let's pick a single axis of categorization or stay away from ontology entirely.
The only reasonable axis that I think can be assessed objectively is which project is the main subject of discussion, but that assumes there's a main subject, ie it excludes "Write a service for both NixOS and nix-darwin with minimal duplication".
So either we pick a mangeable set like {Nix, Nixpkgs, NixOS, nix-darwin, home-manager, ecosystem} where "ecosystem" is a catch-all, or we don't pick anything and keep it simple.
The GitHub action one isn't even about NixOS, and I don't want to assess the other labels. Because that's what they are, labels; not a hierarchy.
Let's avoid having to do an RFC 140² Simple Tutorial Paths.

roberth:
> remove unnecessary nesting

This adds nesting? I'm confused.

fricklerhandwerk:
@Roberth I agree categorisation is a problem, and hierarchical structure usually fails to convey the right meaning. This PR is only to restore broken URLs though and I'll just merge it for that purpose.

Refining the navigation structure and thinking about labeling should continue regardless. While this will only increase the number of redirects over time, this is what it takes. Wikis have the advantage of providing that by default, but also other disadvantages such lack of branch&merge version control. Setting up an immutable and convenient web document infrastructure is not our rabbit hole, although having it filled would sure help keeping sanity.

PR #609, opened 2023-06-21
Nix language tutorial: restructure introduction

fricklerhandwerk:
this collapses the "don't panic" section into a motivating introduction and gives
the document a more distinctive name.

this also addresses feedback stating that having "basics" in the title
may be misleading about the effort required to work through the
tutorial.

this also fixes a long-standing gap of not mentioning function arguments as another form of name assignment.

@NixOS/documentation-team

### DIFF
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -5,7 +5,9 @@
 The Nix language is designed for conveniently creating and composing *derivations* – precise descriptions of how contents of existing files are used to derive new files.
 It is a domain-specific, purely functional, lazily evaluated, dynamically typed programming language.
 
-:::{Notable uses of the Nix language}
+:::{admonition} Notable uses of the Nix language
+:class: note
+
 
 - {term}`Nixpkgs`
 
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -240,15 +240,14 @@ let x=1;y=2;in x+y
 
 Values in the Nix language can be primitive data types, lists, attribute sets, and functions.
 
-We show primitive data types and lists as examples in the context of [attribute sets](attrset).
+We show examples of primitive data types and lists in the context of [attribute sets](attrset).
 Later in this section we cover special features of character strings: [string interpolation](string-interpolation), [file system paths](file-system-paths), and [indented strings](indented-strings).
 We deal with [functions](functions) separately.
 
-There are two ways to assign names to values in Nix: [attribute sets](attrset) and [`let` expressions](let).
-
+[Attribute sets](attrset) and [`let` expressions](let) are used to assign names to values.
 Assignments are denoted by a single equal sign (`=`).
 
-Whenever you see an equal sign (`=`) in Nix language code:
+Whenever you encounter an equal sign (`=`) in Nix language code:
 - On its left is the assigned name.
 - On its right is the value, delimited by a semicolon (`;`).
 
@@ -1031,6 +1030,9 @@ Wherever you see a colon (`:`) in Nix language code:
 - On its left is the function argument
 - On its right is the function body.
 
+Function arguments are the third way, apart from [attribute sets](attrset) and [`let` expressions](let), to assign names to values.
+Notably, values are not known in advance: the names are used as placeholders that are filled when [calling a function](calling-functions).
+
 Function declarations in the Nix language can appear in different forms.
 Each of them is explained in the following, and here is an overview:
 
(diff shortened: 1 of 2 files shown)


source/tutorials/first-steps/nix-language.md
```diff
@@ -1,28 +1,53 @@
...
+
+  that is, lists and attribute sets
+
+- Functions and operators
+
+  to produce and transform data
+
+- Name assignment
+
+  to manipulate data as units
```
roberth, line 43: I suppose this refers to scope and perhaps `let` bindings specifically?

```suggestion
- Name assignment

  to reuse data, functions and function arguments
```

I don't quite understand the intent behind these "units", as I can't imagine how naming something would be possible without it being unitary in some given context.

Maybe I'm [Mann-Gelling](https://freethoughtblogs.com/pharyngula/2023/06/19/today-i-learned-about-gell-mann-amnesia/)?

roberth:
What is the process here? Are you extending the introduction because it's too short? Is the desired length of an introduction documented somewhere?

fricklerhandwerk:
No, the intent is to focus the introduction on the motivation for the particular approach. I was thinking about cutting down on the [elements of programming](https://mitp-content-server.mit.edu/books/content/sectbyfn/books_pres_0/6515/sicp.zip/full-text/book/book-Z-H-10.html#%_sec_1.1), but didn't find a good way so there they still are. How would you illustrate to someone not necessarily familiar with functional programming that a language is small and has few fundamental moving parts?

yukiisbored (approved):

roberth:
> How would you illustrate to someone not necessarily familiar with functional programming that a language is small and has few fundamental moving parts?

I'd sneak in the three elements of lambda calculus (variable reference, lambda abstraction, application):

- functions: you can describe complex behavior by introducing a name for something you don't have
- function application: you can use such behaviors by filling in the thing that was missing, as many times as you need
- names: you can use a name to refer to something you've introduced with a function, or what's made available to you by the programming environment

So this de-emphasizes the role of "primitive values" compared to The Elements Of Programming you referenced.
Not coincidentally that's the purpose of lambda calculus: a simplified model of computation, and indeed all data can be expressed using only a function primitive. (e.g. consider `true = a: b: a; false = a: b: b;` or the more general encoding schemes that exist for "usable" data types)

Nix is a functional language, so maybe this is a useful model?
When you have the lambda calculus part down, and the sugar like `let`, `inherit`, and the unrelated monster `with`, you're going to be thinking about exactly all the other things that make up your problem domain, which is of course the bread and butter of what we're documenting. So that seems like a sensible boundary for dividing up the material "reference style", but for a learning experience, I guess we'll want to return to this subject to "exercise" the lambda calculus thinking. It's not just a thing you use but a cognitive skill.

source/tutorials/first-steps/nix-language.md
fricklerhandwerk: I think the original intent behind the "Elements" is rather syntactic, and names are syntactic units, allowing you to refer concisely to semantic units (the things we assign names to).

source/tutorials/first-steps/nix-language.md
```diff
@@ -1112,7 +1079,9 @@ Each of them is explained in the following, and here is an overview:
   ```
 
 Functions have no names.
-We say they are anonymous, and call such a function a *lambda*.
+We say they are anonymous, and call such a function a *lambda*.[^lambda]
+
+[^lambda]: The term *lambda* is a shorthand for [lambda abstraction](https://en.m.wikipedia.org/wiki/Lambda_calculus#lambdaAbstr) in the [lambda calculus](https://en.m.wikipedia.org/wiki/Lambda_calculus).
```
yukiisbored, line 1084: Don't use the mobile links since desktop users will be greeted with a mobile version of Wikipedia.

fricklerhandwerk: ...which is better in so many ways, but indeed, I did that inadvertently because my browser redirects me to it automatically. Will fix.

fricklerhandwerk:
```suggestion
[^lambda]: The term *lambda* is a shorthand for [lambda abstraction](https://en.wikipedia.org/wiki/Lambda_calculus#lambdaAbstr) in the [lambda calculus](https://en.wikipedia.org/wiki/Lambda_calculus).
```

source/tutorials/first-steps/nix-language.md
```diff
@@ -1,28 +1,31 @@
...
-# Nix language basics
-
-The Nix language is used to declare packages and configurations to be built by [Nix][nix-manual].
+# Reading the Nix language without fear
 
+The Nix language is designed for conveniently creating and composing *derivations* – precise descriptions of how contents of existing files are used to derive new files.
 It is a domain-specific, purely functional, lazily evaluated, dynamically typed programming language.
 
-Notable uses of the Nix language are:
+:::{Notable uses of the Nix language}
```
fricklerhandwerk, line 8:
```suggestion
:::{admonition} Notable uses of the Nix language
```

PR #621, opened 2023-06-27
tutorials/first-steps/nix-language: Fix the nixpkgs example

CyberShadow:
The example didn't build for me, and on the Matrix room, it was pointed out that `fetchTarball` doesn't work with `mirror://`.

I fixed the code to build, and tried to update the explanation to match.

I also added a command which builds the package.

### DIFF
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -2052,8 +2052,6 @@ Explanation:
   `fetchurl` itself comes from the outer function's arguments.
 - The `meta` attribute is itself an attribute set, where the `license` attribute has the value that was assigned to the nested attribute `lib.licenses.gpl3Plus`.
 
-The package can be built by running `nix-build -E 'with import <nixpkgs> {}; callPackage ./hello.nix {}'`.
-
 ## References
 
 - [Nix manual: Nix language][manual-language]
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -2048,7 +2048,8 @@ Explanation:
 
 - This expression is a function that takes an attribute set which must have exactly the attributes `lib`, `stdenv`, and `fetchurl`.
 - It returns the result of evaluating the function `mkDerivation`, which is an attribute of `stdenv`, applied to a recursive set.
-- The recursive set passed to `mkDerivation` uses its own `pname` and `version` attributes in the argument to the function `fetchurl`, which is an attribute of the argument attribute set.
+- The recursive set passed to `mkDerivation` uses its own `pname` and `version` attributes in the argument to the function `fetchurl`.
+  `fetchurl` itself comes from the outer function's arguments.
 - The `meta` attribute is itself an attribute set, where the `license` attribute has the value that was assigned to the nested attribute `lib.licenses.gpl3Plus`.
 
 The package can be built by running `nix-build -E 'with import <nixpkgs> {}; callPackage ./hello.nix {}'`.
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -2022,7 +2022,7 @@ Explanation:
 ### Package
 
 ```{code-block} nix
-{ lib, stdenv }:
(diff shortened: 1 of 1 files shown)


source/tutorials/first-steps/nix-language.md
```diff
@@ -2046,11 +2046,13 @@ This example is a (simplified) package declaration from Nixpkgs.
 
 Explanation:
 
-- This expression is a function that takes an attribute set which must have exactly the attributes `lib` and `stdenv`.
+- This expression is a function that takes an attribute set which must have exactly the attributes `lib`, `stdenv`, and `fetchurl`.
```
CyberShadow, line 2049: Perhaps it way be worthwhile to mention here why `nixpkgs` expressions don't need to use `...`. (The answer is that `callPackage` uses `builtins.functionArgs` to introspect the function, so it knows what arguments it expects.)

fricklerhandwerk: While it's true, I don't think we should pollute overwhelmed beginners' minds with additional information. `callPackage` is entirely out of scope here. That bit should be documented in Nixpkgs contributing guides or the part on `callPackage`.

source/tutorials/first-steps/nix-language.md
```diff
@@ -2046,11 +2046,13 @@ This example is a (simplified) package declaration from Nixpkgs.
...
 - It returns the result of evaluating the function `mkDerivation`, which is an attribute of `stdenv`, applied to a recursive set.
-- The recursive set passed to `mkDerivation` uses its own `pname` and `version` attributes in the argument to the built-in function `fetchTarball`.
+- The recursive set passed to `mkDerivation` uses its own `pname` and `version` attributes in the argument to the function `fetchurl`, which is an attribute of the argument attribute set.
```
fricklerhandwerk, line 2051:
```suggestion
- The recursive set passed to `mkDerivation` uses its own `pname` and `version` attributes in the argument to the function `fetchurl`.
  `fetchurl` itself comes from the outer function's arguments.
```

Not sure about the phrasing. It's a bit clumsy, but I agree it should be made explicit.

source/tutorials/first-steps/nix-language.md
```diff
@@ -2046,11 +2046,13 @@ This example is a (simplified) package declaration from Nixpkgs.
...
 - The `meta` attribute is itself an attribute set, where the `license` attribute has the value that was assigned to the nested attribute `lib.licenses.gpl3Plus`.
 
+The package can be built by running `nix-build -E 'with import <nixpkgs> {}; callPackage ./hello.nix {}'`.
```
fricklerhandwerk, line 2054: While it's good to make sure the example is correct, I'm not sure we should have it here. It introduces an entirely new concept with `callPackage`, which is quite magical.

For a moment I contemplated if we could make a "naive" example with the tools provided here, but it would be misleading because no one does it that way. Also I wouldn't want to go deeply into assessing the various ways to access packages here. It's a can of worms on its own. This tutorial is only about the language.

I think it's better to leave it out. @NixOS/documentation-team what do you think?

fricklerhandwerk (commented):
Thanks for testing and fixing. We lose a bit of exercise fidelity with dropping use of a built-in. I don't think it's a big problem, and having a more authentic example is probably worth it.

Could need some polishing, otherwise looks good.

source/tutorials/first-steps/nix-language.md
CyberShadow: For what it's worth, here is my experience as a beginner:

1. Read https://nixos.org/manual/nix/stable/language/index.html
2. Read this tutorial
3. Be confused why `...` is not required here. Was the language spec out of date? I had to ask on the Matrix channel to learn that it works because the caller does introspection.

fricklerhandwerk: I understand. Still I'd like not to expand the scope here. The `callPackage` pattern is surely worth an introduction of it's own, and we could import and slightly update this article: https://summer.nixos.org/blog/callpackage-a-tool-for-the-lazy/ (it needs a new home anyway) and then link to it the end of this tutorial.

fricklerhandwerk (approved):

PR #645, opened 2023-07-13
Module system deep dive tutorial

infinisil:
Effectively a raw dump of https://github.com/tweag/summer-of-nix-modules

Overall structure seems usable, but needs some work


source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+```
+
+## Declaring generate.script option
+
+In order for modules to be useful, we need to have options, so let's
+declare one. Options are declared by defining an attribute with
+`lib.mkOption` under the `options` attribute. Here we're defining the
+option `generate.script`.
+
+While there are many attributes to customize options, the most
```
hugosenari, line 51: Where I can find those other attributes?

infinisil: Afaik the source has the only complete listing: https://github.com/NixOS/nixpkgs/blob/fa1d4611cc2ce41d7b93c4a8e8db0efaaf286475/lib/options.nix#L66-L88

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+}
+```
+
+Then we run
+```bash
+nix-instantiate --eval eval.nix -A config.generate.script
+```
+
+Trying to evaluate the `generate.script` option however, we get an error
+that the option is used but not defined, indicating that we need to
```
hugosenari, line 94: This always confounded me:
- `= config.` _used_: basic¹ user may never see this
- `config. =` _defined_: to his POV this is where he _used_ or _defined_ it
- `options. =` _defined?_: to me, this is where I defined it (I'm not basic user)


_¹basic_:
lower basic nix: never saw a nix file before :baby: 
basic nix modules use: `config. = ""`, use existing options with static values
mid nix modules use: `config. = config.`, use existing options reusing existing values
upper mid nix modules use: `config. = lib.mkIf config.`, use existing options using some kind of logic
higher nix modules use: `options. =`, define his own options

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+element has to match the given nested type. In our case we want `str` to
+be the nested type, which is a generic string type.
+
+Note that the difference between `str` and `lines` is in their merging
+behavior:
+- For `lines`, multiple definitions get merged by concatenation with
+  newlines
+- For `str`, multiple definitions are not allowed. Which in this case is
+  mostly irrelevant however, since we can't really define a list element
+  multiple times.
```
hugosenari, line 164: Maybe we should be using _str_ instead of _lines_ at line 70, usually, scripts aren't concatenated anyway.
Then we could move this to line 56
It would reduce complexity (number of concepts) in section of lists.

infinisil: I intentionally tried to introduce as many different types as possible to show what's there. I think it's somewhat common for scripts to be concatenated though, e.g. `systemd.services.<name>.{script,preStart}` is like that

source/tutorials/module-system/module-system.md
infinisil: In the module system, `config = ` are usually called definitions, while `options = ` are called declarations

source/tutorials/module-system/module-system.md
hugosenari: > I intentionally tried to introduce as many different types as possible to show what's there.

Agree with this, but we are in a list section, reader have to bash his head against lists, e we're talking about strings,
Ok that computationally speaking a string is a list of chars, and lines is internally represented as a list of strings. But no one needs to know it until the next level. ;-)

Other problem of this, I usually read documentations again trying to find concepts that I know they exist in that document like a reference/dictionary. Include string in list, would be like including the definition of some S word in L page.

That is why suggested move this to line 56 (where we introduce lines)
Also, I was thinking in add other string types strMatching and enum (ok is not a string but mostly used with string)

This suggestion also reduce our str definition that has this _Which in this case is mostly irrelevant however, since we can't really define a list element multiple times._ that may cause confusion, since we are telling something important and saying that it doesn't matter.

hugosenari:
Your PR make content about modules in this site infinitely more comprehensive. :heart: 
Thanks

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
++            map attrForLocation path.locations;
++        in "path=${
++            lib.concatStringsSep "\\|" attributes
++          }";
++      in map paramForPath config.map.paths;
++  };
++}
+```
+
+## Arrival marker
```
proofconstruction, line 779:
```suggestion
## The Arrival Marker
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+       (marker: marker.location != null)
+       (lib.concatMap (user: [
+-        user.departure
++        user.departure user.arrival
+       ]) (lib.attrValues config.users));
+ 
+     map.center = lib.mkIf
+```
+
+## Connecting user paths
```
proofconstruction, line 820:
```suggestion
## Connecting Markers by Paths
```

proofconstruction (dismissed):
Updated section titles

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
+# Module system introduction
+
```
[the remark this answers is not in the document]
fricklerhandwerk: Actually this is really good, just reorder and reword slightly.

```suggestion
Much of the power in NixOS comes from the module system.
It provides mechanisms for conveniently declaring and automatically merging interdependent attribute sets, making it easy to compose configurations in a type-safe way.

In this tutorial you'll learn what a module is and how to define one, what options are and how to declare them, how to express dependencies between modules, and a practical way to use Nix to wrap external APIs.

You'll write your first modules to interact with the Google Maps API, declaring options which represent map geometry, location pins, and more.

Be prepared to see some Nix errors: during the tutorial, you will first write some *incorrect* configurations, creating opportunities to discuss the resulting error messages and how to resolve them, particularly when discussing type checking.
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+diff --git a/default.nix b/default.nix
+index 1797133..f7569bd 100644
+--- a/default.nix
++++ b/default.nix
+@@ -1,3 +1,3 @@
+-{
++{ lib, ... }: {
+   
+ }
+```
```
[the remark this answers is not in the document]
fricklerhandwerk: Not sure if we should say "matching", as the more functionally minded people may read this as pattern matching. This is not even remotely the case.

In general, I was just recently thinking about it: While it's true that there are multiple "syntaxes" for the module system, we probably should teach their existence as an advanced step. How about we stick to `{ ... }: { }` as the main form? I don't think it's worthwhile showing an empty attrset just [for the sake of simplicity](https://m.xkcd.com/2587/).

infinisil:
@proofconstruction Please feel free to push to the branch here directly or open a separate PR!

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+
+## Successful evaluation: Assigning a string to generate.script
+
+We can make type checking pass by assigning a string to the option,
+giving us our first successful evaluation of the `generate.script`
+option.
+
+In this case, we assign a script which calls the Google Maps Static API
+to generate a world map, then displaying the result using icat
```
zmitchell, line 128:
```suggestion
to generate a world map, then displaying the result using `icat`
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+   };
+   
+ }
+```
+
+TODO: Create derivations to get these commands
+
+## A new list option: Declaring generate.requestParams
+
+Let's introduce another option, generate.requestParams. For its type,
```
zmitchell, line 153:
```suggestion
Let's introduce another option, `generate.requestParams`. For its type,
```

source/tutorials/module-system/module-system.md
zmitchell: I think how it's written makes sense. Don't block on this. If anything I would just remove the entire "Which in this case ..." sentence since by it's own admission it's not terribly relevant.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
++      "scale=2"
++    ];
+   };
+   
+ }
+```
+
+## Dependencies between options: Using generate.requestParams
+
+A collection of modules generally only has a single option that is meant
```
zmitchell, line 197: This could be clearer. What I think you trying to say is that a user may only specify a single option, which causes several other options to then be evaluated (or something like that). When I read "a collection of modules generally only has a single option that is meant to be evaluated" I wonder how that can be the case. How can several modules all work together to provide a _single_ option? What's the point of the modules that don't contain that option? These are all questions a new user will have after reading this.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+to be evaluated. That's the option that generates the final result we're
+interested in, which in our case is `generate.script`.
+
+In order to build up abstractions on that, we have the ability for
+options to depend on other options. In this case we want the
+`generate.script` option to use the values of `generate.requestParams`.
+We can access the values of options by adding the `config` argument to
+the argument list at the top and using e.g.
+`config.generate.requestParams` to access that options value.
```
zmitchell, line 206: You need to explain why this works the way that it does. This is the thought process of a new user in this scenario:
- I've defined `generate.script` and `generate.requestParams` next to each other in the `config` attribute.
- If they're next to each other why can't I refer to them? Can't I use `rec`?
- The correct way is to add a function argument named `config`, but I'm currently defining `config`. How on earth does that make sense?

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+
+## Conditional definitions: Introducing map.zoom option
+
+We now introduce the new option `map.zoom` in order to control the zoom
+level of the map. We'll use a new type for it, `nullOr <type>`, which
+accepts the value `null`, but also values of its argument type. We're
+using `null` here to mean an inferred zoom level.
+
+For this option, we'll set a default value which should be used if it's
+not defined otherwise, which we can do using `mkOption`'s `default`
```
zmitchell, line 243: This isn't specific to this tutorial, but we need some kind of terminology for the attributes that make up an attribute set passed as an argument. For example, it's incorrect to say that `buildInputs` is an argument to `mkDerivation` since `mkDerivation` is a function that takes a single argument and that argument is an attrset. Maybe something like "input attribute"?

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+--- a/default.nix
++++ b/default.nix
+@@ -8,6 +8,13 @@
+     generate.requestParams = lib.mkOption {
+       type = lib.types.listOf lib.types.str;
+     };
++
++    map = {
++      zoom = lib.mkOption {
++        type = lib.types.nullOr lib.types.int;
```
zmitchell, line 264: How can this ever be null if there's a default value?

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
++      (lib.mkIf (config.map.zoom != null)
++        "zoom=${toString config.map.zoom}")
+     ];
+   };
+   
+```
+
+## Declaring map.center option
+
+Similarly, let's declare a map.center option, declaring where the map
```
zmitchell, line 284:
```suggestion
Similarly, let's declare a `map.center` option, declaring where the map
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+should be centered.
+
+We'll be using a small utility for geocoding location names, aka turning
```
zmitchell, line 287:
```suggestion
We'll be using a small utility for geocoding location names, i.e. turning
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+@@ -0,0 +1,3 @@
++{ lib, config, ... }: {
++
++}
+```
+
+## Submodule types: Declaring map.markers option
+
+One of the most useful types of the module system is the `submodule`
+type. This type allows you to define a nested module system evaluation,
```
zmitchell, line 359: I don't know what these two sentences mean.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
++  options = {
++    map.markers = lib.mkOption {
++      type = lib.types.listOf markerType;
++    };
++  };
+ 
+ }
+```
+
+## Single option namespace: Defining markers in generate.requestParams
```
zmitchell, line 397: I'm not sure what "single option namespace" means here.

source/tutorials/module-system/module-system.md
zmitchell, line 1: Something I've felt as I'm going through this is that I have no sense of where I am in relation to the overall goals of the project i.e. as I'm defining a particular option, how much is there left to do in this tutorial. It feels like a never ending list of "now we're going to do this". There should be some kind of overview at the beginning so that the user can put each step into a larger context.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
+
+Since all modules in `imports` are treated the same, we can also freely
```
zmitchell, line 399: I'm not sure what this sentence means.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
++    generate.requestParams = let
++      paramForMarker = marker:
++        let
++          attributes =
++            [
++              "$(geocode ${
++                lib.escapeShellArg marker.location
++              })"
++            ];
++        in "markers=${
```
zmitchell, line 428: It's kind of hard to understand what's happening here on a language level. Are you joining each attribute with the `\\|` string?

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
...
++        default = {};
++      };
++    };
++  };
++
+ in {
+ 
+   options = {
++
++    users = lib.mkOption {
```
zmitchell, line 502: This is as far as I got before I mentally ran out of gas.

zmitchell (commented):
My overall comment on this is that it's actually two separate tutorials in length. I got about half way through before I mentally ran out of gas, and I only lasted that long because I was trying to complete the review. I suspect a new user will not make it that far before losing interest or needing to take a break. I propose that this is broken into two tutorials:
- Module basics (defining a module, evaluating a module, types, etc)
- Good module APIs (using options to define other options, whatever is in the second half of this tutorial, etc).

fricklerhandwerk:
This thing is a monster and I suspect it's too much work for any of us to break it down into something more palatable. But it IMO should still be out there to have something at all, so we could label it as such and present it as an appendix. We'd still need a briefer introduction to the basics, but that can still be done independently.

source/tutorials/module-system/module-system.md
fricklerhandwerk: I suggest splitting that out into an issue. (GitHub apparently has a UI for that in the `...` menu on the comment.)

source/tutorials/module-system/module-system.md
fricklerhandwerk: If it's explicitly set with `config.zoom = null;`

source/tutorials/module-system/module-system.md
fricklerhandwerk: All of Nix is running on fumes, having gas is clearly a luxury.

source/tutorials/module-system/module-system.md
fricklerhandwerk: Added introductory notes to each sections.

fricklerhandwerk:
@asymmetric @proofconstruction @infinisil @zmitchell this should be ready to go. Take another glance, fire the missiles, and fix forward from here.

zmitchell:
I'll take a look this weekend, I'm pretty busy tomorrow.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
+# Deep dive demo: Wrapping the world in modules
```
asymmetric, line 1: This is too long to fit nicely in the sidebar (see deployed preview)

```suggestion
# Module system
```

asymmetric (changes requested):

source/tutorials/module-system/module-system.md
fricklerhandwerk: I wouldn't want this to pretend to be the definitive module system tutorial. The title should also say what something about what to expect. (Also we can change names later on)

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+```
+
+Here, the label for each `marker` is only propagated to the CLI parameters if `marker.style.label` is set.
+
+## Defining a Default Label
+
+Right now, if a label is not explicitly set, none will show up.
+But since every `users` attribute has a name, we could use that as an automatic value instead.
+
+This `firstUpperAlnum` function allows you to retrieve the first character of the username, with the correct type for passing to `departure.style.label`:
```
asymmetric, line 776: What does `Alnum` mean?

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+             ++ [
++              "color:${marker.style.color}"
+               "$(geocode ${
+                 lib.escapeShellArg marker.location
+               })"
+```
+
+## Marker Styling: Size
+
+In case you set many different markers, it would be helpful to have the ability to change their size individually.
```
asymmetric, line 897: Because this introduces `enum` by itself, I would move this up, before Color, which introduces `enum` inside an `either`

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+
+So far, you've created an option for declaring a *destination* marker, as well as several options for configuring the marker's visual representation.
+
+Now we want to compute and display a route from the user's location to some destination.
+
+The new option defined in the next section will allow you to set an *arrival* marker, which together with a destination allows you to draw *paths* on the map using the new module defined below.
+
+To start, create a new `path.nix` file with the following contents:
+
+```diff
```
asymmetric, line 958:
```suggestion
```nix
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+            builtins.map attrForLocation path.locations;
+        in "path=${
+            lib.concatStringsSep "\\|" attributes
+          }";
+      in builtins.map paramForPath config.map.paths;
+  };
+}
+```
+
+The `path.nix` module defines an option for declaring a list of paths on our `map`, where each path is a list of strings for geographic locations.
```
asymmetric, line 992: This is where the defining/declaring terminology gets confusing.

```suggestion
The `path.nix` module declares an option for defining a list of paths on our `map`, where each path is a list of strings for geographic locations.
```

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
++        default = 5;
++      };
++    };
++  };
++
+   pathType = lib.types.submodule {
+```
+
+:::{note}
+The `ints.between <lower> <upper>` type allows integers in the given (inclusive) range.
```
asymmetric, line 1112: This should be part of the normal flow of the text, similar to every other type we introduced.

asymmetric (changes requested):
- Use of "you" and "we" inconsistently
- I would remove all of the headers that don't directly relate to the module system. E.g. remove "Centering the Map", leave "Depenencies between options". Rationale: people don't actually care about this toy program, they're here to learn the module system 
- I feel like some sections (especially towards the ends) are unnecessary: they don't introduce any radically new concepts, and only serve to extend the already contrived demo example. Examples: "Path styling: color", "Further styling". If someone has gotten that far, they probably know how to look for additional types? If not, we should point them to where to look.
- I've left suggestions and general comments. The general comments are for broader changes, which I will do in a separate PR after this is merged.

source/tutorials/module-system/module-system.md
asymmetric: I mean, it's the *only* module system tutorial? I think I don't understand the argument.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+It may help playing it back alongside the tutorial to better keep track of changes to the code you will work on.
+
+### What will you learn?
+
+In this tutorial you'll learn
+- what a module is
+- how to define one
+- what options are
+- how to declare them
+- how to express dependencies between modules
```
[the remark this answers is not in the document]
fricklerhandwerk: Meh, I'm not a fan of punctuation where visual structure already does the job.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+
+## Empty module
+
+We have to start somewhere.
+The simplest module is just a function that takes any attributes and returns an empty attribute set.
+
+Write the following into a file called `default.nix`:
+
+```nix
+# default.nix
```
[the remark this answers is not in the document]
fricklerhandwerk: Ah yeah thanks, I was hopelessly searching for the documentation because I dimly remembered that was a thing.

fricklerhandwerk:
@asymmetric thanks a lot for the thorough review, a bunch of great catches there. I'll process it when I find time.

> * Use of "you" and "we" inconsistently

I agree. That would need going through all of the text again, to which I'm not really inclined. I suggest fixing that up incrementally later. I'll gladly insta-merge PRs.

> * I would remove all of the headers that don't directly relate to the module system. E.g. remove "Centering the Map", leave "Depenencies between options". Rationale: people don't actually care about this toy program, they're here to learn the module system

Yeah that bothered me too. It would need some thinking about good titles. Do you think it would be good enough as a a follow-up?

> * I feel like some sections (especially towards the ends) are unnecessary: they don't introduce any radically new concepts, and only serve to extend the already contrived demo example. Examples: "Path styling: color", "Further styling". If someone has gotten that far, they probably know how to look for additional types? If not, we should point them to where to look.

That's true, it gets quite repetitive in the last quarter. It would make sense to reframe those as exercises. But it's not really straightforward, because it does require some tinkering to get there without help, so scaling the difficulty to what we assume our target audience will be able to manage will need some testing put into it. I suggest we leave that for later.

> * I've left suggestions and general comments. The general comments are for broader changes, which I will do in a separate PR after this is merged.

Thank you very much!

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+There are several types available under [`lib.types`](https://nixos.org/manual/nixos/stable/#sec-option-types-basic) in the Nixpkgs library.
+
+You have just declared `scripts.output` with the `lines` type, which specifies that the only valid values are strings, and that multiple definitions should be joined with newlines.
+
+:::{note}
+The name and attribute path of the option is arbitrary.
+Here we use `scripts`, because we will add another script later, and call this one `output`, because it will output the resulting map.
+:::
+
+## Evaluating modules
```
infinisil, line 117: This section should be at the very start. `{ }` is not a module, it's just an attribute set. Only calling `evalModules` on it makes it be interpreted as module. So it's weird to talk about modules before `evalModules` is introduced.

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
...
+
+Then define the value for that option where you make the raw script reproducible by wrapping a call to it in `writeShellApplication`:
+
+```diff
+# default.nix
+   config = {
++    scripts.geocode = pkgs.writeShellApplication {
++      name = "geocode";
++      runtimeInputs = with pkgs; [ curl jq ];
++      text = "exec ${./geocode}";
```
infinisil, line 500: Might as well fetch this (and `map`) using Nix, so there's no need for a manual download :)

infinisil (commented):
Not a full review, looking pretty good from what I've seen though!

zmitchell (approved):

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1289 @@
+# Deep dive demo: Wrapping the world in modules
+
+Much of the power in Nixpkgs and NixOS comes from the module system.
+It provides mechanisms for conveniently declaring and automatically merging interdependent attribute sets that follow dynamic type constraints, making it easy to express modular configurations.
+
+## Overview
+
+This tutorial follows [@infinisil](https://github.com/infinisil)'s [presentation on modules](https://infinisil.com/modules.mp4)  ([source](https://github.com/tweag/summer-of-nix-modules)) for  participants of [Summer of Nix](https://github.com/ngi-nix/summer-of-nix) 2021.
```
infinisil, line 8: Fyi I just updated https://infinisil.com/modules.mp4 to use H.264 instead, so this should now easily play in browsers :)

source/tutorials/module-system/module-system.md
```diff
@@ -0,0 +1,1027 @@
+# Module system introduction
+
+Note: This tutorial was created from https://github.com/tweag/summer-of-nix-modules, as presented by @infinisil at Summer of Nix 2021, presentation can be seen here (might not work in browser, `mpv` should work): https://infinisil.com/modules.mp4
```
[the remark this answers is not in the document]
asymmetric: https://github.com/NixOS/nix.dev/issues/785

source/tutorials/module-system/module-system.md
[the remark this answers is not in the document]
asymmetric: https://github.com/NixOS/nix.dev/pull/786

asymmetric:
What do you think of [this way](https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#adding-a-caption) of doing highlights? It's quite time-consuming so would like to gather feedback before proceeding much further.

The advantage is that we get syntax highlighting, because the block is marked as `nix`, instead of `diff`. The downside is we loose the diff -- we don't show what was removed, just the new snippet with highlight. The color can be changed, of course.

### Before

modules-before

### After

modules

fricklerhandwerk:
> What do you think of this way of doing highlights?

It's not great IMO. It loses the one thing which is the whole point of `diff`.

source/tutorials/module-system/module-system.md
fricklerhandwerk:
```suggestion
# Module system deep dive

Or: *Wrapping the world in modules*
```

PR #648, opened 2023-07-16
correct dd command, removing .zst from if=*.img.zst

gerbrent:
the command to image the SD card included `if=nixos-sd-image-......img.zst` though the intention here is to write the non-compressed image to the SD, not the compressed zst image.

.img.zst > .img

As written this command would not function as intended.


proofconstruction:
Thanks for catching this!

Issue #601, opened 2023-06-19
First steps tutorial - Packaging existing software


zmitchell:
## Project description
Migrating a project with an already working build system to a Nix-based workflow can be difficult. Doing a very basic derivation with `runCommand` or `mkDerivation` is covered in #599, but this follows the happy path of not having no dependencies at all outside of the native Rust build tools. This is not the reality for most developers working with Nix.

A more realistic scenario is that developers have a working project, set up a build using `mkDerivation` or some other simple builder, and immediately run into issues regarding missing dependencies. We'd like this tutorial to walk a user through that exact situation e.g. how do you find the dependencies that are missing, how do you know whether to use `some_pkg` vs. `some_pkg.dev` for missing headers or development libraries, how do you add platform-specific dependencies, etc.

## Steps

- [ ] Submit a PR with an outline of the tutorial
- [ ] Submit a PR with a draft of the tutorial

## Considerations
There's obviously many, many directions you could take this tutorial, so special care should be taken to keep the scope small and focus on the most common issues a new Nix user will struggle with when trying to build software with Nix.

PR #650, opened 2023-07-18
add draft of packaging existing software tutorial

proofconstruction:
Closes https://github.com/NixOS/nix.dev/issues/601, or will eventually.

Following (part of) the requirement specification in that Issue, this tutorial currently focuses on the experience of encountering a few common build failures and their solution by passing different attributes to `mkDerivation`. This has the potential to grow substantially if we want to cover more `mkDerivation` features in tutorial style rather than a how-to guide, but I think we should implement that elsewhere as a collection of disconnected (or at least narratively-unrelated) recipes, rather than in this tutorial.

This tries to address most points raised in previous discussions in [Documentation Team](https://discourse.nixos.org/t/2023-07-10-documentation-team-meeting-notes-62/30291) and [Learning Journey Working Group](https://discourse.nixos.org/t/2023-07-13-learning-journey-working-group-meeting-notes-17/30393) meetings, in particular by
- resolving missing dependency errors (requiring use of both `buildInputs` and `nativeBuildInputs`),
- writing a (simple) custom `installPhase` (this project is actually organically "broken", in the sense that [its Makefile](https://github.com/atextor/icat/blob/master/Makefile) does not include an `install` target, which helpfully dispels some of the Nix magic), and
- packaging a non-trivial but uncomplicated project -- `icat`:
  - arose naturally in https://github.com/nixos/nix.dev/pull/645,
  - wasn't already packaged, and
  - is simple enough to use for an intro tutorial

*I explain this choice to use this package in the prose by linking to that not-yet-merged tutorial, so this PR currently should not pass CI; we can remove that justification without harming anything*

Anticipating one comment: the (current) final section where I walk through preparing our package for contribution to `nixpkgs` perhaps should be moved, but I chose to include it here because I believe we should always aim to help users of the configuration features (like the majority of NixOS users) build facility with `mkDerivation` so they can competently use Nix to package their own stuff, as well as growing competent-packagers into contributors and maintainers.

Concerns this doesn't yet address:
- setting cflags, `configure` & `install` prefixes, and `pkg-config` includes
- using `lib`, `dev`, or other outputs for missing headers and development libraries (although `xorg.libX11.dev` is one of the dependencies; maybe the part involving that can be expanded to discuss this briefly?)
- platform-specific dependencies; this `icat` package is too simple to have issues here, and builds fine on x86_64-linux (NixOS), aarch64-linux (NixOS), and aarch64-darwin

*It turns out that the `icat` @infinisil uses in #645 is actually the [`kitty` feature](https://github.com/kovidgoyal/kitty/tree/master/kittens/icat) by the same name; an entirely different project with equivalent functionality*


source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,870 @@
...
+
+Nix is automatically working with the `Makefile` that comes with `icat`, which indeed lacks an `install` target. The `README` in the `icat` repository only mentions using `make` to build the tool, presumably leaving installation up to us. We've now discovered one limit to what Nix can do for us automatically: it doesn't read minds. Fortunately, it does still make the fix quite straightforward to implement.
+
+If you haven't read the tutorials on creating [ad-hoc](https://nix.dev/tutorials/first-steps/dev-environment) or [declarative](https://nix.dev/tutorials/first-steps/declarative-and-reproducible-developer-environments) development environments, do that now before proceeding through the rest of this tutorial; dropping into a `nix-shell` is a crucial component in the Nix user toolbox, and indispensible for debugging.
+
+To enter a useful development shell, we'll pass the dependencies from `nativeBuildInputs` and `buildInputs` to `nix-shell -p`. We'll also make sure to include `git`, so we can clone the `icat` GitHub repository:
```
fricklerhandwerk, line 632: We can also just replace `nix-build` by `nix-shell`, which will avoid any manual steps. Then `$src` will be the source directory, and we can copy stuff out of it when needed.

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,870 @@
...
+```
+
+However, this does not prevent the binary from being produced; an `icat` executable is now present in the local directory, and it's up to us to decide what to do with it.
+
+### installPhase
+In order to make packages available for other packages to depend on, Nix copies everything to the Nix store (at `/nix/store`), and symlinks them from there into build contexts and development environments.
+
+The `Makefile` doesn't provide an installation step, so we must produce one for our derivation, using the [`installPhase` attribute](https://nixos.org/manual/nixpkgs/stable/#ssec-install-phase), which contains a list of command strings to execute to accomplish the installation.
+
+The `icat` executable is only used at runtime, and isn't a compile-time input for anything else at this point, so we only need to concern ourselves with the `bin` output. In Nix, the result of a build is copied to a location stored in the `$out` variable accessible in the derivation's component scripts; we'll create a `bin` directory within that, and then copy our `icat` executable there:
```
fricklerhandwerk, line 665: Don't hint at things that are not relevant yet. Just say we have to copy the build results to the output path and show how to do it.

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,870 @@
...
+  installPhase = ''
+    mkdir -p $out/bin
+    cp icat $out/bin
+  '';
+}
+```
+
+After running our `nix-build` command one last time, we can `ls` in the local directory to find a `result` symlink to the Nix store, with `result/bin/icat` the executable we built. Success!
+
+## Contributing our Work
```
fricklerhandwerk, line 695: A tutorial on making good contributions to Nixpkgs is a high value target, but it would open a real can of worms here. Especially we'd have to clarify what should go into the manual (and how to structure it there) and what should be part of a step-by-step tutorial. The other thing are the million little details that we'd have to get right to make sure this procedure will in fact lead to success. That should absolutely involve Nixpkgs maintainers. I already see a handful of things that won't pass review, but let's not get into that and instead split that part out for now and focus on polishing the tutorial to get as close to what's needed as possible, avoiding additional steps later (e.g. not grab the HEAD of master by default).

fricklerhandwerk (commented):
I tried to focus on the overall structure in this first pass.

The overall structure is great, and I see this becoming a very valuable piece of learning material. The two major conceptual issues I'd like to address first are:
1. There are a few superfluous movements.

  We should guide learners along the happy path with minimal friction, and only run them into problems where there is something to be learned for reuse (such as reading build errors – I very much like how you progress through the build dependencies). Just show them how to do the right thing immediately. Words of warning, highlighting common user error patterns, and lightweight explanations along with links to reference documentation are fine, but they should complement the text, not be an integral part of the narrative.
  
2. The narrative style is a bit too verbose for the purpose here (although I find it hard to separate that requirement from my own taste, so take it with a grain of salt).

  We're dumping a lot of new stuff on learners, and we should rather use words sparingly, only saying what's needed. Not everyone is fluent in English (in fact, most of the world's population), and a loose style, while it may seem easy-going for you or who you imagine your readers to be, will likely impose a significant cognitive overhead. This may sound like it's in conflict with sounding friendly or compassionate, but the real service we're providing here is to make the substance as easy to understand as possible. I'd rather take the risk of sounding cold or impersonal than hiding the essentials between semantically empty words.
  
If we skip the Nixpkgs contribution part, we definitely need a "next steps" section that would point to largely non-existent material. What we also should definitely add are pointers to ongoing discussion around package structure and coding conventions, to note that much of that stuff is still in flux. But would already be polish.

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,567 @@
...
+
+```nix
+# icat.nix
+...
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/bin
+    cp icat $out/bin
+	runHook postInstall
+  '';
```
proofconstruction, line 556:
```suggestion
  installPhase = ''
    runHook preInstall
    mkdir -p $out/bin
    cp icat $out/bin
	  runHook postInstall
  '';
```

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,567 @@
...
+(packaging-existing-software)=
+
+
+# Packaging Existing Software With Nix
+
+One of Nix's primary use-cases is in addressing common difficulties encountered while packaging software, like *managing dependencies*.
+
+In the long term, Nix helps tremendously in alleviating that stress, but when *first* (re)packaging existing software with Nix, it's common to encounter missing dependencies preventing builds from succeeding.
+
+If you haven't already read the tutorial on making a derivation, please go do so before reading this!
```
infinisil, line 18: Should have a link to this tutorial

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,567 @@
...
+We'll begin by considering `hello`, a feature-complete implementation of the famous "hello world", which requires no external dependencies.
+
+Then we'll move to progressively more complex packages with their own separate dependencies, leading us to use additional derivation features.
+
+Along the way, we'll encounter and address Nix error messages, build failures, and a host of other issues, developing our iterative debugging techniques as we go.
+
+## Packages in Nix
+Before we proceed, an important point of clarification: we conventionally use the term "package" by analogy to other systems, although this term is not a proper concept in Nix.
+
+For the purposes of this tutorial, by "package" we mean "a Nix function which takes an attribute set of 'dependencies' and produces a derivation", where "dependencies" could be other packages or configuration parameters.
```
infinisil, line 31: This is a bit confusing, usually derivations themselves are understood as packages, not the functions producing them.

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,567 @@
...
+
+
+(packaging-existing-software)=
+
+
+# Packaging Existing Software With Nix
+
+One of Nix's primary use-cases is in addressing common difficulties encountered while packaging software, like *managing dependencies*.
+
+In the long term, Nix helps tremendously in alleviating that stress, but when *first* (re)packaging existing software with Nix, it's common to encounter missing dependencies preventing builds from succeeding.
```
roberth, line 16: Good expectation management. Some thoughts
- Flaws also include downloading from the internet
- Discovering dependencies by letting the build fail may be a good practice. We don't want to add unused dependencies.

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,567 @@
...
+
+If you haven't already read the tutorial on making a derivation, please go do so before reading this!
+
+In this tutorial, we'll see how to create Nix derivations to package C/C++ software, taking advantage of the [`nixpkgs` `stdenv`](https://nixos.org/manual/nixpkgs/stable/#chap-stdenv) which automates much of the work of building self-contained C/C++ packages.
+
+We'll begin by considering `hello`, a feature-complete implementation of the famous "hello world", which requires no external dependencies.
```
roberth, line 22:
```suggestion
We'll begin by considering `hello`, a feature-complete implementation of the famous "hello world", which requires no dependencies besides what's already in `stdenv`.
```

source/tutorials/packaging-existing-software.md
roberth, line 31: I prefer to define it in terms of the value returned from the function.
You may well define a package without defining any functions:
```nix
with import <nixpkgs> {};
stdenv.mkDerivation {}
```

roberth: (but don't do that)

infinisil (changes requested):
Reviewed together in the Learning journey meeting

Take a look at the notes in https://discourse.nixos.org/t/2023-07-20-learning-journey-working-group-meeting-notes-18/30926#reviewing-nixdev-650-3

@zmitchell: Wording can be more concise, a bit too conversational, can make sentences shorter.
@zmitchell: Take a look at https://nix.dev/contributing/style-guide#voice regarding "we" and "you"

source/tutorials/packaging-existing-software.md
proofconstruction: > I prefer to define it in terms of the value returned from the function.

I changed it to this; thanks

source/tutorials/packaging-existing-software.md
proofconstruction: Removed reference to this

source/tutorials/packaging-existing-software.md
proofconstruction: Clarified this

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+### Fetching Source from GitHub
+While `fetchTarball` required `url` and `sha256` arguments, more are needed for [`fetchFromGitHub`](https://nixos.org/manual/nixpkgs/stable/#fetchfromgithub).
+
+The source is hosted on GitHub at `https://github.com/atextor/icat`, which already gives the first two arguments:
+- `owner`: the name of the account controlling the repository; `owner = "atextor"`
+- `repo`: the name of the repository to fetch; `repo = "icat"`
+
+You can navigate to the project's [Releases page](https://github.com/atextor/icat/releases) to find a suitable `rev`, such as the git commit hash or tag (e.g. `v1.0`) corresponding to the release you want to fetch. In this case, the latest release tag is `v0.5`.
+
+As in the `hello` example, a hash must also be supplied. This time, instead of using `lib.fakeSha256` and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command. You need the SHA256 hash of the *contents* of the tarball, so you will need to pass the `--unpack` and `--type sha256` arguments too:
```
infinisil, line 285:
```suggestion
As in the `hello` example, a hash must also be supplied. This time, instead of using `lib.fakeSha256` and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command. You need the SHA256 hash of the *contents* of the tarball (as opposed to the hash of the tarball file itself), so you will need to pass the `--unpack` and `--type sha256` arguments too:
```

The details here are more intricate (such as the fact that we need this because GitHub occasionally changes the hash of the tarballs, which not all sites do; or how there's no standard directory hash, Nix uses its own NAR format for that; or how there's also `nix-prefetch-github` (and others), but they're not official), but I don't think this needs to be mentioned here.

infinisil: Also, it would be great if you used one line per sentence, [as described in the nix.dev style guidelines](https://nix.dev/contributing/style-guide#one-line-per-sentence), makes reviewing such parts much nicer :smile:

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+# icat.nix
+{ pkgs
+, lib
+, stdenv
+}:
+
+stdenv.mkDerivation {
+  name = "icat";
+
+  src = pkgs.fetchFromGitHub {
```
infinisil, line 270:
```suggestion
{ lib
, stdenv
}:

stdenv.mkDerivation {
  name = "icat";

  src = fetchFromGitHub {
```

> > Don’t have pkgs in the arguments, bad practice.
>
> Why is this bad practice?

Because all `pkgs.<attr>` can be accessed by adding `<attr>` to the argument list. This then also allows people to override `<attr>` using `.override { <attr> = ...; }`

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+       For full logs, run 'nix log /nix/store/l5wz9inkvkf0qhl8kpl39vpg2xfm2qpy-icat.drv'.
+```
+
+A compiler error! The `icat` source was pulled from GitHub, and Nix tried to build what it found, but compilation failed due to a missing dependency: the `imlib2` header. If you [search for `imlib2` on search.nixos.org](https://search.nixos.org/packages?channel=23.05&from=0&size=50&sort=relevance&type=packages&query=imlib2), you'll find that `imlib2` is already in `nixpkgs`.
+
+You can add this package to your build environment by either
+- adding `imlib2` to the set of inputs to the expression in `icat.nix`, and then adding `imlib2` to the list of `buildInputs` in `stdenv.mkDerivation`, or
+- adding `pkgs.imlib2` to the `buildInputs` directly, since `pkgs` is already in-scope.
+
+Because `callPackage` is used to provide all necessary inputs in `nixpkgs` as well as in the `nix-build` invocation, the first approach is the one currently favored, and you should use it here:
```
infinisil, line 369:
```suggestion
You can add this package to your build environment by adding `imlib2` to the set of inputs to the expression in `icat.nix`, and then adding `imlib2` to the list of `buildInputs` in `stdenv.mkDerivation`:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+, stdenv
+, imlib2
+}:
+
+stdenv.mkDerivation {
+  name = "icat";
+
+  src = pkgs.fetchFromGitHub {
+    owner = "atextor";
+	repo = "icat";
```
infinisil, line 384: Indentation needs some cleanup.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+       > make: *** [Makefile:16: icat.o] Error 1
+       For full logs, run 'nix log /nix/store/bw2d4rp2k1l5rg49hds199ma2mz36x47-icat.drv'.
+```
+
+You can see a few warnings which should be corrected in the upstream code, but the important bit for this tutorial is `fatal error: X11/Xlib.h: No such file or directory`: another dependency is missing.
+
+In addition to the widespread practice of prefixing a project name with `lib` to indicate the libraries of that project, in Nixpkgs it's also common to separate headers, libraries, binaries, and documentation into different output attributes of a given [derivation](https://nixos.org/manual/nix/stable/language/derivations.html).
+
+:::{note}
+Determining from where to source a dependency is currently a somewhat-involved process: it helps to become familiar with searching the `nixpkgs` source for keywords, in addition to checking discussion platforms like [the official NixOS Discourse](https://discourse.nixos.org).
```
infinisil, line 442:
```suggestion
Determining from where to source a dependency is currently a somewhat-involved process: it helps to become familiar with searching the `nixpkgs` source for keywords.
```

I don't think it's necessary to mention discourse here, it's not directly useful for this.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+       >    45 | #include <X11/Xlib.h>
+       >       |          ^~~~~~~~~~~~
+       > compilation terminated.
+       > make: *** [Makefile:16: icat.o] Error 1
+       For full logs, run 'nix log /nix/store/bw2d4rp2k1l5rg49hds199ma2mz36x47-icat.drv'.
+```
+
+You can see a few warnings which should be corrected in the upstream code, but the important bit for this tutorial is `fatal error: X11/Xlib.h: No such file or directory`: another dependency is missing.
+
+In addition to the widespread practice of prefixing a project name with `lib` to indicate the libraries of that project, in Nixpkgs it's also common to separate headers, libraries, binaries, and documentation into different output attributes of a given [derivation](https://nixos.org/manual/nix/stable/language/derivations.html).
```
infinisil, line 439:
```suggestion
```

I don't think it should be necessary to talk about multiple outputs here.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+
+:::{note}
+Determining from where to source a dependency is currently a somewhat-involved process: it helps to become familiar with searching the `nixpkgs` source for keywords, in addition to checking discussion platforms like [the official NixOS Discourse](https://discourse.nixos.org).
+:::
+
+You will need the `Xlib.h` headers from the `X11` C package, the Nixpkgs derivation for which is `libX11`, available in the `xorg` package set. The `Xlib` headers in turn live in the `dev` output of `xorg.libX11`. Add this to your derivation's input attribute set and to `buildInputs`:
```
infinisil, line 445:
```suggestion
You will need the `Xlib.h` headers from the `X11` C package, the Nixpkgs derivation for which is `libX11`, available in the `xorg` package set. Add this to your derivation's input attribute set and to `buildInputs`:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+  name = "icat";
+
+  src = pkgs.fetchFromGitHub {
+    owner = "atextor";
+    repo = "icat";
+    rev = "v0.5";
+    sha256 = "0wyy2ksxp95vnh71ybj1bbmqd5ggp13x3mk37pzr99ljs9awy8ka";
+  };
+
+  buildInputs = [ imlib2 xorg.libX11.dev ];
```
infinisil, line 466:
```suggestion
  buildInputs = [ imlib2 xorg.libX11 ];
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+       >       |   ^~~~~~~
+       > icat.c: In function 'main':
+       > icat.c:319:33: warning: ignoring return value of 'write' declared with attribute 'warn_unused_result' [8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wunused-result-Wunused-result8;;]
+       >   319 |                                 write(tempfile, &buf, 1);
+       >       |                                 ^~~~~~~~~~~~~~~~~~~~~~~~
+       > gcc -o icat icat.o -lImlib2
+       > installing
+       > install flags: SHELL=/nix/store/8fv91097mbh5049i9rglc73dx6kjg3qk-bash-5.2-p15/bin/bash install
+       > make: *** No rule to make target 'install'.  Stop.
+       For full logs, run 'nix log /nix/store/x1d79ld8jxqdla5zw2b47d2sl87mf56k-icat.drv'.
```
infinisil, line 517:
```suggestion
```

We can see the full logs, no need to have the repeat of the last 10 lines here. Same for other errors.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+  buildInputs = [ imlib2 xorg.libX11.dev ];
+}
+```
+
+:::{note}
+Only add the top-level `xorg` derivation to the input attrset, rather than the full `xorg.libX11.dev`, as the latter would cause a syntax error. Because Nix is lazily-evaluated, including the dependency this way is safe to do and doesn't actually include all of `xorg` into the build context.
+:::
+
+
+### `buildInputs` and `nativeBuildInputs`
```
infinisil, line 475: Title seems wrong :)

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+
+  installPhase = ''
+    mkdir -p $out/bin
+    cp icat $out/bin
+  '';
+}
+```
+
+### Phases and Hooks
+Nix package derivations are separated into [phases](https://nixos.org/manual/nixpkgs/unstable/#sec-stdenv-phases), each of which is intended to control some aspect of the build process.
```
infinisil, line 558:
```suggestion
Nixpkgs `stdenv.mkDerivation` derivations are separated into [phases](https://nixos.org/manual/nixpkgs/stable/#sec-stdenv-phases), each of which is intended to control some aspect of the build process.
```

This is specific to Nixpkgs, not Nix.

Also we should only link to the stable manual.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+
+You saw earlier how the `stdenv` expected the project's `Makefile` to have an `install` target, and failed when it didn't. To fix this, you defined a custom `installPhase`, containing instructions for copying the `icat` binary to the correct output location, in effect installing it.
```
infinisil, line 560:
```suggestion
You saw earlier how the `stdenv.mkDerivation` expected the project's `Makefile` to have an `install` target, and failed when it didn't. To fix this, you defined a custom `installPhase`, containing instructions for copying the `icat` binary to the correct output location, in effect installing it.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+
+Up to that point, the `stdenv` automatically determined the `buildPhase` information for the `icat` package.
```
infinisil, line 562:
```suggestion
Up to that point, the `stdenv.mkDerivation` automatically determined the `buildPhase` information for the `icat` package.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,589 @@
...
+The hash cannot be known until after the tarball has been downloaded and unpacked, but Nix will complain if the hash supplied to `fetchTarball` was incorrect, so it is common practice to supply a fake one with `lib.fakeSha256` and change the derivation definition after Nix reports the correct hash:
+
+```nix
+# hello.nix
+{ lib
+, stdenv
+}:
+
+stdenv.mkDerivation {
+  src = builtins.fetchTarball {
```
infinisil, line 57: > How should it be done otherwise?

```suggestion
{ lib
, stdenv
, fetchzip
}:

stdenv.mkDerivation {
  src = fetchzip {
```

infinisil (changes requested):
Overall this looks pretty good, nice work! I think we can merge this after the things I mention are addressed

infinisil (approved):
Looking good to me!

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+:::
+
+Now run the `nix-build` command with the new argument:
+
+```console
+$ nix-build -A hello
+error: derivation name missing
+```
+
+This new failure occurs with the *derivation*, further down in the file than the initial error on line 2 about the `lib` argument not having a value; the previous error was successfully resolved by changing the expression passed to `nix-build`.
```
zmitchell, line 121: You can replace this entire sentence with something along the lines of "this is a _new_ error, let's explore where it comes from".

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+### Naming a Derivation
+Every derivation needs a `name` attribute, which must either be set directly or constructed by `mkDerivation` from `pname` and `version` attributes, if they exist.
```
zmitchell, line 124:
```suggestion
Every derivation must have a name which is set via `name` or `pname` and `version` attributes.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+            8|
+            9|   src = fetchzip {
+             |   ^
+           10|     url = "https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz";
+
+       error: hash mismatch in file downloaded from 'https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz':
+         specified: sha256:0000000000000000000000000000000000000000000000000000
+         got:       sha256:0xw6cr5jgi1ir13q6apvrivwmmpr5j8vbymp0x6ll0kcv6366hnn
+```
+
```
zmitchell, line 163:
```suggestion
You're finally back to figuring out the correct hash.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+...
+configure: creating ./config.status
+config.status: creating Makefile
+...
+building
+... <many more lines omitted>
+```
+Great news: the derivation built successfully!
+
+The console output shows that `configure` was called, which produced a `Makefile` that was then used to build the project; it wasn't necessary to write any build instructions in this case, because the `stdenv` build system is based on `autoconf`, which automatically detected the structure of the project directory.
```
zmitchell, line 202:
```suggestion
The console output shows that `configure` was called, which produced a `Makefile` that was then used to build the project.
It wasn't necessary to write any build instructions in this case because the `stdenv` build system is based on `autoconf`, which automatically detected the structure of the project directory.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+### Build Result
+Check your working directory for the result:
+
+```console
+$ ls
+default.nix hello.nix  result
+```
+
+This result is a symbolic link to a Nix store location containing the built binary; you can call `./result/bin/hello` to execute this program:
```
zmitchell, line 212:
```suggestion
This `result` is a symbolic link to a Nix store location containing the built binary; you can call `./result/bin/hello` to execute this program:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+```console
+$ ./result/bin/hello
+Hello, world!
+```
+
+Congratulations, you have successfully packaged your first program with Nix!
+
+Next, you'll package another piece of software with external-to-`stdenv` dependencies that present new challenges, requiring you to make use of more `mkDerivation` features.
```
zmitchell, line 221:
```suggestion
Next, you'll package another piece of software with dependencies that present new challenges, requiring you to make use of more `mkDerivation` features.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+## Something Bigger
+Now you will package a somewhat more complicated program, [`icat`](https://github.com/atextor/icat), which allows you to render images in your terminal.
+
+To start, modify the `default.nix` from the previous section by adding a new attribute for `icat`:
```
zmitchell, line 226: Most builds probably won't be packaging two completely unrelated pieces of software in the same build, so it probably makes more sense to start with a new `default.nix` for `icat`.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+stdenv.mkDerivation {
+  name = "icat";
+
+  src = fetchzip {
+    ...
+  };
+}
+```
+
+Now to download the source code. `icat`'s upstream repository is hosted on [GitHub](https://github.com/atextor/icat), so you should slightly modify the previous [source fetcher](https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers), this time using `fetchFromGitHub` instead of `fetchzip`, updating the argument attribute set to the function accordingly:
```
zmitchell, line 257:
```suggestion
Now to download the source code.
`icat`'s upstream repository is hosted on [GitHub](https://github.com/atextor/icat), so you should modify the previous [source fetcher](https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers), this time using `fetchFromGitHub` instead of `fetchzip`, updating the argument attribute set to the function accordingly:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+    ...
+  };
+}
+```
+
+### Fetching Source from GitHub
+While `fetchzip` required `url` and `sha256` arguments, more are needed for [`fetchFromGitHub`](https://nixos.org/manual/nixpkgs/stable/#fetchfromgithub).
+
+The source is hosted on GitHub at `https://github.com/atextor/icat`, which already gives the first two arguments:
+- `owner`: the name of the account controlling the repository; `owner = "atextor"`
```
zmitchell, line 279:
```suggestion
- `owner`: the name of the account controlling the repository; `owner = "atextor";`
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+- `repo`: the name of the repository to fetch; `repo = "icat"`
```
zmitchell, line 280:
```suggestion
- `repo`: the name of the repository to fetch; `repo = "icat";`
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+You need the SHA256 hash of the *contents* of the tarball (as opposed to the hash of the tarball file itself), so you will need to pass the `--unpack` and `--type sha256` arguments too:
+
+```console
+$ nix-prefetch-url --unpack https://github.com/atextor/icat/archive/refs/tags/v0.5.tar.gz --type sha256
+path is '/nix/store/p8jl1jlqxcsc7ryiazbpm7c1mqb6848b-v0.5.tar.gz'
+0wyy2ksxp95vnh71ybj1bbmqd5ggp13x3mk37pzr99ljs9awy8ka
+```
+
+Now you can supply the correct hash to `fetchFromGitHub`:
```
zmitchell, line 298:
```suggestion
Now you can supply the correct hash to `fetchFromGitHub` from the beginning:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+       >       |   ^~~~~~~
+       > icat.c:39:10: fatal error: Imlib2.h: No such file or directory
+       >    39 | #include <Imlib2.h>
+       >       |          ^~~~~~~~~~
+       > compilation terminated.
+       > make: *** [Makefile:16: icat.o] Error 1
+       For full logs, run 'nix log /nix/store/l5wz9inkvkf0qhl8kpl39vpg2xfm2qpy-icat.drv'.
+```
+
+A compiler error! The `icat` source was pulled from GitHub, and Nix tried to build what it found, but compilation failed due to a missing dependency: the `imlib2` header.
```
zmitchell, line 343:
```suggestion
A compiler error!
The `icat` source was pulled from GitHub, and Nix tried to build what it found, but compilation failed due to a missing dependency: the `imlib2` header.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+If you [search for `imlib2` on search.nixos.org](https://search.nixos.org/packages?channel=23.05&from=0&size=50&sort=relevance&type=packages&query=imlib2), you'll find that `imlib2` is already in `nixpkgs`.
+
+You can add this package to your build environment by adding `imlib2` to the set of inputs to the expression in `icat.nix`, and then adding `imlib2` to the list of `buildInputs` in `stdenv.mkDerivation`:
+
+```nix
+# icat.nix
+{ lib
+, stdenv
+, fetchFromGitHub
+, imlib2
```
zmitchell, line 354:
```suggestion
, imlib2 # new dependency
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+  name = "icat";
+
+  src = fetchFromGitHub {
+    owner = "atextor";
+    repo = "icat";
+    rev = "v0.5";
+    sha256 = "0wyy2ksxp95vnh71ybj1bbmqd5ggp13x3mk37pzr99ljs9awy8ka";
+  };
+
+  buildInputs = [ imlib2 ];
```
zmitchell, line 367:
```suggestion
  buildInputs = [ imlib2 ]; # Add the dependency to the build
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+}
+```
+
+Another error, but compilation proceeds further this time:
```
zmitchell, line 371:
```suggestion
Run `nix-build -A icat` again and you'll encounter another error, but compilation proceeds further this time:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+       > compilation terminated.
+       > make: *** [Makefile:16: icat.o] Error 1
+       For full logs, run 'nix log /nix/store/bw2d4rp2k1l5rg49hds199ma2mz36x47-icat.drv'.
+```
+
+You can see a few warnings which should be corrected in the upstream code, but the important bit for this tutorial is `fatal error: X11/Xlib.h: No such file or directory`: another dependency is missing.
+
+
+:::{note}
+Determining from where to source a dependency is currently a somewhat-involved process: it helps to become familiar with searching the `nixpkgs` source for keywords.
```
zmitchell, line 397: This doesn't actually help a user solve their problem, they're almost guaranteed not to be familiar with `nixpkgs` source. I would recommend to instead tell them how to search `nixpkgs` or whatever it is they _actually_ need to do.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+Add this to your derivation's input attribute set and to `buildInputs`:
+
+```nix
+# icat.nix
+{ lib
+, stdenv
+, fetchFromGitHub
+, imlib2
+, xorg
```
zmitchell, line 410:
```suggestion
, xorg # new dependency
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+  name = "icat";
+
+  src = fetchFromGitHub {
+    owner = "atextor";
+    repo = "icat";
+    rev = "v0.5";
+    sha256 = "0wyy2ksxp95vnh71ybj1bbmqd5ggp13x3mk37pzr99ljs9awy8ka";
+  };
+
+  buildInputs = [ imlib2 xorg.libX11 ];
```
zmitchell, line 423:
```suggestion
  # Note that we only add the libX11 attribute instead of the entire 'xorg' derivation!
  buildInputs = [ imlib2 xorg.libX11 ];
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+}
+```
+
+:::{note}
+Only add the top-level `xorg` derivation to the input attrset, rather than the full `xorg.libX11`, as the latter would cause a syntax error.
+
+Because Nix is lazily-evaluated, including the dependency this way is safe to do and doesn't actually include all of `xorg` into the build context.
```
zmitchell, line 430:
```suggestion
Because Nix is lazily-evaluated, using `xorg.libX11` means that we only include the `libX11` attribute and the derivation doesn't actually include all of `xorg` into the build context.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+### installPhase
+The `stdenv` is automatically working with the `Makefile` that comes with `icat`: you can see in the console output that `configure` and `make` are executed without issue, so the `icat` binary is compiling successfully.
+
+The failure occurs when the `stdenv` attempts to run `make install`: the `Makefile` included in the project happens to lack an `install` target, and the `README` in the `icat` repository only mentions using `make` to build the tool, leaving the installation step up to users.
+
+To add this step to your derivation, use the [`installPhase` attribute](https://nixos.org/manual/nixpkgs/stable/#ssec-install-phase), which contains a list of command strings to execute to perform the installation.
+
+Because the `make` step completes successfully, the `icat` executable is available in the build directory, and you only need to copy it from there to the output directory.
+
+In Nix, this location is stored in the `$out` variable, accessible in the derivation's component scripts; create a `bin` directory within that and copy the `icat` binary there:
```
zmitchell, line 465:
```suggestion
In Nix, the output directory is stored in the `$out` variable, accessible in the derivation's component scripts.
Create a `bin` directory within the `$out` directory and copy the `icat` binary there:
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+    mkdir -p $out/bin
+    cp icat $out/bin
+  '';
+}
+```
+
+### Phases and Hooks
+Nixpkgs `stdenv.mkDerivation` derivations are separated into [phases](https://nixos.org/manual/nixpkgs/stable/#sec-stdenv-phases), each of which is intended to control some aspect of the build process.
+
+You saw earlier how the `stdenv.mkDerivation` expected the project's `Makefile` to have an `install` target, and failed when it didn't.
```
zmitchell, line 498:
```suggestion
You saw earlier how `stdenv.mkDerivation` expected the project's `Makefile` to have an `install` target, and failed when it didn't.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+To fix this, you defined a custom `installPhase`, containing instructions for copying the `icat` binary to the correct output location, in effect installing it.
```
zmitchell, line 500:
```suggestion
To fix this, you defined a custom `installPhase` containing instructions for copying the `icat` binary to the correct output location, in effect installing it.
```

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+...
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/bin
+    cp icat $out/bin
+    runHook postInstall
+  '';
+...
+```
+
```
zmitchell, line 523:
```suggestion
### A successful build
```

Finally building the package is an important step and shouldn't be (accidentally) hidden inside of the "Phases and Hooks" section, where it can be overlooked.

zmitchell (commented):
I've made suggested edits where I thought they were warranted or made the prose flow better, but there's only one or two places where I think the actual content could be adjusted (using `hello.nix` vs. `default.nix`, for example).

Generally I would err on the side of too few words. It's much easier to add words later than it is to condense a paragraph.

source/tutorials/packaging-existing-software.md
```diff
@@ -0,0 +1,567 @@
...
+
+# Packaging Existing Software With Nix
+
+One of Nix's primary use-cases is in addressing common difficulties encountered while packaging software, like *managing dependencies*.
+
+In the long term, Nix helps tremendously in alleviating that stress, but when *first* (re)packaging existing software with Nix, it's common to encounter missing dependencies preventing builds from succeeding.
+
+If you haven't already read the tutorial on making a derivation, please go do so before reading this!
+
+In this tutorial, we'll see how to create Nix derivations to package C/C++ software, taking advantage of the [`nixpkgs` `stdenv`](https://nixos.org/manual/nixpkgs/stable/#chap-stdenv) which automates much of the work of building self-contained C/C++ packages.
```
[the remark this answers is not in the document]
proofconstruction: It's not *specific* to C/C++, but that's all we package here

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+    "keywords": "Nix, packaging"
+---
+
+
+(packaging-existing-software)=
+
+
+# Packaging Existing Software With Nix
+
+One of Nix's primary use-cases is in addressing common difficulties encountered while packaging software, like *managing dependencies*.
```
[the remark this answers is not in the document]
proofconstruction: Originally I used the `master` branch of `icat` which required putting `pkg-config` into `nativeBuildOutputs`, and the idea was to eventually add a third section to this tutorial where we package something else which requires setting `NIX_LDFLAGS` and such to get stuff linked appropriately. Without that, no reason to keep this suggestive emphasis.

source/tutorials/learning-journey/packaging-existing-software.md
proofconstruction: I should learn to use fewer semicolons in this writing.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+An important point of clarification: the term "package" is used conventionally by analogy to other systems, although the term does not refer to a proper concept in Nix.
+
+For the purposes of this tutorial, "package" means something like "result of a derivation"; this is the artifact you or others will use, as a consequence of having "packaged existing software with Nix".
+:::
+
+## A Simple Project
+To start, consider this skeleton derivation:
+
+```nix
+{ stdenv }:
```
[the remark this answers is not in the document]
proofconstruction: Good reminder, though this raises an important question: what is the intended first-reading order of the Learning Journey? Would be helpful to explain that in [`index.md`](https://github.com/NixOS/nix.dev/blob/7b866ebc5e96e5a1d484d04ba518fcdf6a9d48ef/source/tutorials/learning-journey/index.md)

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+
+stdenv.mkDerivation {	};
```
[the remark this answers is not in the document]
proofconstruction: Here too re: https://github.com/NixOS/nix.dev/pull/650#discussion_r1288666805

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+# default.nix
+let
+  pkgs = import <nixpkgs> { };
+in
+{
+  hello = pkgs.callPackage ./hello.nix { };
+}
+```
+
+This allows you to use `nix-build -A hello` to realize the derivation in `hello.nix`, similar to the current convention used in `nixpkgs`.
```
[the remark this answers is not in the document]
proofconstruction: I don't disagree, but it's not *too* much more to type and we eventually want people to help maintain the broader project.

source/tutorials/learning-journey/packaging-existing-software.md
proofconstruction: Agreed, but developing those heuristics is probably another whole tutorial, which we probably could just point to from this note once it exists (if)

source/tutorials/learning-journey/packaging-existing-software.md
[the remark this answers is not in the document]
infinisil: I agree with @proofconstruction, it was also me who originally said that we don't want to use `nix-build -E 'with import <nixpkgs> {}; callPackage ./hello.nix { }`, because that's already an antipattern in Nixpkgs.

And the alternative of using just a single buildable file like this:

```nix
with import <nixpkgs> {};
stdenv.mkDerivation {
  ...
}
```

has the problem that it doesn't support `.override` (`callPackage` adds that), and it would be different from the Nixpkgs convention. It would also mean that if you have multiple packages, Nixpkgs would get imported multiple times and in potentially different ways and different versions.

source/tutorials/learning-journey/packaging-existing-software.md
infinisil: I don't agree, I think showing the `default.nix` with multiple programs is really nice, it teaches the user how files can be reused for packages, and it brings them a step closer to grokking Nixpkgs.

source/tutorials/learning-journey/packaging-existing-software.md
infinisil: Yeah I think this would be better as a separate tutorial, it's really quite involved (grep, `unsafeGetAttrPos`, `nix edit`, the structure of Nixpkgs, backtracking through definitions)

source/tutorials/learning-journey/packaging-existing-software.md
infinisil: `xorg` is not a derivation, but a "package set" (not formally defined, but just an attribute set where the values are packages)

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,531 @@
...
+In this tutorial, you'll create your first Nix derivations to package C/C++ software, taking advantage of the [`nixpkgs` `stdenv`](https://nixos.org/manual/nixpkgs/stable/#chap-stdenv) which automates much of the work of building self-contained C/C++ packages.
+
+The tutorial begins by considering `hello`, an implementation of "hello world" which only requires dependencies already in `stdenv`.
+
+Next, you will build more complex packages with their own dependencies, leading you to use additional derivation features.
+
+You'll encounter and address Nix error messages, build failures, and a host of other issues, developing your iterative debugging techniques along the way.
+
+:::{note}
+An important point of clarification: the term "package" is used conventionally by analogy to other systems, although the term does not refer to a proper concept in Nix.
```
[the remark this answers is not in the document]
infinisil: So Nix doesn't have a package concept, but Nixpkgs definitely does, it's just not formally defined. I think this would be a concise and accurate description:

```suggestion
A _package_ is an informally defined Nixpkgs concept referring to a Nix derivation representing an installation of some project.
Packages have mostly standardised attributes and output layouts, allowing them to be discovered in searches and installed into environments alongside other packages.
```

source/tutorials/learning-journey/packaging-existing-software.md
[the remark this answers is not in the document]
zmitchell: I'm not going to block on this suggestion, I just think it's important to remember that most users don't care about doing things the `nixpkgs` way because it doesn't provide any value _to them_.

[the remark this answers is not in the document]
infinisil: I think there's a lot of value in doing it like this, independently of Nixpkgs:
- It cleanly separates the definition of the package (`hello.nix`) from the importing of its dependencies (Getting some Nixpkgs path in `default.nix`). This is just good practice in general from a separation-of-concerns point-of-view, but it also allows reusing the files in various ways, e.g. you could call the same `hello.nix` with different versions of dependencies, or you could copy the `hello.nix` to `hello-alt.nix` and change it to get a different build with the same dependencies.
- `callPackage` adds the `.override` functionality, which allows easy overriding of dependencies and custom package configuration, fairly clearly separated from the implementation.

But still, we cannot deny Nixpkgs existing, it adds more benefits on top:
- All packages in Nixpkgs are packaged using this very same convention, allowing users to either copy package expressions directly from Nixpkgs, modify and build them, or the other way around, easily upstream packages to Nixpkgs.
- All packages in Nixpkgs have a `.override`, and people are used to being able to change dependencies and configuration using it.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -0,0 +1,535 @@
...
+
+stdenv.mkDerivation {
+  src = fetchzip {
+    url = "https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz";
+    sha256 = lib.fakeSha256;
+  };
+}
+```
+
+Save this file to `hello.nix` and try to build it with `nix-build`, observing your first build failure:
```
zmitchell, line 72:
```suggestion
Save this file to `hello.nix` and try to build it with `nix-build`, observing your first build failure.
Note that this isn't the error you might expect!
In order to see the error message about the SHA256 hash the build has to be functioning, and it's not quite there yet.
```

source/tutorials/learning-journey/packaging-existing-software.md
zmitchell: Discussed online: mention `nix-locate`, we don't mind that it's a third-party tool

zmitchell (approved):
Approved pending one or two changes

infinisil:
Needs a rebase after #671

PR #655, opened 2023-07-20
Update nix-language.md

Carlos-Marques:
description example: (e.g. `file.nix`)

is different from code example: `nix-instantiate --eval test.nix --arg pkgs 'import <nixpkgs> {}'`

this is not consistent with the rest of the document and the change should avoid some minor mistakes


fricklerhandwerk:
Thanks a lot, good catch! Happy to see someone is reading it carefully.

PR #656, opened 2023-07-20
adding note about using etcher when on non-nixos system.

henrik-ch:
small note to make it easier for newcomers on the RPI tutorial.

### DIFF
--- a/source/tutorials/nixos/installing-nixos-on-a-raspberry-pi.md
+++ b/source/tutorials/nixos/installing-nixos-on-a-raspberry-pi.md
@@ -44,6 +44,10 @@ You can download a more recent image from [Hydra](https://hydra.nixos.org/job/ni
 clicking on the latest successful build (marked with a green checkmark), and copying the link to the build product image.
 :::
 
+:::{note}
+It may be more convenient to use a software like [Etcher](https://www.balena.io/etcher/) to flash the image to your SD card if you are on a non-NixOS system.
+:::
+
 Your terminal should be printing kernel messages as they come in.
 
 Plug in your SD card and your terminal should print what device it got assigned, for example `/dev/sdX`.


proofconstruction (approved):
Good idea

source/tutorials/nixos/installing-nixos-on-a-raspberry-pi.md
```diff
@@ -44,6 +44,10 @@ You can download a more recent image from [Hydra](https://hydra.nixos.org/job/ni
 clicking on the latest successful build (marked with a green checkmark), and copying the link to the build product image.
 :::
 
+:::{note}
+It may be more convenient to use a software like [Etcher](https://www.balena.io/etcher/) to flash the image to your SD card if you are on a non-NixOS system.
```
fricklerhandwerk, line 48: I would refrain from proliferating the term "non-NixOS": rationale here https://discourse.nixos.org/t/should-we-give-a-name-to-nix-on-non-nixos/21020/3 

Also I'm not sure what's wrong with `dd` on any system except WSL. It's there, it's fast. The `dmesg` thing is for convenience and could be highlighted as optional. You can also `lsblk` before and after plugging in the storage medium.

proofconstruction: While I agree with these points, I think this just underscores the fact that we shouldn't be maintaining this documentation. But where should it go? The unofficial Wiki already has some (more) information, so I'm in favor of just removing this page entirely. If there is any information in here that isn't in the Wiki, we could submit the diff there, and wash our hands of it.

henrik-ch: > I would refrain from proliferating the term "non-NixOS": rationale here https://discourse.nixos.org/t/should-we-give-a-name-to-nix-on-non-nixos/21020/3

Thanks for that link Valentin - that makes sense - I am thinking of making a small PR to edit the non-NixOS part out - will link that PR to this one.

> Also I'm not sure what's wrong with `dd` on any system (except WSL, where you'd use some Windows stuff for that anyway). It's there, it's fast. The `dmesg` thing is for convenience and could be highlighted as optional. You can also `lsblk` before and after plugging in the storage medium. I don't like introducing more moving parts to the process.

I can see the point of not introducing more than one tool to do the job. 
I postulate that for a beginner, using Etcher is a lot more straightforward than flashing the card from the command line.
To counter that argument we can say that people that aren't comfortable in the command line shouldn't touch Nix and NixOS in the first place, but I think that we are unnecessarily excluding less experienced users by taking that approach.

PR #660, opened 2023-07-23
Fix minor wording error

R0b0ty:
This minor wording error is found in the section titled, "**Create a shell environment**", where the verbs "Press" and "type" are adjacent to each other.  The proposed fix removes the "Press" verb altogether.


henrik-ch:
Thank you for spotting - now fixed in #661

PR #666, opened 2023-07-28
Fix typo in faq.md

michaelshmitty:
Just wanted to suggest a quick typo fix without too much overhead. Is this an okay way to do it or should this be more involved?


michaelshmitty:
Also excellent work on this documentation! I've been doing a refresher course on using Nix and nix.dev is a huge improvement to the documentation compared to when I first started using and learning Nix a few years ago. 👍

fricklerhandwerk (approved):
Thanks very much! Direct fixes are the best.

PR #672, opened 2023-08-10
introduce automatic testing for `nix-repl` blocks

mightyiam:


mightyiam:
The doctester app seems to fail when executed in the build phase. I'm guessing this has something to do with failing to creating a file that is part of creating a PTY.

Perhaps as a flake "check" this would be run in an environment that would allow such files to be created?

Can something be done otherwise to allow these files to be created?

mightyiam:
Failure:
https://github.com/NixOS/nix.dev/actions/runs/5832308481/job/15817476652?pr=672#step:5:907

Relevant crate:
https://docs.rs/pty-process/latest/pty_process/index.html

mightyiam:
Worked around this by making this a flake app, instead.

mightyiam:
Regarding the fenced code block language tag, the following

<pre>```text nix-repl</pre>

would work, but it's not nice. I'd like to use the Sphinx API to create an extension that prepends `text ` to `nix-repl`, avoiding the issue.

mightyiam:
I feel that there is some open source rust crate potential somewhere in this but haven't thought of where to draw that line yet.

mightyiam:
Yes, there are no tests. I wonder whether anyone will insist on tests...

infinisil (changes requested):
We should consider moving the Rust project into a separate repository, it should also definitely have tests and code comments and somebody should take a somewhat closer look at the code.

domenkozar:
Have you seen https://github.com/OceanSprint/tesh

mightyiam:
Well... today is the last day of https://thaigersprint.org/, during which I've been working primarily on this. [The documentation example tester](https://github.com/mobusoperandi/eelco) has been extracted and released. And I've updated this pull request.

All examples are skipped except for two. Those two are repl examples.

As it stands in this PR branch, repl examples and expression examples can be written and/or edited, as long as they're within the limited feature set of eelco. That is, channels cannot be used. Examples that consist of multiple files are not supported. Could be other things we wanna do that are not supported.

mightyiam:
Should this be merged at this stage? Is it better than nothing? Is the design of eelco reasonable?

source/tutorials/nix-language.md
```diff
@@ -109,14 +109,13 @@ The following example is a Nix expression adding two numbers:
 
 #### Interactive evaluation
 
-Use [`nix repl`] to evaluate Nix expressions interactively (by typing them on the command line):
-
-```shell-session
-$ nix repl
-Welcome to Nix 2.13.3. Type :? for help.
+Use the Nix REPL to evaluate Nix expressions interactively (by typing them on the command line).
+Execute it as `$ nix repl`, and you have a repl:
```
fricklerhandwerk, line 113: That is not an improvement because now there is no link to reference docs. It was good as it was.

source/tutorials/nix-language.md
```diff
@@ -109,14 +109,13 @@ The following example is a Nix expression adding two numbers:
...
 
+```nix-repl
 nix-repl> 1 + 2
 3
+
```
fricklerhandwerk, line 118:
```suggestion
```

Why is there an empty line?

source/guides/best-practices.md
```diff
@@ -158,7 +158,7 @@ On NixOS, `$NIX_PATH` can be set permanently with the [`nix.nixPath`](https://se
 
 To quickly obtain packages for demonstration, we use the following concise pattern:
 
-```nix
+```nix skip
```
fricklerhandwerk, line 161: Since you're only testing nix-repl blocks, there should be no need to spam every other block with redundant information (the purpose of which is not evident without "just knowing" what it means).

fricklerhandwerk (commented):
I still have to play with it to see how it behaves in case of errors. From the consumer perspective I'd expect it not to require changing anything about the documentation source though.

fricklerhandwerk:
Closing this for now as it hasn't seen any progress in a long time and is far out of date. Feel free to reopen if there is any notable update.

mightyiam:
Sorry for taking so long.

PR #679, opened 2023-08-18
Add line numbers to Nix, Python code blocks

asymmetric:
These could also be referenced in an article's prose.

For an example, see https://line-nos.nix-dot-dev.pages.dev/tutorials/learning-journey/packaging-existing-software#a-new-command


infinisil (approved):
Reviewed during the docs team call today, looking good!

infinisil:
Using the copy button only copies the first line, but that was also a problem before. The line numbers can't be selected which is good.

Issue #481, opened 2023-03-20
Transfer ownership of infrastructure


zmitchell:
If we want to implement new infrastructure or make changes we need to transfer ownership of the infrastructure. We also need to decide who to transfer ownership _to_.

- [x] DNS
- [x] web hosting provider ([looks like Cloudflare](https://github.com/NixOS/nix.dev/issues/319))
- [x] newsletter
- [ ] [`CODEOWNERS`](https://github.com/NixOS/nix.dev/blob/master/.github/CODEOWNERS)
- [x] Cachix credentials

Depends on: https://github.com/NixOS/nix.dev/pull/262

asymmetric:
@zmitchell we should list what infrastructure we're talking about.

What I can think of:

- [ ] DNS
- [ ] web hosting provider
- [ ] newsletter

fricklerhandwerk:
I moved the list to the issue description. I also removed the `make-official` label, because there is either endless proliferation or losing history once they become obsolete. We should use tracking issues consistently to capture task closures.

domenkozar:
Done!

asymmetric:
@domenkozar as shown by the linked PR, the transfer is still not 100% complete :sweat_smile:

domenkozar:
I don't think it's related to the transfer, but it you want to keep it open that's fine.

asymmetric:
You're right, I'll close this and open a separate issue for the `CODEOWNERS`.

PR #680, opened 2023-08-18
`CODEOWNERS`: remove domenkozar, fricklerhandwerk

asymmetric:
The reasoning is:
- domenkozar has transferred ownership of the infrastructure #481
- unless someone is a team lead on a specific sub-team (as is the case for zmitchell), we (people with write access to nix.dev) should all be able to merge PRs

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,44 +11,8 @@
 # This also holds true for GitHub teams. Since almost none of our teams have write
 # permissions, you need to list all members of the team with commit access individually.
 
-# This file
-.github/CODEOWNERS @domenkozar
+* @NixOS/documentation-team
 
-# Dependabot
-.github/dependabot.yml @domenkozar
-
-# GitHub Actions
-.github/workflows/ @domenkozar
-
-# Meta-documentation
-CONTRIBUTING.md @domenkozar
-LICENSE.md @domenkozar
-README.md @domenkozar @fricklerhandwerk
-cla/ @domenkozar
-
-# GitPod
-.gitpod* @domenkozar
-
-# gitignore
-.gitignore @domenkozar
-
-# Nix machinery
-*.nix @domenkozar
-flake.lock @domenkozar
-
-# Python machinery
-poetry.lock @domenkozar
-pyproject.toml @domenkozar
-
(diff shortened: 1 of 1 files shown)


infinisil:
> * we (people with write access to nix.dev) should all be able to merge PRs

GitHub's code owners feature doesn't have an influence on who can merge PR's unless you enable it (which is not the case I'm pretty sure). So it's really just an easy way to automatically request reviews from specific people when certain files change.

This can be thought of a mechanism to signal who's responsible for which part of the code, and as such it would be great to always have at least one code owner for every file.

asymmetric:
@infinisil oops you're right, this doesn't currently block merges (I could even self-merge this PR).

> So it's really just an easy way to automatically request reviews from specific people when certain files change.

OK, but then should we really automatically ask reviews of @fricklerhandwerk and @domenkozar for everything under `source/`? I mean, they probably added themselves, so :shrug: 

@proofconstruction This can be thought of a mechanism to signal who's responsible for which part of the code

Agreed. So who's responsible for what? In my mind, it's pretty clear-cut with @zmitchell. As for the rest, I'm not so sure.

proofconstruction:
Now I'm thinking that Domen should remain in CODEOWNERS in those places *if he chooses to continue being responsible for those things*, and we should consider just leaving the file as-is, close this PR, re-close the infrastructure ownership one, and then open a separate Issue to discuss this further so everyone interested can have a say over a longer period (maybe 2 weeks?).

> @proofconstruction Agreed. So who's responsible for what? In my mind, it's pretty clear-cut with @zmitchell. As for the rest, I'm not so sure.

I don't know either. Nobody has edited the file to add themselves and take ownership of anything. Part of that is probably a broader lack of a concrete plan, or at least lack of focus on [this plan](https://github.com/NixOS/nix.dev/blob/master/maintainers/documentation-project-2023.md) (which may already be complete as-written; we should discuss the evaluation & final report soon), and partly a lack of resources (funding, time, energy, attention, etc) on the part of individuals. Speaking for myself, I'd be open to taking ownership of some pieces but I don't know which ones (they may not exist yet) and I haven't been very good so far about taking care of the things I *have* said I'd do (attention is *my* scarcest resource).

PR #681, opened 2023-08-19
Proposal for a table overview in `Shell.nix` tutorial

VuiMuich:
As mentioned on reddit IMO it would be nice to have a condened overview as a table at the beginning.

(Draft as this is just a stylistic proposal as discussion basis and eventually should porbably be expanded on all tutorial articles.)

### DIFF
--- a/source/tutorials/learning-journey/shell-dot-nix.md
+++ b/source/tutorials/learning-journey/shell-dot-nix.md
@@ -4,20 +4,12 @@
 
 ## Overview
 
-### What will you learn?
-
 <!-- Give a brief description of what the reader will learn so that they know whether the topic interests them. -->
-How to create and configure reproducible shell environments
-
-### How long will it take?
-30 minutes
-
 <!-- Give some indication of how long it will take to complete the tutorial so that the reader knows whether to continue. -->
-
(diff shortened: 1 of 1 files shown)


zmitchell:
I'm generally in support of this, but the exact format could use some bikeshedding since we'd want it to look nice regardless of the length of the content in each part of the table.

VuiMuich:
As far as I can tell the mark down renderer is responsible for line wrapping the individual cells.
Tbh I have no idea if or how it would be possible to enforce a static width of the columns, is this possible in the custom css?
I tested with `glow` to render in the terminal and it automatically to a relative compact table with three lines and with `litemdview` which offers a very nice responsive result.

VuiMuich:
Also was considering adding the tooltip link to the time itself and changing the clock for an icon more representing "intercativity":

[📖]: ## "Reading time"
[🕹]: ## "Interactive time"

| [📖]     | [🕹]     | What will you learn?                                        | What will you need?                       |
| :---:  | :---:  | :---                                                        | :---                                      |
| [10 min][📖] | [30 min][🕹] | How to create and configure reproducible shell environments | A basic understanding of the Nix language |

PR #695, opened 2023-08-24
Update packaging existing software tutorial

proofconstruction:
We discussed feedback from Discourse and Reddit in today's docs team meetings, and decided it would be best to drop the `name` attribute and go straight for `pname` and `version` in *Packaging Existing Software with Nix*.

### DIFF
--- a/source/tutorials/learning-journey/packaging-existing-software.md
+++ b/source/tutorials/learning-journey/packaging-existing-software.md
@@ -46,12 +46,21 @@ As you progress through this tutorial, you will update this several times, addin
 ### Hello, World!
 GNU Hello is an implementation of the "hello world" program, with source code accessible [from the GNU Project's FTP server](https://ftp.gnu.org/gnu/hello/).
 
-To begin, you should add a `name` attribute to the set passed to `mkDerivation`; every derivation needs a name, and Nix will throw `error: derivation name missing` without one.
+Every derivation needs a package name and version, so to begin, you should add a `pname` attribute and a `version` string to the set passed to `mkDerivation`.
+
+The most recent version of `hello` available on the FTP server is 2.12.1.
+
+:::{note}
+For historical reasons, Nix uses the `pname` and `version` attributes to construct a `name` attribute.
+
+If either of `pname` or `version` are missing, `name` cannot be produced, and Nix will throw `error: derivation name missing`.
+:::
 
 ```diff
 ...
 stdenv.mkDerivation {
-+ name = "hello";
++ pname = "hello";
++ version = "2.12.1";
 ...
 ```
 
@@ -71,7 +80,8 @@ The hash cannot be known until after the tarball has been downloaded and unpacke
 }:
 
 stdenv.mkDerivation {
-  name = "hello";
+  pname = "hello";
+  version = "2.12.1";
 
   src = fetchzip {
     url = "https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz";
@@ -219,7 +229,7 @@ in
 }
 ```
 
-Now copy `hello.nix` to a new file, `icat.nix`, and update the `name` attribute in that file:
+Now copy `hello.nix` to a new file, `icat.nix`, and update the `pname` attribute in that file, leaving the `version` string unchanged for now:
 
 ```nix
 # icat.nix
@@ -229,7 +239,8 @@ Now copy `hello.nix` to a new file, `icat.nix`, and update the `name` attribute 
 }:
 
 stdenv.mkDerivation {
-  name = "icat";
+  pname = "icat";
+  version = "2.12.1";
 
   src = fetchzip {
     ...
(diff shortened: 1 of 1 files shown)


source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -46,12 +46,21 @@ As you progress through this tutorial, you will update this several times, addin
 ### Hello, World!
 GNU Hello is an implementation of the "hello world" program, with source code accessible [from the GNU Project's FTP server](https://ftp.gnu.org/gnu/hello/).
 
-To begin, you should add a `name` attribute to the set passed to `mkDerivation`; every derivation needs a name, and Nix will throw `error: derivation name missing` without one.
+Every derivation needs a package name and version, so to begin, you should add a `pname` attribute and a `version` string to the set passed to `mkDerivation`.
```
zmitchell, line 49: Nitpick: both `pname` and `version` are attributes. Referring to one as an attribute and the other as a string is slightly misleading.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -300,8 +313,34 @@ stdenv.mkDerivation {
 }
 ```
 
+Now is a good time to change the `version` string also.
+
+To avoid repeating yourself, and to make later updates to the package a little bit easier, you should set the `version` to `0.5` and then use [string interpolation](https://nix.dev/tutorials/first-steps/nix-language#string-interpolation) in the `rev` attribute passed to `fetchFromGitHub`.
+
+For `version` to be accessible to `rev`, the attribute set passed to `stdenv.mkDerivation` must be [*recursive*](https://nix.dev/tutorials/first-steps/nix-language#recursive-attribute-set-rec), so you should also add the `rec` keyword to avoid seeing `error: undefined variable 'version':
```
zmitchell, line 320: It may be better practice to use a `let` binding instead of `rec` since `rec` can make it hard to trace data flow.

zmitchell (approved):

source/tutorials/learning-journey/packaging-existing-software.md
proofconstruction:
```suggestion
Every derivation needs a package name and version, so to begin, you should add `pname` and `version` attributes to the set passed to `mkDerivation`, setting these to appropriate strings.
```

Good point. How about this?

source/tutorials/learning-journey/packaging-existing-software.md
proofconstruction: Normally I would, but last time I submitted, I was [told not to use `let`](https://github.com/NixOS/nixpkgs/pull/239035#discussion_r1240091786) :upside_down_face:

proofconstruction: Ah, so we're actually supposed to do [this](https://nixos.org/manual/nixpkgs/stable/#mkderivation-recursive-attributes)

source/tutorials/learning-journey/packaging-existing-software.md
zmitchell: Looks good to me 👍

source/tutorials/learning-journey/packaging-existing-software.md
zmitchell: Sigh

Maybe you're "supposed" to do the last one, but that's so much more complicated than the other solutions that I'm inclined to say just throw your hands up and use `rec`.

source/tutorials/learning-journey/packaging-existing-software.md
```diff
@@ -46,12 +46,21 @@ As you progress through this tutorial, you will update this several times, addin
 ### Hello, World!
 GNU Hello is an implementation of the "hello world" program, with source code accessible [from the GNU Project's FTP server](https://ftp.gnu.org/gnu/hello/).
 
-To begin, you should add a `name` attribute to the set passed to `mkDerivation`; every derivation needs a name, and Nix will throw `error: derivation name missing` without one.
+Every derivation needs a package name and version, so to begin, you should add a `pname` attribute and a `version` string to the set passed to `mkDerivation`.
+
+The most recent version of `hello` available on the FTP server is 2.12.1.
+
+:::{note}
+For historical reasons, Nix uses the `pname` and `version` attributes to construct a `name` attribute.
```
fricklerhandwerk, line 54:
```suggestion
For historical reasons, Nix uses the `pname` and `version` attributes to construct a `name` attribute passed to the underlying call to [`builtins.derivation`]().
```

Please add a link to the Nix manual

source/tutorials/learning-journey/packaging-existing-software.md
fricklerhandwerk: `finalAttrs` is absolutely the right way to do it and severely underdocumented. In suppose has to be taught somewhat separately because it's tied into overrides, but for now we could do a compromise and use either let or rec with a note that there also exist finalAttrs, and then later reorder the tutorials to start with consuming existing packages. Since I took it upon me to extend the curriculum running into these issues is valuable input.

fricklerhandwerk:
This is outdated. Most fixes were already applied in other PRs.

PR #704, opened 2023-09-03
Use bold instead of monospace for emphasis

J3RN:
It appears that in the `towards-reproducibility-pinning-nixpkgs.md` page, the monospace code formatting is used for emphasis.  This is not the intended use of this formatting, and the rest of the page uses bold for emphasis.


infinisil (approved):
Thanks!

PR #701, opened 2023-08-28
implement the vision

proofconstruction:
This PR implements the grand vision [here](https://github.com/NixOS/nix.dev/blob/master/CONTRIBUTING.md#vision)

Just moved some stuff around

Done with :heart: in the Docs Team Meeting with @fricklerhandwerk @alejandrosame

PR #725, opened 2023-09-11
simple server client test

olafklingt:
proposal to replace the complicated server client test with a simple test.
... unfortunately its so simple that one doesn't have to deal with network issues at all.

### DIFF
--- a/source/tutorials/nixos/integration-testing-using-virtual-machines.md
+++ b/source/tutorials/nixos/integration-testing-using-virtual-machines.md
@@ -1,6 +1,6 @@
 (integration-testing-vms)=
 
-# Integration testing using virtual machines
+# Integration testing with NixOS virtual machines
 
 This tutorial aims to be compatible with NixOS release 22.11.
 
--- a/source/tutorials/nixos/integration-testing-using-virtual-machines.md
+++ b/source/tutorials/nixos/integration-testing-using-virtual-machines.md
@@ -254,13 +254,11 @@ result=$(readlink -f ./result) rm ./result && nix-store --delete $result
 
 ## Tests with multiple virtual machines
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
This is a very good idea. A tutorial should focus on the principle. But before we remove the example altogether, let's either move it to the manual where all examples are collected, or link to more third-party sources, so it's more likely for learners to find something that best matches their use case.

olafklingt:
> But before we remove the example altogether, let's either move it to the manual where all examples are collected, or link to more third-party sources, 

Is the Manual the right place for such recipes?
I put two examples that i wrote before on the wiki:
https://nixos.wiki/wiki/NixOS_Testing_library

fricklerhandwerk:
> Is the Manual the right place for such recipes?

I'd say it depends how you frame it. If it's a pure example, it should be in the manuals. Having a small collection there would be good.

A recipe would summarize the pattern and link to specific reference documentation for details, so it doesn't really need fully working examples.

Given the introduction in the tutorial already shows the pattern, we may leave it at that for now and consider reproducing it in isolation in the recipes section for better discoverability later.

olafklingt:
Ok I moved the example derived from @Gabriella439 's blog to the wiki.

fricklerhandwerk:
@olafklingt Thank you very much for the thorough research! Although the diff got quite large, in essence it's just polish. Each small change is documented (and hopefully well-enough explained) in the commit log.

This is really a great piece of exposition to an extremely powerful feature. Once we get #701 in, it will also have its due place in the learning journey. I'd like to merge it as it is now, but a brief last check from the @NixOS/documentation-team won't hurt.

fricklerhandwerk (approved):

PR #726, opened 2023-09-12
homepage: put link to first steps at the top.

GuillaumeDesforges:
In my opinion, the link to the first steps should be the first thing a new user should see.

Indeed, the first actionnable thing for a new user should be to "just get started".

I think that putting any link before means that there are all sorts of "paths" proposed, instead of one clear path.

If possible, I would go further and make the call to action even more visible. Usually it is done with a button-like UI.

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -9,7 +9,9 @@ myst:
 
 # Welcome to nix.dev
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
That's good. The assorted things should be replaced by a concise technically oriented description of what Nix is about.

PR #728, opened 2023-09-13
change "build task" to "derivation" in tutorial

g-re-g:
When reading this document as a newcomer I have no preconceived notion of what a "build task" is. This tutorial is trying to explain what a "derivation" is but then uses the term "build task" in several locations. I don't know if there is a difference between the two terms but the official terminology seems to be "derivation". I don't see any reason to confuse readers by interchanging between the two terms. 

Similarly, if there _is_ a difference between a "build task" and a "derivation" it would be helpful to explain that somewhere.

### DIFF
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -1859,7 +1859,6 @@ It is an error if the network request fails.
 (derivations)=
 ## Derivations
 
-A build task in Nix is called a *derivation*.
 
 Derivations are at the core of both Nix and the Nix language:
 - The Nix language is used to describe derivations.
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -1917,7 +1917,7 @@ There is more depth to the inner workings of derivations, but at this point it s
 
 :::
 
-String interpolation on derivations is used to refer to other build results as file system paths when declaring new derivations.
+String interpolation on derivations is used to refer to their build results as file system paths when declaring new derivations.
 
 This allows constructing arbitrarily complex compositions of derivations with the Nix language.
 
--- a/source/tutorials/first-steps/nix-language.md
+++ b/source/tutorials/first-steps/nix-language.md
@@ -1739,13 +1739,13 @@ For historical reasons, some of the functions in `pkgs.lib` are equivalent to [`
 So far we have only covered what we call *pure expressions*:
 declaring data and transforming it with functions.
 
-In practice, describing build tasks requires observing the outside world.
+In practice, describing derivations requires observing the outside world.
 
 There is only one impurity in the Nix language that is relevant here:
 reading files from the file system as *build inputs*
 
-Build inputs are files that build tasks refer to in order to describe how to derive new files.
-When run, a build task will only have access to explicitly declared build inputs.
+Build inputs are files that derivations refer to in order to describe how to derive new files.
+When run, a derivation will only have access to explicitly declared build inputs.
 
 The only way to specify build inputs in the Nix language is explicitly with:
 
@@ -1861,12 +1861,12 @@ It is an error if the network request fails.
 
 A build task in Nix is called a *derivation*.
 
-Build tasks are at the core of both Nix and the Nix language:
-- The Nix language is used to describe build tasks.
-- Nix runs build tasks to produce *build results*.
-- Build results can in turn be used as inputs for other build tasks.
+Derivations are at the core of both Nix and the Nix language:
+- The Nix language is used to describe derivations.
+- Nix runs derivations to produce *build results*.
+- Build results can in turn be used as inputs for other derivations.
 
-The Nix language primitive to declare a build task is the built-in impure function `derivation`.
+The Nix language primitive to declare a derivation is the built-in impure function `derivation`.
 
 It is usually wrapped by the Nixpkgs build mechanism `stdenv.mkDerivation`, which hides much of the complexity involved in non-trivial build procedures.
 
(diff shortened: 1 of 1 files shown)


source/tutorials/first-steps/nix-language.md
```diff
@@ -1861,12 +1861,12 @@ It is an error if the network request fails.
 
 A build task in Nix is called a *derivation*.
```
g-re-g, line 1862: This description should probably be changed as I suspect many folks, myself included, don't know what a "build task" is and so the definition feels circularly incomplete. As if a dictionary said:

Derivation (n) : See _build task_

But then there was no entry for _build task_

fricklerhandwerk:
```suggestion
```

Let's remove it then.

source/tutorials/first-steps/nix-language.md
```diff
@@ -1918,7 +1918,7 @@ There is more depth to the inner workings of derivations, but at this point it s
 
 :::
 
-String interpolation on derivations is used to refer to other build results as file system paths when declaring new build tasks.
+String interpolation on derivations is used to refer to other build results as file system paths when declaring new derivations.
```
fricklerhandwerk, line 1921:
```suggestion
String interpolation on derivations is used to refer to their build results as file system paths when declaring new derivations.
```

fricklerhandwerk (approved):
Thanks for the proposal. We discussed this with the @NixOS/documentation-team and agree this is a reasonable change. A few small fixups. 

Please also change the [first occurrence of build task](https://github.com/NixOS/nix.dev/pull/728/files#diff-e07c1c53347d65c8d474c94912ab24825952ee9b2cab073d8699afdfc6464934R59) (the link will do a funny thing, expanding the folded sections *very slowly*) to:

> Derivations that describe how to obtain build results

We may even eventually go back to calling things "build outputs" or even just "inputs" and "outputs", but it gets ever more vague without context. But let's just keep it consistent for now. The original idea to say "build task" instead of derivation everywhere was the hope that it would be more intuitive for people with some familiarity with build systems. Still to this day, there is no plain-language definition of derivations, but we're getting there slowly.

g-re-g:
> Thanks for the proposal. We discussed this with the @NixOS/documentation-team and agree this is a reasonable change. A few small fixups.
> 
> Please also change the [first occurrence of build task](https://github.com/NixOS/nix.dev/pull/728/files#diff-e07c1c53347d65c8d474c94912ab24825952ee9b2cab073d8699afdfc6464934R59) (the link will do a funny thing, expanding the folded sections _very slowly_) to:
> 
> > Derivations that describe how to obtain build results
> 
> We may even eventually go back to calling things "build outputs" or even just "inputs" and "outputs", but it gets ever more vague without context. But let's just keep it consistent for now. The original idea to say "build task" instead of derivation everywhere was the hope that it would be more intuitive for people with some familiarity with build systems. Still to this day, there is no plain-language definition of derivations, but we're getting there slowly.

As a complete newbie, in general, it is fine to have these words that don't come with domain baggage like "derivation". As long as they are clearly explained, it is easier for me to pick up a new word than it is to add a new definition to an existing word that subtly differs from the other definitions it has. To that extent, "derivation" is probably a great word for Nix as, at least in this domain, it is very Nix specific and so, with clear definition, it should be easy to convey the idea to other folks that we are talking about something Nix specific. What's maybe a little strange is that in practice, it seems that the specification for producing software is called the derivation, and the result of acting on a derivation is called a "build output". I think common linguistics would have me think that what is currently called a derivation is more appropriately called a "derivation specification" and the result of performing nix magic upon that specification, which is currently called the "build output" should be called the "derivation".

Currently:
A derivation gets processed by Nix to produce a build output.

What seems more linguistically consistent:
A derivation specification gets processed by Nix to produce a derivation.

fricklerhandwerk:
It's still lacking the first mention of "build task" near the beginning.

source/tutorials/first-steps/nix-language.md
```diff
@@ -1739,13 +1739,13 @@ For historical reasons, some of the functions in `pkgs.lib` are equivalent to [`
...
-In practice, describing build tasks requires observing the outside world.
+In practice, describing derivations requires observing the outside world.
 
 There is only one impurity in the Nix language that is relevant here:
 reading files from the file system as *build inputs*
 
-Build inputs are files that build tasks refer to in order to describe how to derive new files.
-When run, a build task will only have access to explicitly declared build inputs.
+Build inputs are files that derivations refer to in order to describe how to derive new files.
+When run, a derivation will only have access to explicitly declared build inputs.
```
asymmetric, line 1748: I think here and below we should replace "run" with something else. Maybe "evaluated"?

asymmetric (approved):
This LGTM, other than the small nit.

fricklerhandwerk:
> "Build task" is mentioned in the Nix manual's [glossary entry](https://nixos.org/manual/nix/stable/glossary#gloss-derivation) for derivation, as well as in other places (`nix-store --help --realise`).

We can change that as well. The glossary entry should be next to rework now that [the main page for derivations was updated](https://nixos.org/manual/nix/unstable/language/derivations).

source/tutorials/first-steps/nix-language.md
fricklerhandwerk: That will produce confusion with evaluating Nix expressions. The technically correct phrasing is ["running the `builder` executable of a derivation"](https://nixos.org/manual/nix/unstable/glossary#gloss-realise). That is usually condensed to "building a derivation", but I think it's somewhat misleading because it's both easy to confuse with "realising a derivation" (which can mean more things) and because "building" has the same semantic baggage as "build task".

Maybe we can in fact establish a more suitable term, but we shouldn't block this PR on that.

PR #729, opened 2023-09-14
move faqs from recipes to concepts and edit 'Should I enable flakes?'

wamirez:
Made some minor changes to:

- [Should I enable flakes?](https://nix.dev/recipes/faq#should-i-enable-flakes)

such as adding links and correcting typos.


source/concepts/faq.md
```diff
@@ -24,12 +22,31 @@ While flakes reduce complexity in some regards, they introduce additional concep
 There are downsides to relying on [experimental features](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-experimental-features) in general:
 
 - Interfaces and behavior of experimental features could still be changed by Nix developers.
-  This may require you to adapt your code at some point in the future, which will be more effort when it has grown in complexity.
-  Currently there is no agreed-upon plan or timeline for stabilising flakes.
+  This may require you to adapt your code at some point in the future, which will require more effort once it has grown in complexity.
+  Currently there is no agreed-upon plan or timeline for stabilizing flakes.
```
fricklerhandwerk, line 26:
```suggestion
  Currently there is no agreed-upon plan or timeline for stabilising flakes.
```

We have an unwritten rule to use BE and I should actually write it down.

fricklerhandwerk (changes requested):
While I think it's good to move the flakes question to conceptual FAQ, the change here seems a bit messy as it moves all of the items in that file. Since many troubleshooting and recipe questions need to be reworked anyway, in a different PR you could first pick any of them to open a new page, and then move the others accordingly.

source/recipes/troubleshooting.md
```diff
@@ -1,27 +1,28 @@
 # Troubleshooting
 
-## What to do if a binary cache is down or unreachable?
+## Binary cache down or unreachable?
```
fricklerhandwerk, line 3: I think having the full sentence was fine

fricklerhandwerk (commented):
Thanks for taking this on. I really recommend splitting up independent changes into separate PRs. The fixups to the troubleshooting page are fine, we can merge them quickly on their own. Many of the recipe questions need proper updates, and we that will stall the other parts that could move forward faster.

PR #737, opened 2023-10-04
Move faqs flakes

wamirez:
Picked out the question:

- [Should I enable flakes?](https://nix.dev/recipes/faq#should-i-enable-flakes)

and moved it from category FAQs under recipes (subsequently to be removed) into newly created category FAQs but under concepts instead, while adding some links and some minor adjustments.


fricklerhandwerk (approved):
Thanks!

PR #741, opened 2023-10-05
move installation instructions to top level

fricklerhandwerk:
@NixOS/documentation-team


infinisil (approved):
This could use a redirect in `_redirects`, otherwise looking good!

PR #745, opened 2023-10-06
Add EditorConfig file and fix errors

asymmetric:
To ensure basic standards.t

### DIFF
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -1,0 +1,20 @@
+name: "Check EditorConfig"
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: cachix/install-nix-action@v23
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - name: Checking EditorConfig
+      run: nix run nixpkgs#editorconfig-checker
+    - if: ${{ failure() }}
+      run: |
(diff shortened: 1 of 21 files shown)


fricklerhandwerk (approved):

.github/workflows/editorconfig.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: cachix/install-nix-action@v23
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - name: Checking EditorConfig
+      run: nix run nixpkgs#editorconfig-checker
+    - if: ${{ failure() }}
+      run: |
```
yukiisbored, line 18: I think it's better to create a comment on the PR if it fails (see https://github.com/NixOS/nix.dev/blob/master/.github/workflows/deploy.yml#L52) since it's more visible that way.

infinisil: I don't think that's necessary. As long as the check fails, CI will be red and users will read the causing error.

I think comments should only be used when it's hard to automatically verify something, so a reviewer needs to take a look at it manually.

infinisil (approved):

.github/workflows/editorconfig.yml
fricklerhandwerk: We could add a note to the contributor guides how to check for CI failures. They do tend to look scary to beginners. Let's merge this here though, it's a strict improvement.

PR #752, opened 2023-10-09
add introduction to cross compilation tutorial

fricklerhandwerk:
remove references to "towards reproducibility" as that's not very
focused, and at that point readers will likely be familiar with the
concept such that they only need the reference, if anything


proofconstruction (approved):
Good to avoid tarballs and use the release ref instead. Also to explain what we're doing in the tutorial.

Issue #750, opened 2023-10-09
Merge the pinning pages into one


fricklerhandwerk:
**Problem**

This is one topic that should be handled in one place. 

**Approaches**
Also add examples.

**Willing to help?**
As always.

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

olafklingt:
I started related work here https://github.com/NixOS/nix.dev/pull/303 i did not get very far but it might provide some input to whoever works on this.

PR #753, opened 2023-10-09
point to the pinning reference instead of the tutorial

fricklerhandwerk:


infinisil (changes requested):
Not sure about this. There's no link back from the reference on pinning to the pinning tutorial. I think either that should be added, or we should keep linking to the tutorial (which does link to the reference).

fricklerhandwerk:
We should merge these two anyway, since the "tutorial" doesn't really have much information: https://github.com/NixOS/nix.dev/issues/750

This PR is preparation for that.

infinisil:
Feels like this should just be a commit in a PR that does the pinning revising then

PR #754, opened 2023-10-09
move python dev env sample to recipes

fricklerhandwerk:


fricklerhandwerk:
@infinisil thanks for spotting the bug, fixed. We moved a lot of code today.

PR #755, opened 2023-10-09
move direnv instructions to recipes, update and rework text

fricklerhandwerk:


proofconstruction (approved):
With the latest changes to the shell.nix tutorials, this moves the direnv explainer out to a recipe.

wamirez:
How about creating a new category "Use Cases" under Recipes, where all new recipes get organized. Maybe that way the sidebar gets less bulky over time, increasing readability.

fricklerhandwerk:
Yeah, we definitely need some more structure there, as we will surely accumulate more things.

PR #757, opened 2023-10-09
 move question on impurities to FAQs section

wamirez:
Moved:

- [Are there some known impurities in builds?](https://nix.dev/recipes/faq#are-there-some-known-impurities-in-builds)

from recipes/faq.md to concepts/faq.md with some minor changes/additions.

File system (ext4) bulletpoint deleted because it seems to have ended up being a different issue. See: https://github.com/snabblab/snabblab-nixos/issues/42#issuecomment-220933393

### DIFF
--- a/source/concepts/faq.md
+++ b/source/concepts/faq.md
@@ -60,7 +60,7 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
 - The [Nix documentation team](https://nixos.org/community/teams/documentation.html) focuses on improving documentation and learning materials for stable features and common principles.
   When using flakes, you will have to rely more heavily on user-to-user support, third-party documentation, and the source code.
 
-## Are there any known impurities in builds?
+## Are there any impurities left in sandbox builds?
 
 Yes, there are:
 
--- a/source/concepts/faq.md
+++ b/source/concepts/faq.md
@@ -68,5 +68,4 @@ Yes, there are:
 - System's current time/date.
 - Linux kernel parameters, such as: [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
 - Timing behaviour of the build system—parallel Make build does not get the correct inputs in some cases.
-- Builds that access the home directory or fetch data from the Internet.
 - When random values are used, e.g., from `/dev/random` or `/dev/urandom`.
--- a/source/concepts/faq.md
+++ b/source/concepts/faq.md
@@ -69,4 +69,4 @@ Yes. There is:
 - The filesystem used for building (see also [`TMPDIR`](https://nixos.org/manual/nix/stable/command-ref/env-common.html#env-TMPDIR)).
 - Linux kernel parameters, such as: [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
 - Timing behaviour of the build system—parallel Make build does not get the correct inputs in some cases.
-- When random values are used, e.g., from `/dev/random` or `/dev/urandom`.
+- Insertion of random values, e.g., from `/dev/random` or `/dev/urandom`.
--- a/source/concepts/faq.md
+++ b/source/concepts/faq.md
@@ -60,12 +60,15 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
 - The [Nix documentation team](https://nixos.org/community/teams/documentation.html) focuses on improving documentation and learning materials for stable features and common principles.
   When using flakes, you will have to rely more heavily on user-to-user support, third-party documentation, and the source code.
 
-### Are there some known impurities in builds?
+## Are there any known impurities in builds?
 
-Yes.
+Yes, there are:
 
-- CPU (we try hard to avoid compiling native instructions, but rather hardcode supported ones)
-- current time/date
-- FileSystem (ext4 has a known bug creating [empty files on power loss](https://github.com/NixOS/nixpkgs/issues/15581))
-- Kernel
-- Timing behaviour of the build system (parallel Make not getting correct inputs in some cases)
+- CPU architecture—great effort being made to avoid compilation of native instructions in favour of hardcoded supported ones.
+- System's current time/date.
+- Linux kernel parameters e.g. [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
+- Timing behaviour of the build system—parallel Make build does not get the correct inputs in some cases.
+- Builds that access the home directory or fetch data from the Internet.
+- When randomness is added.
+
+See [The Nix Hour #12 \[purity at eval-, build- and runtime\]](https://www.youtube.com/watch?v=96jCLfKAa5M&t=913s) for more information and examples.
--- a/source/concepts/faq.md
+++ b/source/concepts/faq.md
@@ -62,10 +62,11 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
 
 ## Are there any impurities left in sandbox builds?
 
-Yes, there are:
+Yes. There is:
 
 - CPU architecture—great effort being made to avoid compilation of native instructions in favour of hardcoded supported ones.
 - System's current time/date.
+- The filesystem used for building (see also [`TMPDIR`](https://nixos.org/manual/nix/stable/command-ref/env-common.html#env-TMPDIR)).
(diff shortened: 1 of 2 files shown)


source/concepts/faq.md
```diff
@@ -59,3 +59,16 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
...
+
+## Are there any known impurities in builds?
+
+Yes, there are:
+
+- CPU architecture—great effort being made to avoid compilation of native instructions in favour of hardcoded supported ones.
+- System's current time/date.
+- Linux kernel parameters e.g. [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
+- Timing behaviour of the build system—parallel Make build does not get the correct inputs in some cases.
+- Builds that access the home directory or fetch data from the Internet.
```
infinisil, line 71: The home directory can't be accessed in the sandbox, so that shouldn't be listed here.

source/concepts/faq.md
```diff
@@ -59,3 +59,16 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
...
+- When randomness is added.
```
infinisil, line 72:
```suggestion
- When randomness is used, e.g. from `/dev/random` or `/dev/urandom`
```

source/recipes/faq.md
```diff
@@ -62,16 +62,6 @@ See <https://nixos.wiki/wiki/Nix_channels>
 
 See <https://github.com/nix-community/home-manager>
 
-### Are there some known impurities in builds?
-
-Yes.
-
-- CPU (we try hard to avoid compiling native instructions, but rather hardcode supported ones)
-- current time/date
-- FileSystem (ext4 has a known bug creating [empty files on power loss](https://github.com/NixOS/nixpkgs/issues/15581))
```
infinisil, line 71: This should also be in the updated list

source/concepts/faq.md
```diff
@@ -59,3 +59,16 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
...
+
+See [The Nix Hour #12 \[purity at eval-, build- and runtime\]](https://www.youtube.com/watch?v=96jCLfKAa5M&t=913s) for more information and examples.
```
infinisil, line 74: Happy to see it being useful! However I don't think it should be added here. Anything worthwhile that I mention during Nix hours should be added to the documentation, such that it's not necessary to link to them. Also, that way I can just link to the docs in future Nix hours :)

infinisil (changes requested):

source/concepts/faq.md
wamirez: I understand. However, since sandboxing is not enabled by default, wouldn't it be pertinent to add it to the list? How about:

```suggestion
- Builds with sandboxing disabled that access the home directory or fetch data from the Internet.
```

source/recipes/faq.md
wamirez: Digging further on the issue there's this:

> [domenkozar](https://github.com/domenkozar) commented [May 23, 2016](https://github.com/NixOS/nixpkgs/issues/15581#issuecomment-220964000)
> It turns out the issue was completely different impurity: https://github.com/snabblab/snabblab-nixos/issues/42#issuecomment-220933393
> 
> I'll open a trivial PR soon to address this (testing now if it works)

And:

> [domenkozar](https://github.com/domenkozar) commented [May 23, 2016](https://github.com/snabblab/snabblab-nixos/issues/42#issuecomment-220938218)
> Confirmed! It's due to https://github.com/snabblab/snabblab-nixos/blob/master/modules/sudo-in-builds.nix
> 
> If I uncomment that module, initrd builds successfully.

source/concepts/faq.md
infinisil: It is enabled by default on Linux, though disabled by default on Darwin. But once the sandbox is disabled you can do pretty much anything and it doesn't make sense to have a list for that.

I do think it's worthwhile to mention at the beginning that these are impurities left when the sandbox is enabled.

source/recipes/faq.md
infinisil: The filesystem can definitely be an impurity, see e.g. https://github.com/NixOS/nixpkgs/pull/104457#issuecomment-989106739

wamirez: @infinisil, do you think "File system dependency injections" would be a good way to generalize this case?

https://blog.ploeh.dk/2017/01/27/dependency-injection-is-passing-an-argument/
https://en.wikipedia.org/wiki/Dependency_injection

And thank you very much for your feedback (and patience). I lack the intuition/knowledge but am learning a great deal!

infinisil: No problem :)

I don't think dependency injection would be appropriate here. That term is mainly for testing when you want to change some underlying dependency with something that you can control in the test. But we're not changing any dependency here, it's just that the underlying file system differs from system to system.

infinisil: Here's a suggested wording:

```markdown
- The filesystem used for building (see also [`TMPDIR`](https://nixos.org/manual/nix/stable/command-ref/env-common.html?highlight=TMPDIR#env-TMPDIR)).
```

wamirez:
@infinisil, it seems good to merge now :)

source/concepts/faq.md
```diff
@@ -59,3 +59,14 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
   Improvements to experimental features have a low priority.
 - The [Nix documentation team](https://nixos.org/community/teams/documentation.html) focuses on improving documentation and learning materials for stable features and common principles.
   When using flakes, you will have to rely more heavily on user-to-user support, third-party documentation, and the source code.
+
+## Are there any impurities left in sandbox builds?
```
roberth, line 63:
```suggestion
## Are there any impurities left in sandboxed builds?
```

infinisil (approved):
Otherwise it looks good!

source/concepts/faq.md
```diff
@@ -59,3 +59,14 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
...
+
+Yes. There is:
+
+- CPU architecture—great effort being made to avoid compilation of native instructions in favour of hardcoded supported ones.
+- System's current time/date.
+- The filesystem used for building (see also [`TMPDIR`](https://nixos.org/manual/nix/stable/command-ref/env-common.html#env-TMPDIR)).
+- Linux kernel parameters, such as: [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
+- Timing behaviour of the build system—parallel Make build does not get the correct inputs in some cases.
+- Insertion of random values, e.g., from `/dev/random` or `/dev/urandom`.
```
roberth, line 72:
```suggestion
- Insertion of random values, e.g., from `/dev/random` or `/dev/urandom`.
- Differences between Nix versions. For instance, a new Nix version might introduce a new environment variable. A statement like `env > $out` is not promised by Nix to result in the same output, going into the future.
```

source/concepts/faq.md
```diff
@@ -59,3 +59,15 @@ There are downsides to relying on [experimental features](https://nixos.org/manu
...
   When using flakes, you will have to rely more heavily on user-to-user support, third-party documentation, and the source code.
+
+## Are there any impurities left in sandboxed builds?
+
+Yes. There is:
+
+- CPU architecture—great effort being made to avoid compilation of native instructions in favour of hardcoded supported ones.
+- System's current time/date.
+- The filesystem used for building (see also [`TMPDIR`](https://nixos.org/manual/nix/stable/command-ref/env-common.html#env-TMPDIR)).
+- Linux kernel parameters, such as: [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
```
roberth, line 70:
```suggestion
- Linux kernel parameters, such as:
  - [IPv6 capabilities](https://github.com/NixOS/nix/issues/5615).
  - binfmt interpreters, such as those configured with [`boot.binfmt.emulatedSystems`](https://search.nixos.org/options?show=boot.binfmt.emulatedSystems&query=binfmt)
```

roberth: Last suggestion from me. Let's merge after this.

PR #762, opened 2023-10-18
allow rec in packaging functions

wegank:
The best practice on `rec` is often interpreted in nixpkgs PRs as "rewriting `rec` using `let ... in` fixes the overriding logic", whereas it isn't the case. The second pitfall is therefore less justified.

In the [discussion about the same topic](https://discourse.nixos.org/t/avoid-rec-expresions-in-nixpkgs/8293) on Discourse, the most liked post also points out that avoiding `rec` in packaging functions doesn't make much sense.

Hence this PR.

### DIFF
--- a/source/recipes/best-practices.md
+++ b/source/recipes/best-practices.md
@@ -36,6 +36,10 @@ There are a couple of pitfalls:
 - It's possible to introduce a hard to debug error `infinite recursion` when shadowing a name, the simplest example being `let b = 1; a = rec { b = b; }; in a`.
 - Combining with overriding logic such as the [`overrideAttrs`](https://nixos.org/manual/nixpkgs/stable/#sec-pkg-overrideAttrs) function in {term}`Nixpkgs` has a surprising behaviour of not overriding every reference.
 
+:::{note}
+For packaging functions not supporting `finalAttrs: { ... }`, the use of `rec` is legitimate, see [this post for details](https://discourse.nixos.org/t/avoid-rec-expresions-in-nixpkgs/8293/4).
+:::
+
 :::{tip}
 Avoid `rec`. Use `let ... in`.
 


fricklerhandwerk:
The correct resolution here would be recommending the `final:` pattern, but that's not documented in the Nixpkgs manual yet and still only rarely used. @shamrocklee has a Nixpkgs PR open that I'd like to get merged once it's in shape, but we could put something here on nix.dev in the interim. @infinisil @proofconstruction what do you think?

proofconstruction:
What situations exist where the [`finalAttrs`](https://nixos.org/manual/nixpkgs/stable/#mkderivation-recursive-attributes) pattern cannot be used? (Other than aesthetically)

wegank:
As far as I know, the `finalAttrs` pattern is only supported by a few language helpers, such as `buildNimPackage`. For the most common ones, like `buildGoModule`, `buildPythonPackage`, or `rustPlatform.buildRustPackage`, the support isn't there.

fricklerhandwerk:
I'd prefer to recommend `finalAttrs:` directly where applicable, right there in the highlighted block. That would also be a good opportunity to place that link to the Nixpkgs manual so it's easier to find. All the details on using `rec` around overriding should go to the Nixpkgs manual. I want to keep nix.dev a thin guidance wrapper around reference documentation.

SuperSandro2000:
finalAttrs also has it pitfalls like FODs used in src where it changing rev breaks the FOD which is unnoticed because the original is probably in the cache.

fricklerhandwerk:
@NixOS/documentation-team decided to avoid the confusion by not mentioning the `overrideAttrs` case, which is indeed unrelated to the best practice here.

Superceded by https://github.com/NixOS/nix.dev/pull/790

Still, it needs better docs in Nixpkgs: https://github.com/NixOS/nixpkgs/issues/266485

PR #765, opened 2023-10-18
Mention rebuilds in best practices

zmitchell:
This is a small PR that adds a sentence about needless rebuilds to the "Reproducible source paths" section of the "Best Practices" page. This section currently mentions impurities, but there are actual practical concerns around using a fixed name, such as cache misses locally and in CI. Ask me how I know.

### DIFF
--- a/source/recipes/best-practices.md
+++ b/source/recipes/best-practices.md
@@ -216,6 +216,7 @@ The problem is that now your build is no longer reproducible, as it depends on t
 That cannot declared in the source code, and results in an impurity.
 
 If someone builds the project in a directory with a different name, they will get a different store path for `src` and everything that depends on it.
+This can be the cause of needless rebuilds both locally and in CI.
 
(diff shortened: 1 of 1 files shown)


source/recipes/best-practices.md
```diff
@@ -216,6 +216,7 @@ The problem is that now your build is no longer reproducible, as it depends on t
 That cannot declared in the source code, and results in an impurity.
 
 If someone builds the project in a directory with a different name, they will get a different store path for `src` and everything that depends on it.
+This can be the cause of needless rebuilds both locally and in CI.
```
fricklerhandwerk, line 219:
```suggestion
This can be the cause of needless rebuilds.
```

That's generally true, I see no need to specify it further.

fricklerhandwerk (approved):
Very good point. Incidentally I just ran into exactly this problem the other day.

PR #768, opened 2023-10-21
nixos-configuration-on-vm.md: Show how to run headless

### LINKED ISSUE #767: Running NixOS VMs headlessly
On current https://nix.dev/tutorials/nixos/nixos-configuration-on-vm if I run the suggested `./result/bin/run-nixos-vm`, on a headless machine (server) I get error:

```
gtk initialization failed
```

It makes some sense due to the following sentence being

> This command opens a window

but it would be very nice if this article could also be used on headless servers.

nh2:
Fixes #767.

### DIFF
--- a/source/tutorials/nixos/nixos-configuration-on-vm.md
+++ b/source/tutorials/nixos/nixos-configuration-on-vm.md
@@ -180,13 +180,23 @@ ls -R ./result
     run-nixos-vm
 ```
 
-Run the virtual machine:
+To run the virtual machine in a new window (requires GUI environment):
 
 ```shell-session
 ./result/bin/run-nixos-vm
 ```
 
-This command opens a window that shows the boot process of the virtual machine and ends at the `gdm` login screen where you can log in as `alice` with the password `testpw`.
+This will spawn a new QEMU window showing the boot process of the virtual machine and end at the `gdm` login screen where you can log in as `alice` with the password `testpw`.
+
+Spawning window requires running the command from a graphical environment.
+To run it in the current terminal instead to get a login via the serial console of the VM, use QEMU's `-nographic` flag (you can pass any QEMU flags to `run-nixos-vm`):
+
+```shell-session
+./result/bin/run-nixos-vm -nographic
+```
+
+Note that serial consoles in VMs do not automatically adapt to your terminal size, likely causing visual corruption when interacting with the prompt line.
+You can [resize](https://www.earth.li/~noodles/blog/2022/04/resizing-consoles-automatically.html) the serial console, or use configure an SSH server into your VM for easier non-GUI interaction.
 
 Running the virtual machine will create a `nixos.qcow2` file in the current directory.
 This disk image file contains the dynamic state of the virtual machine.


source/tutorials/nixos/nixos-configuration-on-vm.md
```diff
@@ -180,13 +180,19 @@ ls -R ./result
     run-nixos-vm
 ```
 
-Run the virtual machine:
+To run the virtual machine in a new window (requires GUI environment):
```
fricklerhandwerk, line 183:
```suggestion
Run the virtual machine in the current terminal:
```

Let's not show all the options here, that's what reference documentation is for. The purpose of this tutorial (as any tutorial here) is to serve as a first encounter, with pointers to all the details.

I prefer the terminal version as that is or at least appears more lightweight, and showing a the flag here should provide enough of a hint that omitting it may give you a GUI.

source/tutorials/nixos/nixos-configuration-on-vm.md
```diff
@@ -180,13 +180,19 @@ ls -R ./result
...
 ```
 
-This command opens a window that shows the boot process of the virtual machine and ends at the `gdm` login screen where you can log in as `alice` with the password `testpw`.
+To run it in the current terminal instead, pass `-nographic`:
+
+```shell-session
+./result/bin/run-nixos-vm -nographic
+```
+
+You will see the boot process of the virtual machine and end at the `gdm` login screen where you can log in as `alice` with the password `testpw`.
```
fricklerhandwerk, line 195: I strongly recommend against describing perceptions, as not everyone can even have them. Describe what the machine does instead.

```suggestion
This will show the boot process of the virtual machine and end at the `gdm` login screen where you can log in as `alice` with the password `testpw`.
```

I currently can't check it myself: will that indeed be a terminal prompt? Isn't GDM a graphical tool?

source/tutorials/nixos/nixos-configuration-on-vm.md
```diff
@@ -180,13 +180,19 @@ ls -R ./result
...
 
-Run the virtual machine:
+To run the virtual machine in a new window (requires GUI environment):
 
 ```shell-session
 ./result/bin/run-nixos-vm
 ```
 
-This command opens a window that shows the boot process of the virtual machine and ends at the `gdm` login screen where you can log in as `alice` with the password `testpw`.
+To run it in the current terminal instead, pass `-nographic`:
```
fricklerhandwerk, line 189: Generally, instead of presenting multiple options - which is distracting to first-time users because they'll be busy with following *one* set of instructions, and also doesn't scale well, since where do you draw the line? – we should link to the manual right afterward, such as:

```
Refer to the Nixpkgs manual for a [list of all commmand-line options to `run-nixos-vm`](<URL>).
```

(If that exists. Otherwise there's likely a `--help` flag?)

fricklerhandwerk (changes requested):
Thanks for pointing this out. I like the general direction of the change, but think it needs some streamlining.

nh2:
@fricklerhandwerk I've updated the PR.

Your comment was right that of course with the terminal, you can't see the configured GDM.

I think it's important to describe both the GUI and non-GUI options, because

* the article as-is just fails with `gtk initialization failed` when invoking `run-nixos-vm` from a non-GUI NixOS environment, which is extremely confusing (and the reason why I opened this PR, as the instructions didn't work out-of-the-box)
* the non-GUI option isn't perfect either, as you need a GUI to see the configured GDM, and there will likely be as many desktop users as headless users
  * further, the `-nographic` solution suffers from terminal corruption because QEMU serial consoles most of the time do not match the size of the terminal you launch them from; I've added a note about this now, so that users don't have the same "what is this rubbish, none of this works!"-feeling as I got initially.

fricklerhandwerk:
> Your comment was right that of course with the terminal, you can't see the configured GDM.

How is this addressed in the PR then?

> the article as-is just fails with gtk initialization failed when invoking run-nixos-vm from a non-GUI NixOS 

We could instead add this to the requirements at the beginning.

> the non-GUI option isn't perfect either

All these problems and the hassle required to work around them indicate that the additional information you collected should be put into the NixOS manual. For instance, that you can pass any QEMU flags, and how to handle terminal scrambling should definitely go into the manual.

I think all this is important to know, but not for the first encounter. Therefore, let's put it in the manual.

fricklerhandwerk:
Feel free to reopen if there is something particular you still want to address here.

PR #775, opened 2023-10-27
add Recipes section under Guides category

fricklerhandwerk:
this follows #701 and recent discussion around reshuffling the FAQ with @wamirez and @infinisil

PR #777, opened 2023-10-27
dissolve the Learning Journey section

fricklerhandwerk:
This does not yet follow #701, because the change to get there would be
too big. The goal is to keep the table of contents meaningful at all
times and the diff of each pull request manageable.

The packaging tutorial is quite important, so it's not wrong to have it
visible on the front page until we have enough material for a packaging
section.

The sharing dependencies article should really be a very brief guide, but
because that would be quite a big change, I moved it to the Guides
section for now.


fricklerhandwerk:
If https://github.com/NixOS/nix.dev/pull/775 gets merged first, I can also move the sharing deps thing into guides here to save ourselves one redirect step.

PR #778, opened 2023-10-27
reduce the sharing dependencies tutorial to a guide

fricklerhandwerk:
the contents do not really warrant a full-blown tutorial

### DIFF
--- a/source/guides/recipes/sharing-dependencies.md
+++ b/source/guides/recipes/sharing-dependencies.md
@@ -1,209 +1,112 @@
 (sharing-dependencies)=
 # Dependencies in the development shell
 
-<!-- Include any foreward you want here -->
+When [packaging software in `default.nix`](packaging-existing-software), you'll want a [development environment in `shell.nix`](declarative-reproducible-envs) to enter it conveniently with `nix-shell` or [automatically with `direnv`](./direnv).
 
-## Overview
+How to share the package's dependencies in `default.nix` with the development environment in `shell.nix`?
 
-### What will you learn?
+## Summary
 
-In this tutorial you'll learn how not to repeat yourself by sharing dependencies between `default.nix`, which is responsible for building the project, and `shell.nix`, which is responsible for providing you with an environment to work in.
-
-### How long will it take?
-
-This tutorial will take approximately 1 hour.
-
-### What will you need?
-
-This tutorial assumes you're familiar with Nixpkgs build helpers (`mkDerivation`, `buildPythonApplication`, etc) and know how to create environments for `nix-shell`.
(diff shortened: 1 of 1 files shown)


source/guides/recipes/sharing-dependencies.md
```diff
@@ -1,209 +1,112 @@
...
-The actual build recipe will be in `package.nix` and `default.nix` will import this file to perform the build.
-
-First create a `package.nix` file like this:
+Use the `inputsFrom` attribute to `pkgs.mkShell`:
 
 ```nix
+# default.nix
+let
+  pkgs = import <nixpkgs> {};
+  build = pkgs.callpackage ./build.nix {};
```
infinisil, line 16:
```suggestion
  build = pkgs.callPackage ./build.nix {};
```

source/guides/recipes/sharing-dependencies.md
```diff
@@ -1,209 +1,112 @@
...
-  version = "0.0.1";
-  format = "pyproject";
-  src = builtins.path { path = ./.; name = "shared_project_source"; };
-  propagatedBuildInputs = [
-    setuptools-scm
-    flask
-  ];
+  inherit build;
+  shell = pkgs.mkShell {
+    inputsfrom = [ build ];
```
infinisil, line 21:
```suggestion
    inputsFrom = [ build ];
```

source/guides/recipes/sharing-dependencies.md
```diff
@@ -1,209 +1,112 @@
...
-Try to build this project by running `nix-build -A build`
+## Complete example
 
-[nix_pills_callpackage]: https://nixos.org/guides/nix-pills/callpackage-design-pattern.html
+Assuming your build is defined in `build.nix`:
 
-## Adding development packages
+```nix
+# build.nix
+{ hello }: hello
```
infinisil, line 39: This feels a bit too minimal to get the point across. It should at least have some dependencies listed in e.g. `buildInputs` which one could then observe to now also have in the `nix-shell`

infinisil (approved):
Generally sounds good, @zmitchell what do you think? You wrote the original text.

source/guides/recipes/sharing-dependencies.md
fricklerhandwerk: We could have a nonsensical derivation where the builder returns true, but have hello as a build input... or something like that.

infinisil: How about a `cowsay nix > $out` in a `runCommand`, then you can have cowsay as the dependency :)

PR #780, opened 2023-10-27
add best practice on reproducible Nixpkgs configuration

fricklerhandwerk:
this is a pattern we're using repeatedly in the tutorials but never explained.


infinisil (approved):
Could also mention that `system` is a further impurity, though not as significant. Still, we can merge like this imo, though needs a rebase :)

Issue #772, opened 2023-10-24
Cloudflare build/deployment failures following flake-ification


yukiisbored:
**Observations**
Cloudflare builds has started failing post https://github.com/NixOS/nix.dev/commit/c3c97f8ee2b65e2b56959fe65b93bcdd9630001f.

See the following:
- https://github.com/NixOS/nix.dev/runs/17954685435
- https://github.com/NixOS/nix.dev/pull/771#issuecomment-1774777335

**Problem**

Cloudflare builds are failing for some reason. I cannot consult the build logs as it is behind Cloudflare.

**Approaches**

- Fix whatever causing Cloudflare build bots to fail
- Deploy to CloudFlare from GitHub actions instead similar to PR previews.
  - Allows GitHub contributors/maintainers to consult build logs and fix issues in the future without requiring Cloudflare admin intervention.
  - Proof of concept: 
    - https://github.com/YukiTestingGround/nix.dev/blob/master/.github/workflows/deploy.yml#L10-L51
 
**Willing to help?**

I'm able to move the deployment process to GitHub actions once the automated Cloudflare deployment is disabled.

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

yukiisbored:
Couple of examples of the site not updating. as of writing:

- https://github.com/NixOS/nix.dev/commit/833a73b7a3b10936746bb61ceabf4881ce722431
  - https://nix.dev/tutorials/nixos/nixos-configuration-on-vm#what-do-you-need still doesn't state "with a graphical environment"
- https://github.com/NixOS/nix.dev/commit/31fa807ccdd0b79cebea6f3f47187d7ec0544c6f
  - https://nix.dev/recipes/best-practices#reproducible-source-paths still doesn't mention rebuilds.

yukiisbored:
@NixOS/infra Could you please pull the build logs so we can fix the build failure? Thanks.

delroth:
@yukiisbored taking this over on the infra side of things. Sorry for the latency, the people who have time didn't have the access (now fixed), the people who have the access don't have time :)

[...]

PR #784, opened 2023-11-02
ci: switch to deploy to Netlify instead of Cloudflare pages (#772)

delroth:
NixOS owned web properties are ~all hosted on Netlify, and the feature set is very close to Cloudflare Pages. This PR migrates the repo's CI to auto-deploy pushed `master` branch commits to Netlify, as well as auto-deploy any pull request and commits to other branches to a preview domain.

This is roughly copied from the current https://search.nixos.org setup (https://github.com/nixos/nixos-search) with a few adjustments based on what was in this repo originally.


delroth:
cc @yukiisbored @fricklerhandwerk

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - uses: cachix/install-nix-action@v23
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
```
fricklerhandwerk, line 15: Is that needed at all, or is there something about Cachix that relies on channels? Ideally we'd have NIX_PATH empty or unset.

fricklerhandwerk (approved):
Looks good, as far as I can tell.

.github/workflows/build-and-deploy.yml
delroth: Not sure, I copied that from your existing config! https://github.com/NixOS/nix.dev/blob/master/.github/workflows/build.yml#L16

asymmetric: Let's keep this a verbatim copy then, and make improvementsin follow-up PR.s

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Build
+      run: |
+        nix-build
+
+        # The netlify action doesn't follow symlinks properly.
+        mkdir ./dist
+        cp -RL ./result/* ./dist/
+
+    - name: Deploy to Netlify
+      uses: nwtgck/actions-netlify@v2.0
```
asymmetric, line 30:
```suggestion
      uses: nwtgck/actions-netlify@v2.1.0
```

Note that it has a new option `enable-github-deployment`, which defaults to `true`. I think it's fine, but FYI

asymmetric: Also I noticed there's an [official action](https://github.com/netlify/actions/tree/master/build), but it's been a while since it was updated. Is the one above better?

asymmetric (commented):
Is it possible to actually do a test deployment of this, to ensure it works?

.github/workflows/build-and-deploy.yml
delroth: The official action just tells Netlify to perform a new build via their own build service (similar to Cloudflare). That wouldn't work here since you want to be able to build via Nix.

Will update to 2.1.0.

delroth:
> Is it possible to actually do a test deployment of this, to ensure it works?

nix.dev is already switched to Netlify and it was deployed via this GH workflow (on my fork repo) :)

fricklerhandwerk:
Downtime and PR friction won't get worse than it already is, so let's merge.

delroth:
Extra fixup in https://github.com/NixOS/nix.dev/pull/789 to fix pushes to master not triggering nix.dev prod updates.

PR #786, opened 2023-11-03
module-system.md: replace comments with captions

asymmetric:
NOTE: This removes line numbers, [this line](
https://github.com/nixos/nix.dev/tree/d6620ebca53a85c815b1c091b38c327fcecc7e48/source/conf.py#L60) stops having an effect.


fricklerhandwerk:
Oh yeah nice!

asymmetric:
Let's merge this. We have no line numbers anyway on all the files marked as diffs, at the moment, so the lack of them in the few "nix" files won't be noticed much.

PR #790, opened 2023-11-06
rephrase best practices for clarity

wamirez:
Added some changes to `source/guides/best-practices.md`.

- Removed a bullet point.
- Simplified `rec` example.

### DIFF
--- a/source/guides/best-practices.md
+++ b/source/guides/best-practices.md
@@ -31,10 +31,8 @@ rec {
 { a = 1; b = 3; }
 ```
 
-There are a couple of pitfalls:
-
-- It's possible to introduce a hard to debug error `infinite recursion` when shadowing a name, the simplest example being `let b = 1; a = rec { b = b; }; in a`.
-- Combining with overriding logic such as the [`overrideAttrs`](https://nixos.org/manual/nixpkgs/stable/#sec-pkg-overrideAttrs) function in {term}`Nixpkgs` has a surprising behaviour of not overriding every reference.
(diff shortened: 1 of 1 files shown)


proofconstruction (approved):
Reviewed in documentation team meeting. This simplifies the section.

source/guides/best-practices.md
```diff
@@ -31,10 +31,8 @@ rec {
 { a = 1; b = 3; }
 ```
 
-There are a couple of pitfalls:
-
-- It's possible to introduce a hard to debug error `infinite recursion` when shadowing a name, the simplest example being `let b = 1; a = rec { b = b; }; in a`.
-- Combining with overriding logic such as the [`overrideAttrs`](https://nixos.org/manual/nixpkgs/stable/#sec-pkg-overrideAttrs) function in {term}`Nixpkgs` has a surprising behaviour of not overriding every reference.
+A common pitfall is to introduce a hard to debug error `infinite recursion` when shadowing a name.
+The simplest example for this is: `let a = 1; in rec { a = a; }`.
```
fricklerhandwerk, line 35:
```suggestion
The simplest example for this is: `let a = 1; in rec { a = a; }`
```

With a colon in between we probably don't need the full stop, as that tends to interfere with reading the expression.

Alternatively, but that could give it too much emphasis right before the tip:

````suggestion
The simplest example for this is:

```{code-block} nix
let a = 1; in rec { a = a; }
```
````

We may want to add `<details>` because it may not be apparent what's trivially weird about it:

> One might expect that the value assigned in the attribute set is taken from the surrounding `let` binding, but names from the `rec` scope take precedence.

PR #798, opened 2023-11-12
Add prose lint checks with Vale

asymmetric:
Tons of errors, including false positives (e.g. "Eelco"). We might want to:

- Disable the spell-checker
- Add lots of exceptions
- Tell it to ignore capitalized words (but it still will complain about e.g. "npm")

For now, we use the default profile plus a rule for title capitalization.

If we wanted, we could import bigger rulesets, like the ones from [Red Hat](https://redhat-documentation.github.io/supplementary-style-guide/), [Microsoft](https://learn.microsoft.com/en-us/style-guide/welcome/) or [Google](https://developers.google.com/style/).


yukiisbored:
I personally prefer if we leave the spellchecker enabled and maintain a vocabulary for Vale, I don't think it adds too much work.

However, for style guidelines, it might be worth to simply choose an existing one instead of starting from scratch. We could always modify them if needed in the future.

asymmetric:
@yukiisbored any idea how to compare the different packages? Or how would you choose among them?

asymmetric:
I decided to reduce the scope of the PR to only the uncontroversial bit, so it can be merged faster.

We can discuss styleguides and spellcheck in a follow-up PR.

WDYT?

infinisil (approved):
More automation is great!

PR #799, opened 2023-11-12
Update nixos-configuration-on-vm.md to fix the stateVersion

joej164:
Was running through this tutorial using NixOS I downloaded and installed last week.  Figured I'd submit a PR to update the tutorial with the current version.

### DIFF
--- a/source/tutorials/nixos/nixos-configuration-on-vm.md
+++ b/source/tutorials/nixos/nixos-configuration-on-vm.md
@@ -46,6 +46,10 @@ In the working directory you will then find two files:
 
 The default NixOS configuration without comments is:
 
+:::{admonition} NixOS
+If you used the Plasma image instead of the Gnome image, your config may look a bit different.
+:::
+
 ```nix
 { config, pkgs, ... }:
 {
@@ -100,6 +104,10 @@ Therefore you will remove the reference to `hardware-configuration.nix`:
 
 The complete `configuration.nix` file now looks like this:
 
+:::{admonition} NixOS
+If you used the Plasma image instead of the Gnome image, your config may look a bit different.
+:::
+
 ```nix
 { config, pkgs, ... }:
 {
@@ -121,7 +129,7 @@ The complete `configuration.nix` file now looks like this:
     initialPassword = "testpw";
   };
 
-  system.stateVersion = "22.11";
+  system.stateVersion = "23.05";
 }
 ```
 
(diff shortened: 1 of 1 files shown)


olafklingt:
thank you for the pr. the direction is definitely correct. but maybe the admonition about the used nixos version needs to be more general, because there is also the minimal image (i have to check which variants matter). at the same time the comment should be kept short.

joej164:
Per your comment, I improved the admonishments.  Happy to try and reduce the message again.

Ideally if I had more time I'd figure out how the version could be a build parameter of the docs and just be updated when new versions are released.

fricklerhandwerk:
For the sake of keeping things working rather than always up-to-date or complete, I'd consider only adding things if they're tested exactly as described. Examples that need tinkering to are not good.

What would help here is pinning the installation image rather than pointing to whatever happens to be the most recent one. We could also add a note that updating the whole thing to the latest version is a welcome contribution. There will be people now and again who go through the motions and it would be easy for them to make a PR.

I wouldn't want to litter everything with notes amounting to "you may be out of luck". It's NixOS after all, it can and therefore should Just Work.

olafklingt:
@fricklerhandwerk 
> What would help here is pinning the installation image 
We are talking here about the nixos-generate-config on the users computer we cannot pin that. At least not in a meaningful way.

olafklingt (approved):

fricklerhandwerk:
Sure we can:
```console
$ nix-shell -p nixos-install-tools -I nixpkgs=channel:nixos-21.05
```

olafklingt:
@joej164 hey your comment pointed us to reformulate the tutorial a bit to so that it is exactly reproducible. So i wrote this PR #804 . I think will make another PR to update it to 23.11 after its release. Feel free to comment here or elsewhere when you think that it can be improved, or that your suggested commend is still needed.

joej164:
I looked it over, looks good.  Glad I was able to help / point out an area of improvement.

Last question....one thing that seemed missing was a tutorial on using NixOS..the configuration.nix file to just update the OS itself.  Everything was focused on Docker container, or using NixOS to build VM's.

Did I miss something?  Should I put in an issue to the above topic?

Anyways, thanks for reviewing the issue.

olafklingt:
Yes you are right a nixOS configuration tutorial is not written jet. 
There are some open issues open that suggest NixOS tutorials:
https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+nixos+configuration
I think what you are missing is already covered in those issues.

joej164:
thanks for the link.  Should I just close this issue then?

Issue #796, opened 2023-11-11
Copyright date incorrect


boseiju:
Hey,

The following string is currently displayed in the footer on every page:

> © Copyright 2016-1980, NixOS Foundation.

The source looks like it should provide an accurate date:

https://github.com/NixOS/nix.dev/blob/ea8357b14943a67749344c319eedf8bf8f6d3c35/source/conf.py#L18

https://github.com/NixOS/nix.dev/blob/ea8357b14943a67749344c319eedf8bf8f6d3c35/source/conf.py#L82

Is this perhaps an infrastructure issue causing the date to be set inaccurately?

yukiisbored:
@boseiju I believe this is one of the things the Nix build environment does to keep things pure. I think we should just hardcode it and have someone update the copyright year as an annual task.

rjpcasalino:
see: https://github.com/sphinx-doc/sphinx/issues/3451
I've opened a PR https://github.com/NixOS/nix.dev/pull/800

rjpcasalino:
We should be able to close this now with #800 being merged.

PR #800, opened 2023-11-14
add SOURCE_DATE_EPOCH to Makefile to fix copyright year

rjpcasalino:
addresses: https://github.com/NixOS/nix.dev/issues/796
fix comes from this comment https://github.com/sphinx-doc/sphinx/issues/3451#issuecomment-373708110

git log time format is used in comment and I ran locally with that and the copyright prints correctly...but it'll break if not building from git so use `date +%s` instead to set `SOURCE_DATE_EPOCH`; `date +%Y` also should work but wasn't sure if it needed unix time or not so just went with that to be safe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ SPHINXOPTS    = -W
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
+SOURCE_DATE_EPOCH = $(shell git log -1 --format=%ct)
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@ SPHINXOPTS    = -W
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
-SOURCE_DATE_EPOCH = $(shell git log -1 --format=%ct)
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):

infinisil (changes requested):
This breaks when not building from git, e.g.

```
❯ nix-build https://github.com/rjpcasalino/nix.dev/tarball/dce01693d0de34719c21bfb4bcd6492309b4aa4c -A build
this derivation will be built:
  /nix/store/vk5x9xgbjfk2xdpkv55i2jh2y893gk96-nix-dev.drv
building '/nix/store/vk5x9xgbjfk2xdpkv55i2jh2y893gk96-nix-dev.drv'...
unpacking sources
unpacking source archive /nix/store/nsryax35vwnz2gk1jwy03n8w3q22kn0m-q5551s40hmlm6p04y0agvkasc39i4zpc-source
source root is q5551s40hmlm6p04y0agvkasc39i4zpc-source
patching sources
configuring
no configure script, doing nothing
building
sphinx-build -b html -d build/doctrees  -W source build/html
make: git: No such file or directory
Running Sphinx v5.3.0

Exception occurred:
  File "/nix/store/3yjxwrzxisvd6kd6w0mhk3a38xlwhk1p-python3.10-sphinx-5.3.0/lib/python3.10/site-packages/sphinx/builders/gettext.py", line 175, in <module>
    timestamp = float(source_date_epoch)
ValueError: could not convert string to float: ''
The full traceback has been saved in /build/sphinx-err-w_ox0iog.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:54: html] Error 2
```

Makefile
```diff
@@ -13,6 +13,8 @@ PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) source
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) source
+# Explicitly set SOURCE_DATE_EPOCH for Sphinx copyright
+SOURCE_DATE_EPOCH = $(shell date +%s)
```
infinisil, line 17:
```suggestion
SOURCE_DATE_EPOCH = $(shell date +%s --date='Jan 1')
```

This will make it use the unix epoch of 1 January of the current year, so it's at least reproducible within the same year :)

infinisil (changes requested):

infinisil (approved):

PR #802, opened 2023-11-19
File sets tutorial

infinisil:
This is a tutorial for the [`lib.fileset`](https://nixos.org/manual/nixpkgs/unstable/#sec-fileset) library I've been working on recently.

Initially this also included a tutorial on builtin features for adding files to the store, but this has now been moved to [a gist](https://gist.github.com/infinisil/e44999f0e5e08cea90c65bbedb20d157), since it's not really best practice.

This work is sponsored by [Antithesis](https://antithesis.com/) :sparkles:


source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,241 @@
...
+
+These builtin features are very limited in functionality and are not recommended for anything non-trivial. For more advanced use cases, the file set library should be used instead.
+
+In this tutorial you'll learn both how to use the builtins and the file set library.
+
+## Setting up a local experiment
+
+To experiment with source file selection, we'll set up a local project.
+
+To start out, create a new directory, enter it, and set up `niv` to manage the Nixpkgs dependency:
```
bzm3r, line 23: Can we use flakes instead of `niv`? My reasons behind this request:

0) There is a non-trivial cost to the proliferation of certain "third party" tools: https://ayats.org/blog/no-flake-utils/

1) It is already hard to navigate the Nix ecosystem (in terms of learning) for a newcomer, due to how long it has taken flakes to just be *the* way to do things. Now that community consensus has been built, official tutorials should focus on using flakes rather than alternatives to flakes? If we would like to still continue to avoid using flakes, then we ought to use a Nix-only solution?

3) It will be helpful to have more tutorials showing flake usage in various scenarios, without involving `flake-utils`, `flake-compat`, etc. (as most other blog tutorials tend to do).

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,241 @@
...
+```nix
+{ runCommand }:
+runCommand "hello" { } ''
+  echo hello world
+''
+```
+
+And try it out:
+```shell-session
+$ nix-build
```
bzm3r, line 65: Should we be using `nix build` here instead?

bzm3r (commented):
Thank you for this! I love reading new tutorials. I have a couple of suggestions, but I'm not really sure what the `nix.dev` policy is around flakes/"experimental commands", so I might be off-base.

source/tutorials/source-file-selection.md
infinisil: Some time ago we made the decision to generally avoid documenting experimental features in official documentation. The documentation team is very constrained on resources, and we don't want to spend this on documenting features that have the potential to break at any time. Especially when there's also an abundance of stable features that need to be documented.

As the new CLI and Flakes [get stabilised](https://discourse.nixos.org/t/stabilising-the-new-nix-command-line-interface/35531), the documentation team will definitely be on-board to update the documentation to use it though. So, I'd like to redirect any complaints about Flakes not being documented to the Nix team, which notably also doesn't have a lot of resources, they need help with stabilisation! :smile: 

Note that I am writing this tutorial fairly generically though, using a `package.nix` file that works regardless of Flakes, and the `default.nix` file can also be imported from Flakes.

It's unfortunate that there's no stable _official_ tool for locking dependencies, but I think `niv` is fine because it only relies on stable features and is generally considered stable itself.

Nevertheless, I will have to add some more notes on Flakes into this tutorial, but not in a good way: Due to Flakes always importing the local directory into the Nix store, it's a dangerous (secret files can easily get added to the store) and wasteful (many store imports for the same directory even if barely anything changed). :sweat_smile:

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,645 @@
...
+|-- package.nix
+`-- string.txt
+```
+
+However there are some subtle problems with this approach:
+- Note how the name of the store path ends with `-select`.
+  So the name of the local directory influenced the result.
+
+  This means that whenever you rename the project directory
+  or a collegue runs it in a different directory name,
```
henrik-ch, line 224:
```suggestion
  or a colleague runs it in a different directory name,
```

fricklerhandwerk (commented):
Given this essentially does a survey of inferior alternatives in the beginning, this should be an explanation, or as we use to call these: a developer story. It's been a while that we added an actual one, but this one is a perfect example. And it would become even clearer if you removed "recommendations" (using passive voice on those is dangerous terrain anyway) and focused on pains and solutions and went away from instructions and replaced them by descriptions.

The goal of a developer story, in my opinion, would be to explain a technique and convince the reader that it's the right tool for the problem under discussion. Most importantly it should not be advertisement, official documentation is not a place for that.

A tutorial on file set combinators can be a lot shorter, and skip directly to instructions how to use it for a particular problem. An explanation and a tutorial are not mutually exclusive.

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,645 @@
...
+      # Don't include .nix files
+      || ! lib.hasSuffix ".nix" pathString;
+  };
+in
+stdenv.mkDerivation {
+  name = "builtins-path";
+  inherit src;
+  nativeBuildInputs = [ tree ];
+  postInstall = ''
+    tree
```
henrik-ch, line 265: I am not sure why the `touch created-in-the-build` was in the previous example, but omitted in this one - maybe it's not important.

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,645 @@
...
+### `builtins.path`
+
+<!-- TODO: Use lib.cleanSourceWith instead and only briefly mention builtins.path? -->
+
+The above problems can be fixed by using [`builtins.path`](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-path) instead.
+It allows customising the name of the resulting store path with its `name` argument.
+And it allows selecting the files that should be included with its `filter` argument.
+
+```nix
+{ runCommand, tree, lib }:
```
henrik-ch, line 246:
```suggestion
{ stdenv, runCommand, tree, lib }:
```

infinisil:
@fricklerhandwerk I'm not really sure what you're concretely suggesting. I think the document is useful as is. I wrote it with the goal of teaching the reader as many things to the user as possible, and I think it does a good job at that. This is the document I'd want people to read if they need to select local source files for builds. I don't really care what it's called.

I guess we can at least avoid `builtins.path`, but other than that I think everything should be included. This also means including path-to-string coercion, since it's _fine_ to use that in many cases, some people don't need to use file sets, indeed most users won't use it.

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,645 @@
...
+```
+
+We could also add `touch $out` to make the build succeed,
+but we'll omit that for the sake of the tutorial, since we only need the build logs.
+This also makes it easier to build it again, since successful derivation builds would get cached.
+From now on we'll also make build outputs a bit shorter for the sake of brevity.
+
+## Builtins
+
+### Coercion of paths to strings
```
infinisil, line 85: @fricklerhandwerk I guess this section so far is specific to files, so maybe this should be reflected in the header.

Importing files into the store is explicitly not supported by `lib.fileset`, so this (or maybe even with `builtins.path` to set the store path name) is still the recommended way to do this.

henrik-ch (commented):
@infinisil my review is partial and done on a version before your push https://github.com/NixOS/nix.dev/commit/6582681bd36644bb6cd57d6d2d478408f74b2e5f

I will let you produce and review again tomorrow.

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,645 @@
...
+As you can see, the `string.txt` file was hashed and added to the store,
+which then allowed the build to access it.
+
+The underlying functionality is the same as `nix-store --add` on an absolute path:
+```shell-session
+$ nix-store --add $(pwd)/string.txt
+/nix/store/j5lwpnlfrngks3bpidfr5hcrhgq0fy78-string.txt
+```
+
+### Coercion of directory paths to strings
```
infinisil, line 158: @fricklerhandwerk This one is specific to directories, indeed this one should be replaced with `lib.fileset`. I'm okay removing this section, though ideally it should be _somewhere_, because it's very common to see this in the wild, people should understand how this mechanism works, even if they shouldn't use it themselves. And unless we deprecate this in Nix itself, people won't stop using it.

source/tutorials/source-file-selection.md
```diff
@@ -0,0 +1,645 @@
...
+
+  This means that:
+  - Even if your build only needs a few files,
+    changing _any_ file in the directory requires rebuilding the derivation,
+    potentially wasting a lot of time.
+
+  - If you have any secrets in the current directory,
+    they get imported into the Nix store too, exposing them to all users on the system!
+
+### `builtins.path`
```
infinisil, line 237: This section should be removed, agreed, it's not good practice to use `builtins.path` (or `lib.cleanSourceWith`). In the future we can at least potentially deprecate `cleanSourceWith`.

fricklerhandwerk:
I strongly disagree we should stuff as many things as possible into any given article, but rather focus on each article's purpose. Learners' capacity to absorb information is limited, and they will usually choose a particular article based on an objective they're following - the only thing to guide them here is the title and possibly a summary or overview. Saying this is a tutorial (or framing it as such) and then blasting them with the information firehose is counterproductive.

You can choose the purpose and the format:
- all the ways to add sources to a build (reference)
- the default/recommended/best/... way to do source file selection (tutorial)
- why source file selection scales best with file set combinators (explanation)
- how to efficiently select source files for a build (guide)

They will all have *some* overlap in the factual information they contain, but will be structured and presented very differently. Trying all of that at once is also inefficient for you, because if the objective isn't clear, you'll just run into conflicting priorities and annoying reviews. :)

I'd also wager that a "just learn the everything" approach doesn't scale because it's simply impractical for most people resource-wise, it takes too much time. The contributor community already has a selection bias because of that, arguably because surviving with Nix is near impossible otherwise -- the other sub-optimal option being cargo-culting your way through.

My suggestion is to pick an objective and follow through with that. Currently you're closest to the explanation, but a tutorial may be better because the text reads like the other options are to be largely displaced by your library, so why teach them to beginners all?

All that said, of course the current form is a clear improvement over not having anything. Therefore I won't block here. But why stop at 70% if we can have 90-100%? (Yes, I'm convinced the "non-functional" aspect of highly focused, frugal communication is worth 30% of the impact.)

source/tutorials/source-file-selection.md
infinisil: Actually this one could be removed as well. It's a related topic, but single files are generally not considered a "source".

With all of these sections removed, indeed this becomes very specific to file sets.

infinisil:
To not lose it, I moved all sections I'm planning to remove here: https://gist.github.com/infinisil/e44999f0e5e08cea90c65bbedb20d157

infinisil:
I'm fairly happy with how it looks now!

Qyriad:
I think the section on `src = ./.` and its caveats is worth keeping some of; a prospective user is *highly* likely to run into that pattern, and otherwise won't know the drawbacks of simply copying it.

infinisil:
@Qyriad Agreed, though I feel like it should be somewhere else (Nix reference docs? A separate tutorial/explanation/gotchas on path-related builtin functionality?).

Qyriad:
I think it's fine to mention it really briefly here, and have a more complete explanation elsewhere (maybe something like "You may commonly see path literals assigned directly to environment variables, like `stdenv.mkDerivation { src = ./. }`, however this has significant downsides. See \[link] for more info")

The Nix reference docs generally should explain major caveats, I think, so a mention there is worthwhile. I'm not sure that should be the More Complete explanation linked to here or a separate nix.dev page like you suggested. If it's a separate page though then that definitely shouldn't be blocking for this PR.

infinisil:
@Qyriad Sounds good, just pushed https://github.com/NixOS/nix.dev/pull/802/commits/436b4aa1b9323439cc4cd28a913d0ca4b96da6f7, how does that look?

Qyriad:
Minor suggestion, otherwise LGTM!

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,555 @@
...
+Using these features directly can be tricky however:
+
+- Coercion of paths to strings, such as the wide-spread pattern of `src = ./.`,
+  makes the derivation dependent on the name of the current directory.
+  It also doesn't allow being more precise about which files to use.
+
+- The [`builtins.path`](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-path) function
+  (and equivalently [`lib.sources.cleanSourceWith`](https://nixos.org/manual/nixpkgs/stable/#function-library-lib.sources.cleanSourceWith))
+  can address these problems.
+  However, it's hard to get the desired path selection using function-based interface.
```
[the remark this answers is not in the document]
infinisil: Good point! I'd word it a bit more lightly though. While it is technically old, it's not deprecated. How about

```suggestion
  can address these problems.
  However, it's hard to get the desired path selection using the `filter` function interface.
```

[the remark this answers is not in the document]
Qyriad: Seems good to me!

infinisil:
Note that in https://github.com/NixOS/nix.dev/pull/806, @fricklerhandwerk made a bunch of suggestions to this PR which we briefly discussed and merged into this one

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,694 @@
...
+
+```shell-session
+$ rm result
+$ nix-build
+error: lib.fileset.difference: Second argument (negative set)
+  (/home/user/select/result) is a path that does not exist.
+  To create a file set from a path that may not exist, use `lib.fileset.maybeMissing`.
+```
+
+Follow the instructions in the error message, and use `maybeMissing` to create a file set from a path that may not exist (in which case the file set will be empty):
```
infinisil, line 378: Note that this is not yet merged: https://github.com/NixOS/nixpkgs/pull/265964

So to make this work for now, you'll have to use that branch instead:
```
nix-shell -p niv --run "niv update nixpkgs --owner tweag --branch fileset.optional"
```

We shouldn't merge this PR until at least that PR is merged.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+Then create a `default.nix` file with the following contents:
+
+```{code-block} nix
+:caption: default.nix
+{
+  system ? builtins.currentSystem,
+  inputs ? import ./nix/sources.nix,
+}:
+let
+  pkgs = import inputs.nixpkgs {
```
fricklerhandwerk, line 118:
```suggestion
  sources ? import ./nix/sources.nix,
}:
let
  pkgs = import sources.nixpkgs {
```

As discussed in #806 this was a fluke on my end.

source/tutorials/file-sets.md
roberth: Unblocked

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+and set up `niv` to pin the Nixpkgs dependency:
+
+```shell-session
+$ mkdir fileset
+$ cd fileset
+$ nix-shell -p niv --run "niv init --nixpkgs nixos/nixpkgs --nixpkgs-branch nixos-unstable"
+```
+
+:::{note}
+We're using the `nixos-unstable` channel branch here, since no stable release has all the features needed for this tutorial.
```
roberth, line 106:
```suggestion
We're using the `nixos-unstable` channel branch here, because the library is part of NixOS 23.11 and up, which has not been released as of writing.
```

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+    fileset = sourceFiles;
+  };
+  postInstall = ''
+    mkdir $out
+    cp -v hello.txt $out
+  '';
+}
+```
+
+The call to `fs.trace` will print the file set that will be used as a derivation input.
```
roberth, line 231:
```suggestion
The call to `fs.trace` prints the file set that will be used as a derivation input. We'll remove it later.
```

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+trace: /home/user/fileset
+trace: - build.nix (regular)
+trace: - default.nix (regular)
+trace: - hello.txt (regular)
+trace: - nix (all files in directory)
+trace: - world.txt (regular)
+/nix/store/vhyhk6ij39gjapqavz1j1x3zbiy3qc1a-fileset
+```
+
+## Union (explicitly exclude files)
```
roberth, line 416:
```suggestion
## Union for explicit sets
```

Inclusion or exclusion is a matter of how you use it, but union is great for defining sets by explicit enumeration.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+trace: - nix (all files in directory)
+trace: - build.nix (regular)
+trace: - string.txt (regular)
+this derivation will be built:
+  /nix/store/zmgpqlpfz2jq0w9rdacsnpx8ni4n77cn-filesets.drv
+...
+/nix/store/6pffjljjy3c7kla60nljk3fad4q4kkzn-filesets
+```
+
+One way to fix this is to use [`unions`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.unions).
```
roberth, line 442:
```
One way to fix this is with [`unions`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.unions).
```

but wait, we also need difference?
Maybe start with `unions` without difference? That's actually a great way to make sure you're not including too much. Nix's own flake does that.

Then later flip it around: if you prefer to include any new or previously unknown files by default, you can use `unions` in a `difference` instead.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
++      (fs.unions [
++        (fs.maybeMissing ./result)
++        ./default.nix
++        ./build.nix
++        ./nix
++      ]);
+ in
+```
+
+Changing any of the excluded files now doesn't necessarily require a rebuild anymore.
```
roberth, line 464:
```suggestion
Changing any of the excluded files now doesn't necessarily cause a new build anymore.
```

Nix never rebuilds, unless you ask for it.

roberth: I might have missed another instance of this inaccurate term, because I know I'm also guilty of it.

source/tutorials/file-sets.md
roberth: Oh I get it now. You also have a section for explicit inclusion.
Probably still best to start with that.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+$ nix-build
+trace: /home/user/fileset
+trace: - hello.txt (regular)
+trace: - world.txt (regular)
+/nix/store/ckn40y7hgqphhbhyrq64h9r6rvdh973r-fileset
+```
+
+## Filter
+
+Dealing with a large number of files, independent of their location, can be done programmatically with the [`fileFilter`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.fileFilter) function.
```
roberth, line 482:
```suggestion
With the [`fileFilter`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.fileFilter) function, you can find sets of files by criteria.
```

Actually close to set builder notation, for the mathematically inclined.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+         (fs.maybeMissing ./result)
+-        ./default.nix
+-        ./build.nix
++        (fs.fileFilter (file: lib.hasSuffix ".nix" file.name) ./.)
+         ./nix
+       ]);
+ in
+```
+
+This does not change the result:
```
roberth, line 504:
```suggestion
This does not change the result, even if we add a new `.nix` file.
```

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+'./src/select.c' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.c'
+'./src/select.h' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.h'
+...
+/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset
+```
+
+Only the specified files are used, even when a new one is added:
+
+```shell-session
+$ touch src/select.o
```
roberth, line 600:
```suggestion
$ touch src/select.o src/README.md
```

Add a README for those who know they already filter by git.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+trace:   - select.c (regular)
+trace:   - select.h (regular)
+trace: - world.txt (regular)
+/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset
+```
+
+## Conclusion
+
+We have shown some examples on how to use all of the fundamental file set functions.
+For more complex behavior, they can be composed as needed.
```
roberth, line 723:
```suggestion
For more complex use cases, they can be composed as needed.
```

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
+(file-sets)=
+# Working with local files
+<!-- TODO: Switch all mentions of unstable to stable once 23.11 is out -->
+
+To build a local project in a Nix derivation, its source files must be accessible to the [`builder` executable](https://nixos.org/manual/nix/stable/language/derivations#attr-builder).
```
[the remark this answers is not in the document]
fricklerhandwerk: I disagree. I would have to think hard what "hermetic" is even supposed to mean. An executable being able to read files is something very concrete and down-to-earth in Unix land.

source/tutorials/file-sets.md
infinisil: I found it hard to make the tutorial flow that way, but I can take another look

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+Even though file sets conceptually contain local files, these files are *never* added to the Nix store unless explicitly requested.
+You don't have to worry about accidentally copying secrets into the world-readable store.
+
+In this example, although we pretty-printed the home directory, no files were copied.
+This is also evident from the expression evaluating instantly.
+
+:::{note}
+This is in contrast to coercion of paths to strings such as in `"${./.}"`,
+which copies the whole directory to the Nix store on evaluation!
+:::
```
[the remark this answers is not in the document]
fricklerhandwerk: Agreed, let's stay focus and leave room for an "all the ways to handle files" reference.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+With [experimental Flakes](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake), a local directory containing `flake.nix` is always copied into the Nix store *completely* unless it is a Git repository!
+:::
+
+This implicit coercion also works for files:
+
+```shell-session
+nix-repl> fs.trace /etc/nix/nix.conf
+trace: /etc/nix
+trace: - nix.conf (symlink)
+```
```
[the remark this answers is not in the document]
fricklerhandwerk: I suspect the motivation here was, before having any setup, how to show something the library does? nix.conf won't be a symlink on a vanilla Ubuntu install, but that's a minor issue.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,725 @@
...
+
+In addition to the included file, this also prints its [file type](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-readFileType).
+
+If a given path doesn't exist, the library will complain:
+
+```shell-session
+nix-repl> fs.trace /etc/nix/nix.nix
+error: lib.fileset.trace: Argument (/etc/nix/nix.nix)
+  is a path that does not exist.
+```
```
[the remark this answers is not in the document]
fricklerhandwerk: It's fair to make users aware of this, since there's a good chance they will introduce typos. It's not wrong to have it but I won't complain if it was removed.

source/tutorials/file-sets.md
fricklerhandwerk:
```suggestion
We're using the `nixos-unstable` channel branch here, because the library is part of NixOS 23.11 and up, which has not been released at the of writing.
```

source/tutorials/file-sets.md
fricklerhandwerk: We won't in this tutorial?

source/tutorials/file-sets.md
fricklerhandwerk: That would require rewriting it quite a lot, I suggest making that a follow-up, if at all.

source/tutorials/file-sets.md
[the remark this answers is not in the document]
infinisil: Indeed, `/etc/nix/nix.conf` was the only file I could think of that every user with Nix should have

source/tutorials/file-sets.md
infinisil: Yeah, I'll change the `print` though.

source/tutorials/file-sets.md
infinisil: Yeah, follow up sounds better!

source/tutorials/file-sets.md
infinisil: See https://github.com/NixOS/nix.dev/pull/802#discussion_r1402328626

source/tutorials/file-sets.md
infinisil: Changing this to "find sets of files satisfying a criteria"

source/tutorials/file-sets.md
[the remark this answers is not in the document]
infinisil: I don't think it hurts to include this though, it's just one sentence but it still teaches something related. I now combined it with the above paragraph though, that sounds good.

source/tutorials/file-sets.md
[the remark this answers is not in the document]
infinisil: I now changed this to just creating a file beforehand: [(diff)](https://github.com/NixOS/nix.dev/pull/802/commits/67b688adb6addc00a0a7af482add9460ea6acae5)

source/tutorials/file-sets.md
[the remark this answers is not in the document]
infinisil: Hmm I agree with @fricklerhandwerk here, there's exactly one mention of "hermetic" in the Nix manual, and it doesn't tell you what it means.

Though I also agree with @roberth that it's a bit too deep, `builder` is irrevelant to pretty much everybody using Nix.

How about this:
```suggestion
To build a local project in a Nix derivation,
its source files must be [accessible](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-sandbox)
to the [derivation builder](https://nixos.org/manual/nix/stable/language/derivations#attr-builder).
```

infinisil:
@fricklerhandwerk @roberth In a full read-through I made some adjustments in https://github.com/NixOS/nix.dev/pull/802/commits/69e6dd6d9d3aed8c6cd64a9aa094bd9b7eab9fb8.

And now i also removed the initial part using `src = ./.`. I don't think that added much, it's not good practice. And `src = ./hello.txt` failing was only incidental, it's no reason for wanting the file set library.

source/tutorials/file-sets.md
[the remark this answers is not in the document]
fricklerhandwerk: I insist that "`builder` executable" is unambiguous as opposed to everything else. "derivation builder" is jargon that may as well refer to `mkDerivation` (the rename to "build helper" is very recent) or a remote machine.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+Also use https://nixos.org/manual/nixpkgs/unstable/#sec-functions-library-fileset
+instead of https://nixos.org/manual/nixpkgs/unstable/#sec-fileset
+-->
+
+To build a local project in a Nix derivation,
+its source files must be accessible
+to the [derivation builder](https://nixos.org/manual/nix/stable/language/derivations#attr-builder).
+Since by default, derivations get built in an [isolated environment](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-sandbox)
+that only allows reading from the Nix store,
+the Nix language has built-in features to copy local files to the store and expose the resulting store paths.
```
fricklerhandwerk, line 14:
```suggestion
To build a local project in a Nix derivation, source files must be accessible to its [`builder` executable](https://nixos.org/manual/nix/stable/language/derivations#attr-builder).
Since by default, the `builder` runs in an [isolated environment](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-sandbox) that only allows reading from the Nix store, the Nix language has built-in features to copy local files to the store and expose the resulting store paths.
```

I think it's not wrong to keep these simple mechanisms transparent. I won't fight though, and leave it as an optional suggestion.

source/tutorials/file-sets.md
```diff
@@ -23,23 +24,27 @@ Using these features directly can be tricky however:
   can address these problems.
   However, it's often hard to express the desired path selection using the `filter` function interface.
 
-In this tutorial you'll learn how to use the Nixpkgs [`fileset` library](https://nixos.org/manual/nixpkgs/unstable/#sec-fileset) to work with local files in derivations.
-It abstracts over built-in functionality and offers a safer and more convenient interface.
+In this tutorial you'll learn how to use Nixpkgs' [`lib.fileset` library](https://nixos.org/manual/nixpkgs/unstable/#sec-fileset) to work with local files in derivations.
+It abstracts over built-in functionality with a safer and more convenient interface.
```
fricklerhandwerk, line 28: I'd revert both sentences. Apostrophising Nixpkgs looks really weird, and `abstracting over X with Y`, while grammatically correct can be misleading to be read as "built in functionality having a safer interface".

source/tutorials/file-sets.md
```diff
@@ -23,23 +24,27 @@ Using these features directly can be tricky however:
...
 
 ## File sets
 
-The file set library is based on the concept of _file sets_,
+The basic concept is that of a _file set_,
```
fricklerhandwerk, line 32: Revert too (or if you want file set as singular, say "based on the concept of a file set". "The basic concept is" - of what? What's the subject? And starting with "the basic concept" even if you say "of the file set library" is probably not even important, and the original sentence said what we want: that the file set revolves around this basic concept.

source/tutorials/file-sets.md
```diff
@@ -48,7 +53,7 @@ trace: /home/user (all files in directory)
 null
 ```
 
-All functions that expect a file set for an argument also accept a [path](https://nixos.org/manual/nix/stable/language/values#type-path).
+All functions that expect a file set for an argument also accept a [path](https://nixos.org/manual/nix/stable/language/values#type-path) instead.
```
fricklerhandwerk, line 56: Revert, because "instead" can be misinterpreted as "mutually exclusive".

source/tutorials/file-sets.md
```diff
@@ -64,14 +69,14 @@ trace: /home/user (all files in directory)
 :::
 
 Even though file sets conceptually contain local files, these files are *never* added to the Nix store unless explicitly requested.
-You don't have to worry as much about accidentally copying secrets into the world-readable store.
+So You don't have to worry as much about accidentally copying secrets into the world-readable store.
```
fricklerhandwerk, line 72: Capitalisation. Not sure if the connective is really needed here. I would (and did) skip it.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+
+Even though file sets conceptually contain local files, these files are *never* added to the Nix store unless explicitly requested.
+So You don't have to worry as much about accidentally copying secrets into the world-readable store.
+
+In this example, although we pretty-printed the home directory, no files were copied.
+This is in contrast to coercion of paths to strings such as in `"${./.}"`,
+which copies the whole directory to the Nix store on evaluation!
+
+:::{warning}
+When using [experimental Flakes](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake), local directories with a `flake.nix` are always copied into the Nix store *completely* unless it is a Git repository!
```
fricklerhandwerk, line 79: What are "experimental Flakes"? Are there non-experimental flakes?

I still think that should be "With the `flakes` experimental feature enabled, ..."

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+
+```shell-session
+$ echo hello > hello.txt
+$ echo world > world.txt
+```
+
+## Adding files to the Nix store
+
+To add files in a given file set to the Nix store,
+we use [`toSource`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.toSource).
```
fricklerhandwerk, line 137: Who is "we" here? It makes sense to say "we" when we're referring to this group writing the tutorial, e.g. when explaining why we're telling people to fetch from `master` rather than `nixpkgs-unstable`. But here here we'd run the danger of saying "we use this, you do you", while we really want to either state a fact "X does Y" or tell people what to do.
```suggestion
Files in a given file set can be added to the Nix store with [`toSource`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.toSource).
```

I don't super like the phrasing, but it should be good enough.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+This function needs a `root` attribute to know which path to use as the root of the resulting store path,
+but only exactly the files in the `fileset` argument are included in the result.
```
fricklerhandwerk, line 139: "root of store path" is potentially confusing. Isn't that `/nix/store`?

There is no "but" here,  because one part doesn't contradict the other.

```suggestion
The argument to this function requires a `root` attribute to determine which source directory to copy to the store, and exactly the files in the `fileset` attribute are included in the result.
```

PS: I really don't like breaking sentences up into multiple lines, because edits are usually based on sentence, not phrase. Don't know about others, but dealing with multiple lines in the GitHub UI almost always make it more tedious for me.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+  postInstall = ''
+    mkdir $out
+    cp -v hello.txt $out
+  '';
+}
+```
+
+The call to `fs.trace` prints the file set that will be used as a derivation input.
+
+The build will succeed (note that it will take some time to fetch the first time around):
```
fricklerhandwerk, line 168: This should take some time already when doing the first `trace`, so either note it there or skip.

```suggestion
Try building it:
```

source/tutorials/file-sets.md
```diff
@@ -312,7 +316,7 @@ result -> /nix/store/xknflcvjaa8dj6a6vkg629zmcrgz10rh-fileset
 Since `src` refers to the whole directory, and its contents change when `nix-build` succeeds, Nix will have to start over every time.
 
 :::{note}
-This will also happen without the file set library, e.g. when setting `src = ./.;` directly.
+This will also happen with `src = ./.`, without using the file set library.
```
fricklerhandwerk, line 251: This sentence structure is potentially confusing. What you probably mean here is "... i.e., without using the file set library", but that makes it a lot more clunky. I'd revert that too.

fricklerhandwerk (commented):
A few kinks found their way in among a lot of good corrections. Ready to merge from my side after resolving the icky ones.

Great stuff, looking forward to getting this out there.

source/tutorials/file-sets.md
infinisil: In the docs team meeting right now, changing it to

> A _file set_ is a data type representing a collection of local files.
File sets can be created, composed, and manipulated with the various functions of the library.

source/tutorials/file-sets.md
infinisil: Changing it to

> All functions that expect a file set for an argument can also accept a [path](https://nixos.org/manual/nix/stable/language/values#type-path).

source/tutorials/file-sets.md
infinisil: Changing to

> Because of that you don't have to worry as much about accidentally copying secrets into the world-readable store.

source/tutorials/file-sets.md
infinisil: > When using the [`flakes` and `nix-command` experimental features](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake),
a local directory within a Flake is always copied into the Nix store *completely* unless it is a Git repository!

In the meeting

source/tutorials/file-sets.md
infinisil: Used
> The argument to this function requires a `root` attribute to determine which source directory to copy to the store.
Only the files in the `fileset` attribute are included in the result.

source/tutorials/file-sets.md
infinisil: Happens here due to niv not being able to re-use the download.

Changed it to an admonition though (see commit)

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+trace: /home/user/fileset
+trace: - hello.txt (regular)
+trace: - world.txt (regular)
+/nix/store/ckn40y7hgqphhbhyrq64h9r6rvdh973r-fileset
+```
+
+## Filter
+
+With the [`fileFilter`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.fileFilter) function,
+you can find sets of files satisfying a criteria.
```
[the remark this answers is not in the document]
infinisil: > The [`fileFilter`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.fileset.fileFilter) function allows filtering file sets such that each included file satisfies the given criteria.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+Create some additional files:
+
+```shell-session
+$ mkdir src
+$ touch build.sh src/select.{c,h}
+```
+
+Then create a file set from only the files to be included explicitly:
+
+```{code-block} diff
```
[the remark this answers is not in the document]
infinisil: Changed it back to a partial file, no diff.

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+'./src/select.c' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.c'
+'./src/select.h' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.h'
+...
+/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset
+```
+
+Only the specified files are used, even when a new one is added:
+
+```shell-session
+$ touch src/select.o README.md
```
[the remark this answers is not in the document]
infinisil: discussed in meeting, it's fine to have it

source/tutorials/file-sets.md
```diff
@@ -0,0 +1,661 @@
...
+$ nix build path:.
+```
+
+However it's also not needed, because by default, `nix build` only allows access to files tracked by Git.
+:::
+
+## Intersection
+
+This is where `intersection` comes in.
+It allows creating a file set that consists only of files that are in _both_ of two given file sets.
```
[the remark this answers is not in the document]
infinisil: Discussed, it's fine without :)

infinisil:
Reviewed together in the docs team meeting today with @fricklerhandwerk, @proofconstruction, @henrik-ch, @DanielSidhion and @wamirez, let's merge this!

PR #806, opened 2023-11-22
review pass on tutorial sections

fricklerhandwerk:
it's a huge diff, but it doesn't change that much:
- start with a `mkDerivation` scaffold that uses `src = ./.`
  - this saves the detour with an intentionally broken derivation and
    the explanation it requires
  - it provides an opportunity to drop another bit of information about
    `toSource`, namely that it always creates a directory in the store
- use two source files to start with, otherwise the "try one file, then
  the whole directory" approach is not well-motivated
  - this obviates spending words on justifying a contrived example,
    it flows more naturally
- mostly use diffs to highlight the changes
  - this is not a strong proposal
  - the explicit-include-union example starts over, so the build-up is
    manageable
  - it works out quite nicely overall: since there is sufficient repetition
    where we reasonably expect learners to have gotten the point,
    the last two sections can be reduced just to the relevant bit.
- rename section headers to match functions, not domain problems
  - one exception is the split of the two union approaches, which should
    be okay
- reword all instructions to be imperative

the review made clear how powerful and elegant the library is, and that the
tutorial is utterly well-structured and ordered in a way that feels natural
and covers all important aspects!

great job @infinisil!

### DIFF
--- a/source/tutorials/file-sets.md
+++ b/source/tutorials/file-sets.md
@@ -69,7 +69,7 @@ which copies the whole directory to the Nix store on evaluation!
 :::
 
 :::{warning}
-With the [`flakes` experimental feature](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake) enabled, a local directory containing `flake.nix` is always copied into the Nix store *completely* unless it is a Git repository!
+With [experimental Flakes](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake), a local directory containing `flake.nix` is always copied into the Nix store *completely* unless it is a Git repository!
 :::
 
 This implicit coercion also works for files:
@@ -514,7 +514,7 @@ trace: - world.txt (regular)
 Notably, the approach of using `difference ./.` explicitly selects the files to _exclude_, which means that new files added to the source directory are included by default.
 Depending on your project, this might be a better fit than the alternative in the next section.
 
-## Union (include)
+## Union (explicitly include files)
 
 To contrast the previous approach, `unions` can also be used to select only the files to _include_.
 This means that new files added to the current directory would be ignored by default.
@@ -556,7 +556,7 @@ stdenv.mkDerivation {
     fileset = sourceFiles;
   };
   postInstall = ''
-    cp -vr $src $out
+    cp -vr . $out
   '';
 }
 ```
@@ -573,20 +573,13 @@ trace: - world.txt (regular)
 this derivation will be built:
   /nix/store/sjzkn07d6a4qfp60p6dc64pzvmmdafff-fileset.drv
 ...
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source' -> '/nix/store/zl4n1g6is4cmsqf02
-dci5b2h5zd0ia4r-fileset'
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/build.sh' -> '/nix/store/zl4n1g6i
-s4cmsqf02dci5b2h5zd0ia4r-fileset/build.sh'
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/hello.txt' -> '/nix/store/zl4n1g6
-is4cmsqf02dci5b2h5zd0ia4r-fileset/hello.txt'
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/world.txt' -> '/nix/store/zl4n1g6
-is4cmsqf02dci5b2h5zd0ia4r-fileset/world.txt'
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/src' -> '/nix/store/zl4n1g6is4cms
-qf02dci5b2h5zd0ia4r-fileset/src'
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/src/select.c' -> '/nix/store/zl4n
-1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.c'
-'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/src/select.h' -> '/nix/store/zl4n
-1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.h'
+'.' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset'
+'./build.sh' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/build.sh'
+'./hello.txt' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/hello.txt'
+'./world.txt' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/world.txt'
+'./src' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src'
+'./src/select.c' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.c'
+'./src/select.h' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.h'
 ...
 /nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset
 ```
@@ -650,7 +643,7 @@ this derivation will be built:
 This includes too much though, as not all of these files are needed to build the derivation as originally intended.
 
 :::{note}
-With the [`flakes` experimental feature](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake) enabled, it's [not really possible](https://github.com/NixOS/nix/issues/9292) to use this function, even with
+With [experimental Flakes](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake), it's [not really possible](https://github.com/NixOS/nix/issues/9292) to use this function, even with
 
 ```shell-session
 nix build path:.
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Nah, please someone tell that Vale thing to shut up. I'm really done for today.

source/tutorials/file-sets.md
```diff
@@ -307,126 +375,168 @@ error: lib.fileset.difference: Second argument (negative set)
...
-/nix/store/clrd19vn5cv6n7x7hzajq1fv43qig7cp-filesets
+trace: /home/user/fileset
+trace: - hello.txt (regular)
+trace: - world.txt (regular)
+/nix/store/ckn40y7hgqphhbhyrq64h9r6rvdh973r-fileset
 ```
 
-Notable with this approach is that new files added to the current directory are _included_ by default.
+Notably, with this approach, new files added to the source directory are _included_ by default.
 Depending on your project, this might be a better fit than the alternative in the next section.
```
infinisil, line 515:
```suggestion
Notably, with this approach of using `difference ./.`, we're explicitly selecting the files to _exclude_, which means that new files added to the source directory are _included_ by default.
Depending on your project, this might be a better fit than the alternative in the next section.
```

source/tutorials/file-sets.md
```diff
@@ -307,126 +375,168 @@ error: lib.fileset.difference: Second argument (negative set)
...
 
-## Only including necessary files
+## Union (include)
```
infinisil, line 517:
```suggestion
## Union (for including only select files)
```

To avoid confusion with the previous paragraph mentioning "included by default"

source/tutorials/file-sets.md
```diff
@@ -307,126 +375,168 @@ error: lib.fileset.difference: Second argument (negative set)
...
 trace: - nix (all files in directory)
-trace: - package.nix (regular)
-trace: - string.txt (regular)
-/nix/store/bzvhlr9h2zwqi7rr9i1j193z9hkskhmk-filesets
+trace: - world.txt (regular)
+/nix/store/vhyhk6ij39gjapqavz1j1x3zbiy3qc1a-fileset
 ```
 
-## Adding file sets together
+## Union (exclude)
```
infinisil, line 416:
```suggestion
## Union (for excluding only select files)
```

Though this also makes it sound like that's what this function does, which would be wrong :thinking:

source/tutorials/file-sets.md
```diff
@@ -435,49 +545,72 @@ And then create a file set from just the ones we're interested in:
...
+fs.trace sourceFiles
+
+stdenv.mkDerivation {
+  name = "fileset";
+  src = fs.toSource {
+    root = ./.;
+    fileset = sourceFiles;
+  };
+  postInstall = ''
+    cp -vr $src $out
```
infinisil, line 559:
```suggestion
    cp -vr . $out
```

The source will be in the local directory

source/tutorials/file-sets.md
```diff
@@ -498,58 +631,71 @@ Building we get
...
+  /nix/store/p9aw3fl5xcjbgg9yagykywvskzgrmk5y-fileset.drv
+...
+/nix/store/cw4bza1r27iimzrdbfl4yn5xr36d6k5l-fileset
 ```
 
-This includes too much though, we don't need all of these files to build the derivation.
+This includes too much though, as not all of these files are needed to build the derivation as originally intended.
+
+:::{note}
+With the [`flakes` experimental feature](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake) enabled, it's [not really possible](https://github.com/NixOS/nix/issues/9292) to use this function, even with
```
infinisil, line 653:
```suggestion
With [experimental Flakes](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake), it's [not really possible](https://github.com/NixOS/nix/issues/9292) to use this function, even with
```

infinisil (changes requested):
Awesome, thanks a lot! Here are some minor counter-suggestions :P

source/tutorials/file-sets.md
```diff
@@ -86,6 +57,21 @@ trace: /home/user (all files in directory)
...
+In this example, although we pretty-printed the home directory, no files were copied.
+This is also evident from the expression evaluating instantly.
+
+:::{note}
+This is in contrast to coercion of paths to strings such as in `"${./.}"`,
+which copies the whole directory to the Nix store on evaluation!
+:::
+
+:::{warning}
+With the [`flakes` experimental feature](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake) enabled, a local directory containing `flake.nix` is always copied into the Nix store *completely* unless it is a Git repository!
```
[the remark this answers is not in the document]
fricklerhandwerk: Oof. But what are "experimental Flakes"?

Unrelated, I think `flakes` should transitively enable `nix-command`, @Ericson2314 wdyt?

source/tutorials/file-sets.md
```diff
@@ -435,49 +545,72 @@ And then create a file set from just the ones we're interested in:
...
+'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/hello.txt' -> '/nix/store/zl4n1g6
+is4cmsqf02dci5b2h5zd0ia4r-fileset/hello.txt'
+'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/world.txt' -> '/nix/store/zl4n1g6
+is4cmsqf02dci5b2h5zd0ia4r-fileset/world.txt'
+'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/src' -> '/nix/store/zl4n1g6is4cms
+qf02dci5b2h5zd0ia4r-fileset/src'
+'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/src/select.c' -> '/nix/store/zl4n
+1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.c'
+'/nix/store/6k6pv78b1dkdhbdlzxar5vb3z3c8fwza-source/src/select.h' -> '/nix/store/zl4n
+1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.h'
```
infinisil, line 589:
```suggestion
'.' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset'
'./build.sh' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/build.sh'
'./hello.txt' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/hello.txt'
'./world.txt' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/world.txt'
'./src' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src'
'./src/select.c' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.c'
'./src/select.h' -> '/nix/store/zl4n1g6is4cmsqf02dci5b2h5zd0ia4r-fileset/src/select.h'
```

With the `cp` change

source/tutorials/file-sets.md
```diff
@@ -94,210 +80,292 @@ trace: /etc/nix
...
 
-Then create a `default.nix` file:
+Then create a `default.nix` file with the following contents:
 
 ```{code-block} nix
 :caption: default.nix
 {
   system ? builtins.currentSystem,
-  sources ? import ./nix/sources.nix,
+  inputs ? import ./nix/sources.nix,
```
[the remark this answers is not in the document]
fricklerhandwerk: Ah, oh. Apparently I refrained from actually committing to that, but it stuck after spending a bit of time thinking it through. Now I remember: I didn't pull through because it would diverge from `niv` upstream docs and likely confuse people. Sorry for the detour.

(I still think it should consider converging with flake conventions, `inputs` or `sources`, either way. `sources` has some appeal because this is what the references are about: remote sources of... stuff. `inputs` is so generic to the point it's meaningless; and they aren't even necessarily used as build inputs.)

source/tutorials/file-sets.md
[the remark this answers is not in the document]
infinisil: Alternative wordings welcome, I think "experimental Flakes" with the link should be clear enough for now though. I'd rather not detangle Flakes in this tutorial :P

source/tutorials/file-sets.md
fricklerhandwerk:
```suggestion
Notably, the approach of using `difference ./.` explicitly selects the files to _exclude_, which means that new files added to the source directory are included by default.
Depending on your project, this might be a better fit than the alternative in the next section.
```

Removing the ominous "we" and the emphasis on "including", which was a bit irritating already because the section is "Union (exclude)" - now that's resolved, yay!

source/tutorials/file-sets.md
fricklerhandwerk: That would be resolved with the other suggestion.

source/tutorials/file-sets.md
fricklerhandwerk:
```suggestion
## Union (explicitly exclude files)
```

?

source/tutorials/file-sets.md
fricklerhandwerk: Yeah but that's another layer of indirection. With `$src` you "only" have to know that non-reserved attrs become env vars to fully grasp the whole thing. Now you'd also have to know that other bit.

source/tutorials/file-sets.md
infinisil: I like it!

source/tutorials/file-sets.md
fricklerhandwerk: Eh. Okay, that's actually a good argument to do that.

source/tutorials/file-sets.md
infinisil:
```suggestion
## Union (explicitly include files)
```

Let's do this though, I already committed the `explicitly exclude files` now

fricklerhandwerk:
Suggestion stand-off. Feel free to merge as you like, it's all yours after all.

source/tutorials/file-sets.md
infinisil: We're already using that fact throughout the entire tutorial though, starting with the first `cp -v hello.txt $out`. It should be consistent, but I also don't want those long `/nix/store/... ->` lines in the log output.

infinisil:
Nice, thanks! Let's merge this one into https://github.com/NixOS/nix.dev/pull/802 now :)

PR #807, opened 2023-11-23
vm improve reproducible command and 23.11

olafklingt:


henrik-ch (approved):
Tested and looks good to me.

source/tutorials/nixos/nixos-configuration-on-vm.md
```diff
@@ -48,7 +46,7 @@ EOF
 Or as a one-liner:
 
 ```shell-session
-nix-shell -I nixpkgs=channel:nixos-22.05 -p "let pkgs = import <nixpkgs> { config = {}; overlays = []; }; nixos = pkgs.nixos { services.xserver.enable = true; services.xserver.desktopManager.gnome.enable = true; }; in nixos.config.system.build.nixos-generate-config"
+nix-shell -I nixpkgs=channel:nixos-23.11 -p "((import <nixpkgs> { config = {}; overlays = []; }).nixos (pkgs.path + /nixos/modules/installer/cd-dvd/installation-cd-graphical-gnome.nix)).config.system.build.nixos-generate-config"
```
fricklerhandwerk, line 49: I'd prefer if it were the exact same code as above just with newlines collapsed, otherwise less proficient readers may get stuck with analyzing the differences. There's no reason for us not to make this exactly identical; it's not there for being readable but for being a single line to paste.

I won't block on this. Feel free to self merge.

fricklerhandwerk (approved):

source/tutorials/nixos/nixos-configuration-on-vm.md
olafklingt: I am considering now to move the multi line version into a footnote with a comment describing whats happening. so the tutorial stays compact.

fricklerhandwerk: I think it's fine that way, because the tutorial is supposed to teach you things and the bulky version is for clarity. If you really want to stash it away, I suggest you rather put it into `<details>` so it's right there where you're reading so there's no need to jump around.

olafklingt: yes i agree :-) as i thought the same.

PR #811, opened 2023-11-27
fix and simplify devmode script

fricklerhandwerk:
previously, stderr was garbled, which made it really annoying to deal
with build errors.

it turns out that the library wrapper around calling a process is not
just counterproductive, but also completely unnecessary.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,6 @@
-{
-  inputs ? import ./nix/sources.nix,
-  system ? builtins.currentSystem,
+{ inputs ? import ./nix/sources.nix
+, system ? builtins.currentSystem
+,
 }:
 let
   pkgs = import inputs.nixpkgs {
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -37,19 +37,32 @@ let
...
+        from livereload import Server
+        from subprocess import Popen, PIPE
+        import shlex
 
         server = Server()
 
-        build_docs = shell("make html")
+        def build():
+          p = Popen(
+            shlex.split("nix-build -A build"),
```
infinisil, line 48: You're always doing a `nix-build` now, so it can't reuse any previously built artifacts. What's the difference in responsiveness on incremental changes?

fricklerhandwerk: Yes, but it turns out not to matter. On my machine it's fast enough to be unnoticeable. Takes about as long as switching between editor and browser at a relaxed pace. At worst, you'll see the page contents updating.

infinisil: Running it myself, a `make html` runs in less than a second, whereas a `nix-build -A build` runs in 10 seconds..

fricklerhandwerk: The number of times I had to restart the shell or `make clean` because something didn't get updated warrant the order of magnitude increase, for me at least. An out-of-tree `make html` would of course be better but also more work. This is correct but slow, now.

infinisil: > `make clean` because something didn't get updated

Then that's a `Makefile` issue we should fix. I really don't think we should use Nix for incremental development, it's not made for that

infinisil: Discussed in the meeting, 10 seconds is a bit long but still fine, and the sphinx problem is annoying. If it gets much longer in the future this needs to be reconsidered.

PR #815, opened 2023-11-28
Add `wiki-js` recipe

drupol:
Hi,

This is my first PR in this project.

Let me know if:

1. Such thing is the right place
2. If it worth adding it here
3. If it worth it, should I change anything?

### DIFF
--- a/source/guides/recipes/wiki-js.md
+++ b/source/guides/recipes/wiki-js.md
@@ -3,7 +3,7 @@
 
 [Wiki.js](https://js.wiki/) is a modern, open-source wiki engine running on Node.js and written in JavaScript. It offers a sleek and user-friendly platform for managing and sharing knowledge.
 
-To quickly set up an instance of Wiki.js on your local NixOS machine, a database service is required, and optionnaly a reverse proxy for HTTPS URLs.
+To quickly set up an instance of Wiki.js on your local NixOS machine, a database service is required, and optionally a reverse proxy for HTTPS URLs.
 
 Below is a streamlined and ready-to-use recipe to spawn this service on your server:
 
--- a/source/guides/recipes/wiki-js.md
+++ b/source/guides/recipes/wiki-js.md
@@ -1,0 +1,52 @@
+(setup-wiki-js)=
+# Setup wiki.js
+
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
@Drupol thanks for the PR! I think generally it's a great idea to present such example setups. Yours is particularly well-written in my opinion.

Your proposal prompts the @NixOS/documentation-team to finally think in all seriousness about how to deal with those more generally. We've briefly discussed it before, but came to no conclusion because there were more pressing issues.

Here are some considerations:
- Here on nix.dev such samples will eventually go stale, it would in principle be better to keep them next to the code (same problem as with Wikis in general)
- There are no conventions or infrastructure to accommodate this in Nixpkgs. The closest seems to be `pkgs/by-name`, but there is currently no pattern to handle modules.
  - I can imagine your text to be in the `README.md` of the package's directory, and it would be a good example to build upon. It would mean though, that for a while no one will be able to find it because there are no hints and facilities for this.
- We can merge it as is, just to have a place for it now, but it will just postpone systematically dealing with a question I think is important already in the mid-term

infinisil:
I generally think that NixOS modules should ideally Just Work™ with `enable = true`. Could the wiki-js module be made to work that way too?

drupol:
@fricklerhandwerk I was wondering where to post this recipe because this might help some other people. TBH, I first though at https://nixos.wiki but since it's not official, I would prefer not to post it there and centralize the documentation somewhere else. Hence my PR here. Let me know where you would like to see it and I'll do the appropriate changes.

@infinisil That's is a great idea, I'll see if I can work on that.

source/guides/recipes/wiki-js.md
```diff
@@ -0,0 +1,52 @@
+(setup-wiki-js)=
+# Setup wiki.js
+
+[Wiki.js](https://js.wiki/) is a modern, open-source wiki engine running on Node.js and written in JavaScript. It offers a sleek and user-friendly platform for managing and sharing knowledge.
+
+To quickly set up an instance of Wiki.js on your local NixOS machine, a database service is required, and optionnaly a reverse proxy for HTTPS URLs.
```
henrik-ch, line 6:
```suggestion
To quickly set up an instance of Wiki.js on your local NixOS machine, a database service is required, and optionally a reverse proxy for HTTPS URLs.
```

henrik-ch (approved):
LGTM, only small nitpick.

PR #820, opened 2023-11-29
tutorials: Map Module 

brianmcgillion:
Finish the tutorial, and update some snippets to be consistent with the previously made ones.

Fix some geocode references.

Full implementation according to the tutorial, [maps-module](https://github.com/brianmcgillion/maps-module), for reference.

### DIFF
--- a/source/tutorials/module-system/module-system.md
+++ b/source/tutorials/module-system/module-system.md
@@ -742,20 +742,23 @@ In the `paramForMarker` function:
 
 ```{code-block} diff
 :caption: marker.nix
-       paramForMarker = marker:
-         let
-           attributes =
--            [
-+            lib.optional
-+              (marker.style.label != null)
-+              "label:${marker.style.label}"
+     requestParams = let
++      paramForMarker = marker:
++        let
++          attributes =
++            lib.optional (marker.style.label != null)
++            "label:${marker.style.label}"
 +            ++ [
-               "$(geocode ${
-                 lib.escapeShellArg marker.location
-               })"
++              "$(${config.scripts.geocode}/bin/geocode ${
++                lib.escapeShellArg marker.location
++              })"
++            ];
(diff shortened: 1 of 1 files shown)


source/tutorials/module-system/module-system.md
```diff
@@ -742,20 +742,23 @@ In the `paramForMarker` function:
...
++                lib.escapeShellArg marker.location
++              })"
++            ];
++        in "markers=\"${lib.concatStringsSep "|" attributes}\"";
++      in
++        builtins.map paramForMarker config.map.markers;
 ```
 
-Here, the label for each `marker` is only propagated to the CLI parameters if `marker.style.label` is set.
+Notice here how we now create a unique `marker` for each user by concatenating the `label` and `location` attributes together, and assigning them to the `requestParams`. The label for each `marker` is only propagated to the CLI parameters if `marker.style.label` is set.
```
fricklerhandwerk, line 761:
```suggestion
Note how we now create a unique `marker` for each user by concatenating the `label` and `location` attributes together, and assigning them to the `requestParams`.
The label for each `marker` is only propagated to the CLI parameters if `marker.style.label` is set.
```

brianmcgillion:
> You know what would be really neat if you're still motivated and have time? If the scripts would be full derivations at the download location. Then we don't have to spend readers' brain power on explaining them and just assign the output of `import (fetchTarball ...)`.

Sure, I can. Time is a bit limited, hence the 3-hour tutorial was spread over 3/4 days. But I can take a look.

brianmcgillion:
@fricklerhandwerk I am having an issue understanding why it is not lexing the code block starting on line 948. Taking that block into a `bla.nix` file, there are no complaints from the text editor (and it builds and runs).

fricklerhandwerk:
Fixed the formatting. The problem may have been nesting quotes, but not sure.

PR #824, opened 2023-12-04
Thunks tutorial

infinisil:
Almost copied verbatim from https://nixos.wiki/wiki/Nix_Evaluation_Performance which I wrote some time ago.

This needs a bit of work, but I think it's not bad as is either. Feedback appreciated!

### DIFF
--- a/source/tutorials/index.md
+++ b/source/tutorials/index.md
@@ -13,5 +13,6 @@ Packaging existing software <packaging-existing-software.md>
 file-sets.md
 nixos/index.md
 cross-compilation.md
+thunks.md
 module-system/module-system.md
 ```
--- a/source/tutorials/thunks.md
+++ b/source/tutorials/thunks.md
@@ -1,0 +1,151 @@
+(thunks)=
+# Thunks
+
+In Nix, _thunks_ are used to implement laziness.
+A thunk is a type of Nix value that is not yet evaluated.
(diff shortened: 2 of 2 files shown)


fricklerhandwerk:
Looks like concepts, but cool stuff! Seems it only needs a bit of polish.

roberth (changes requested):
It would be great to have this as more official documentation, even if not in the Nix manual. 

Regarding the text, I have a problem in general with the phrasing of "allocating a thunk".
You allocate memory, and in that memory you put a state called a thunk.
If you were to put an actual value there, you would still allocate the same memory.
Thunks and values are necessarily of the same size, so allocating either one is really the same operation as the other.
Furthermore, you may allocate on the stack, and we actually do this for some values/thunks. This is super cheap (when permissible, ie without exacerbating the risk of stack overflows, and since 2.20 we'll stop doing allocations of arbitrary size). So what we really care about is _heap allocations_.

A more appropriate classification would be
- heap objects vs stack objects
- values vs thunks, but both of them being states of objects (unfortunately not the exact terminology used in the implementation though, where any `nThunk` can be in a `Value` struct)

I think we should care mostly about heap objects, many of which happen to be thunks at first. Conversely the thunks we care about tend to be on the heap, because those allocations are more expensive, and they may stick around for a long time.

PR #829, opened 2023-12-06
use upstream page redirects for Nix manual

fricklerhandwerk:
follow-up on #764

~~this currently won't build because the pinned versions don't have the
redirects yet, and bumping them would require an expensive local build
that may fail as it does for Hydra.~~

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -27,25 +27,28 @@ let
     '';
     installPhase =
       let
-        # Various versions of the Nix manuals, grep for (nix-manual)=
-        # FIXME: This requires human interaction to update!
-        # See ./CONTRIBUTING.md for details.
+        # Various versions of the Nix manuals, grep for (nix-manual)= to find where they are displayed
+        # FIXME: This requires human interaction to update! See ./CONTRIBUTING.md for details.
         releases = [
           "2.19"
           "2.18"
           "2.13"
         ];
         inputName = version: pkgs.lib.strings.replaceStrings [ "." ] [ "-" ] version;
-        src = version: (import inputs."nix_${inputName version}").default.doc;
+        src = version: inputs."nix_${inputName version}";
+        manual = version: (import (src version)).default.doc;
         copy = version: ''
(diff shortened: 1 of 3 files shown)


default.nix
```diff
@@ -30,22 +30,26 @@ let
...
         ];
         inputName = version: pkgs.lib.strings.replaceStrings [ "." ] [ "-" ] version;
-        src = version: (import inputs."nix_${inputName version}").default.doc;
+        src = version: inputs."nix_${inputName version}";
+        manual = version: (import (src version)).default.doc;
         copy = version: ''
-          cp -R ${src version}/share/doc/nix/manual/* $out/manual/nix/${version}
+          cp -R ${manual version}/share/doc/nix/manual/* $out/manual/nix/${version}
+          # use upstream page redirects
+          sed 's#^\(.*\) \(.*\) #/nix/manual/${version}\1 /nix/manual/${version}\2 #g' ${src version}/doc/manual/_redirects >> $out/_redirects
```
infinisil, line 45: In the first line it's `manual/nix` and in the third it's `nix/manual`, is that the problem?

infinisil (commented):
CI fails, but the intent of the PR sounds good.

default.nix
```diff
@@ -30,22 +30,26 @@ let
...
         '';
       in
       with pkgs.lib.strings;
       ''
-        mkdir -p $out/manual/nix/{${concatStringsSep "," releases}}
-        ${concatStringsSep "\n" (map copy releases)}
+        mkdir -p $out/manual/nix/{${concatStringsSep "," releases},}
```
infinisil, line 50: Looks like you accidentally added a `,` at the end there. Or is this intentional?

fricklerhandwerk: Intentional, added comment.

default.nix
fricklerhandwerk: Fixed, but it was not the problem.

infinisil:
(CI fails)

fricklerhandwerk:
@infinisil @asymmetric should work now

PR #830, opened 2023-12-06
add mutable shortlinks

### LINKED ISSUE #843: Put a page at /manual/nix
**Observations**

Currently there is a nix manual at https://nix.dev/manual/nix/2.13/language/operators

I tried changing versions by deleting the suffix up to /manual/nix and there was no page there.

**Problem**

Given a URL it should be possible to logically go down the hierarchy to get a version list or similar page. 

**Approaches**

Putting a page there, but I'm not sure what page should be there or by what means it should be put there (and how it interacts with the side site navigation). 

**Willing to help?**

yeah

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

fricklerhandwerk:
closes #843 

this allows mapping cleanly from the existing mutable links on nixos.org

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -29,26 +29,37 @@ let
       let
         # Various versions of the Nix manuals, grep for (nix-manual)= to find where they are displayed
         # FIXME: This requires human interaction to update! See ./CONTRIBUTING.md for details.
-        releases = [
-          "2.19"
-          "2.18"
-          "2.13"
-        ];
+        releases = {
+          latest = "2.19";
+          rolling = "2.18";
+          stable = "2.18";
+          prev-stable = "2.13";
+        };
         inputName = version: pkgs.lib.strings.replaceStrings [ "." ] [ "-" ] version;
         src = version: inputs."nix_${inputName version}";
         manual = version: (import (src version)).default.doc;
         copy = version: ''
           cp -Rf ${manual version}/share/doc/nix/manual/* $out/manual/nix/${version}
-          # add upstream page redirects of the form `<from> <to> <status>`, excluding comment lines and empty
+        '';
+        # add upstream page redirects of the form `<from> <to> <status>`, excluding comment lines and empty
+        redirects = version: ''
           sed '/^#/d;/^$/d;s#^\(.*\) \(.*\) #/manual/nix/${version}\1 /manual/nix/${version}\2 #g' ${src version}/doc/manual/_redirects >> $out/_redirects
         '';
+        shortlink = release: version: ''
+          echo /nix/manual/${release} /nix/manual/${version} >> $out/_redirects
+        '';
+        versions = with pkgs.lib; lists.unique (attrsets.attrValues releases);
       in
+      with pkgs.lib.attrsets;
       with pkgs.lib.strings;
       ''
-        # NOTE: the comma in the shell expansion makes it also work for singleton lists
-        mkdir -p $out/manual/nix/{${concatStringsSep "," releases},}
+        mkdir -p $out
         cp -R build/html/* $out/
-        ${concatStringsSep "\n" (map copy releases)}
+        # NOTE: the comma in the shell expansion makes it also work for singleton lists
+        mkdir -p $out/manual/nix/{${concatStringsSep "," versions},}
+        ${concatStringsSep "\n" (map copy versions)}
+        ${concatStringsSep "\n" (map redirects versions)}
+        ${concatStringsSep "\n" (mapAttrsToList shortlink releases)}
       '';
   };
 


asymmetric (changes requested):
- This needs a `niv update`, otherwise the build fails with:

```
Build finished. The HTML pages are in build/html.
installing
sed: can't read /nix/store/yabz0xx6iyd3h5ba6mv30g7gsjgyqhff-source/doc/manual/_redirects: No such file or directory
error: boost::bad_format_string: format-string is ill-formed
```

- The new links on [this page](http://127.0.0.1:5500/reference/nix-manual.html) are a 404

I'd be happy to pick this up if you're too busy atm.

asymmetric:
Also, sorry for leaving a :x: review on a draft, realized after the fact it could've just been a comment :sweat_smile:

fricklerhandwerk:
As mentioned in the top post, I let it rot waiting for an up-to-date Hydra build to happen. Otherwise we have to build Nix locally. If all versions are in the cache, I can update the pin.

asymmetric:
@fricklerhandwerk I ran it locally and it took a few secs, so yes pretty sure they are.

default.nix
```diff
@@ -9,45 +9,76 @@ let
...
+      };
+      inputName = version: pkgs.lib.strings.replaceStrings [ "." ] [ "-" ] version;
+      src = version: inputs."nix_${inputName version}";
+      manual = version: (import (src version)).default.doc;
+      copy = version: ''
+        cp -Rf ${manual version}/share/doc/nix/manual/* $out/manual/nix/${version}
       '';
-  };
+      redirects = version: ''
+        sed 's#^\(.*\) \(.*\) #/nix/manual/${version}\1 /nix/manual/${version}\2 #g' ${src version}/doc/manual/_redirects >> $out/_redirects
```
asymmetric, line 35: There seems to be a problem with this regex, it matches the lines with comments.

Something like this works:

```suggestion
        sed '/# /d;s#^\(.*\) \(.*\) #/nix/manual/${version}\1 /nix/manual/${version}\2 #g' ${src version}/doc/manual/_redirects >> $out/_redirects
```

default.nix
```diff
@@ -9,45 +9,76 @@ let
...
-        cp -R build/html/* $out/
+  nix-dev =
+    let
+      # Various versions of the Nix manuals, grep for (nix-manual)=
+      # FIXME: This requires human interaction to update!
+      # See ./CONTRIBUTING.md for details.
+      # Using Nix from Nixpkgs could simplify handling, but would requires a bit more setup here
+      releases = {
+        latest = "2.19";
+        rolling = "2.18";
```
asymmetric, line 20: Is rolling the new unstable? If this is not the new accepted term, I would refrain from introducing it here.

source/reference/nix-manual.md
```diff
@@ -5,31 +5,31 @@
...
-- [Nix 2.18](https://nix.dev/manual/nix/2.18/)
+- [Nix @rolling@](https://nix.dev/manual/nix/rolling/)
 
   Reference documentation for the Nix version shipped with the {term}`Nixpkgs` and {term}`NixOS` rolling release
 
-- [Nix 2.18](https://nix.dev/manual/nix/2.18/)
+- [Nix @stable@](https://nix.dev/manual/nix/stable/)
 
-  Reference documentation for the Nix version shipped with the current {term}`Nixpkgs` and {term}`NixOS` stable release
+  Reference documentation for the Nix version shipped with the current {term}`Nixpkgs` and {term}`NixOS` @nixpkgs-stable@ stable release
```
asymmetric, line 28: Maybe we can remove all the "Reference documentation" prefix, since it's in every line.

fricklerhandwerk: What do you suggest concretely? Just removing "reference documentation" would leave "Nix shipped with...", but it's the docs and not Nix itself.

default.nix
fricklerhandwerk: We started using "rolling" in the channel branch FAQ entry, and arguably it's the well-known term for what `nixpkgs-unstable` actually is. Also it's already used that way on the Nix manual overview page here, this just rewords it a bit.

fricklerhandwerk:
Would be great if you can finish it, but otherwise I'll pick it up once I get back to it.

source/reference/nix-manual.md
asymmetric: It would leave "For the Nix version shipped with..."

DanielSidhion (approved):
LGTM

infinisil:
It still looks like Nix 2.13 needs to be built locally:
```
these 2 derivations will be built:
  /nix/store/kbfqff6b2ssqxvmg4hjkrmswg34x8nmy-nix-2.13.7.drv
  /nix/store/lgmy85cgivhxfl8f18yya6k794nvg1in-nix-dev.drv
these 2 paths will be fetched (3.16 MiB download, 65.83 MiB unpacked):
  /nix/store/qlzczyamb1dl0xprqlf57589w38nximq-nix-2.18.1-doc
  /nix/store/21krqgvbinpkh9rb41yx16ybkdcyajnh-nix-2.19.3-doc
```

fricklerhandwerk:
I couldn't reproduce this with a fresh store, i.e. `mkdir tmpstore; nix-build --store $PWD/tmpstore`. It fetches everything from the public cache just fine.

fricklerhandwerk:
I can reproduce with `mkdir tmpstore; nix-build --store $PWD/tmpstore`, but don't understand how that's possible.

https://hydra.nixos.org/eval/1803459 shows that the Hydra build succeeded, but it produces [different store paths](https://hydra.nixos.org/build/245934345#tabs-details) than the local evaluation:

```log
/nix/store/w6dc6i4bi02f3xqpqknd6hppp99mxxm9-nix-2.13.7-debug
/nix/store/ki5j23x8ncimj2sx4bd8y9nr7a956xni-nix-2.13.7-dev
/nix/store/2pcpjixx8ar0wjrf4n3jp0m19vij8jk0-nix-2.13.7-doc
/nix/store/9jdq5q39kv2980z7gq3a7nmfk96wbvhp-nix-2.13.7
```

infinisil:
@fricklerhandwerk I think I figured it out, and it's _very_ cursed! Clearing the Nix cache makes it work for me:

```
mv ~/.cache/{nix,nix-broken}
```

And if that also works for you please comment in https://github.com/NixOS/nix/issues/9905

I can go into more details, but the short of it is:
- For some reason, the flake-compat master tarball is marked as immutable in my fetcher cache, so it always used a version from like a year ago
- At that point in time, flake-compat seemed to double-copy store paths, related is https://github.com/edolstra/flake-compat/issues/60
- But Hydra builds from Flakes, not flake-compat. While Flakes does have [a problem causing double store imports](https://github.com/NixOS/nix/issues/9428), which I initially [thought was the problem](https://github.com/NixOS/nix/issues/9428#issuecomment-1922506375), that only happens for local source imports, which is not the case here.
- And this isn't a problem with 2.16 because of https://github.com/NixOS/nix/pull/8096, which started using `lib.cleanSourceWith` on `self.outPath`. And with `flake-compat`, `self.outPath` is provided by `fetchGit`: https://github.com/edolstra/flake-compat/blob/35bb57c0c8d8b62bbfd284272c928ceb64ddbde9/default.nix#L96. And because `fetchGit` doesn't compose with `lib.cleanSourceWith`, this causes a double store import, making it line up exactly with the double store from the old flake-compat, for entirely different reasons!
- Furthermore, starting with 2.15, flake-compat was getting pinned: https://github.com/NixOS/nix/pull/7989

And here's the kicker, and this is a bit of speculation: You can't check that it works with `--store` because flake-compat does [this](https://github.com/edolstra/flake-compat/blob/0f9255e01c2351cc7d116c072cb317785dd33b33/default.nix#L112-L117) on master, which triggers this issue: https://github.com/NixOS/nix/issues/9852

infinisil (approved):
However, those curses are completely unrelated to this PR here, so this PR is looking good.

fricklerhandwerk:
Wow, thanks a lot for digging into it. I had already slated to `nix-diff` my way in, but that would have been hopeless apparently. Holy `$CURSE` this is messed up.

PR #696, opened 2023-08-24
Add "contributing to Nixpkgs" tutorial

proofconstruction:
While writing *Packaging Existing Software with Nix*, I originally had a section at the end where the reader is walked through preparing the package for submission to Nixpkgs. This is a cleaned-up version of that which maintains continuity with the other tutorial, and should help get people contributing faster.

There is a lot more to be said on this topic, but probably all of *that* belongs in a recipe instead.

PR #831, opened 2023-12-06
[CONTINUING] Add "contributing to Nixpkgs" tutorial

bzm3r:
This is a continuation of: https://github.com/NixOS/nix.dev/pull/696/ 

(As requested by @infinisil : https://github.com/NixOS/nix.dev/pull/696/#issuecomment-1837301546)

It applies various suggestions reviewers requested. Each commit has a link to the relevant comment that inspired it. 

For some commits, I've taken slight liberties to help things flow better (for example: the one addressing using `pkgs/by-name` and the one referring to a better a link to the maintainers README. But I hope the overall effect is that it follows the original work by @proofconstruction as closely as possible.


proofconstruction (commented):
Very glad to see you picking God up!

I made a few simple suggestions to help keep this in line with the style guide and contribution guide.

I'm not sure about the need to create a new `packaging` section; what else do you think should go in there?

source/tutorials/packaging/packaging-existing-software.md
```diff
@@ -14,6 +14,8 @@ In the long term, Nix helps tremendously in alleviating that stress, but when *f
 
 In this tutorial, you'll create your first Nix derivations to package C/C++ software, taking advantage of the [Nixpkgs Standard Environment](https://nixos.org/manual/nixpkgs/stable/#part-stdenv) (`stdenv`) which automates much of the work of building self-contained C/C++ packages.
 
+While an automated tool like [`nix-init`](https://github.com/nix-community/nix-init) can produce Nix derivations from a given software repository for many common project types, it is not uncommon for such an automatic generation to not work, or to encounter project types it does not cover. This tutorial will let you handle such failures without frustration, give you the insight needed to modify automatically generated derivations appropriately for your use case, and perhaps even contribute new generators to `nix-init`!
```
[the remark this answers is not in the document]
fricklerhandwerk: I think it's worthwhile, but maybe as part of "next steps".

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,212 @@
...
+$ nix-shell --run
+```
+
+Assuming that the package builds and run on our machine, it is now ready for submission to Nixpkgs!
+
+## Next Steps
+
+All that remains is to make sure your familiarise yourself with the conventions of [contributing to Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md) and the conventions defining [package maintainers](https://github.com/NixOS/nixpkgs/tree/master/maintainers#readme).
+
+:::{note}
```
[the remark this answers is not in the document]
bzm3r: I too sense the awkwardness you're picking up on here. 

Perhaps what we ought to do is rename this section to `Submitting to Nixpkgs`, change the note into normal text, and merge it with the sentence that begins with "All that remains..." I'll play around with it and put in a commit.

bzm3r:
@proofconstruction 


I am not yet sure if there are additional tutorials that would ultimately fall into this category., but I can think of three reasons for creating a `Packaging` section:

* There is an existing tutorial that could be put under packaging (but I figure that should be a separate PR): https://github.com/NixOS/nix.dev/blob/master/source/tutorials/cross-compilation.md

* I think putting your two tutorials into a packaging section addresses this review comment: https://github.com/NixOS/nix.dev/pull/696#discussion_r1306340467

* Providing some higher level organizational structure to a list of tutorials allows a beginner to understand what sort of tasks/issues make up some broad terms that are usually taken for granted. In this particular case: it helps to implicitly define what "packaging" is?

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+(contributing-to-nixpkgs)=
+
+# Contributing to Nixpkgs
+
+[Nixpkgs](https://github.com/nixos/nixpkgs) is one of the largest public repositories by contributions and *the* largest single collection of software packages available on the web.
+
+In this tutorial, which follows {ref}`the tutorial on packaging existing software <packaging-existing-software>`, you will learn how to prepare newly-packaged software for submission to Nixpkgs, becoming a Nixpkgs contributor in the process.
+
+## Tags, Revisions, and Releases
+An earlier version of the `icat.nix` definition created in the {ref}`Packaging Existing Software <packaging-existing-software>` tutorial used the `master` revision of the [upstream `icat` repository](https://github.com/atextor/icat), which would be suitable for individual use but isn't good practice for submission to Nixpkgs.
```
infinisil, line 18:
```suggestion
An earlier version of the `icat.nix` definition created in the {ref}`Packaging Existing Software <packaging-existing-software>` tutorial used the `master` revision of the [upstream `icat` repository](https://github.com/atextor/icat), which could be suitable for individual use but isn't good practice for submission to Nixpkgs.
```

`master` isn't necessarily suitable for individual use :)

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+
+It is much better to use a fixed revision corresponding to a particular release version of the software so maintainers (perhaps you!) can easily check whether the package should be updated.
+
+This was already done in the previous tutorial, but as a reminder, here are the steps:
+1. Check the upstream repository for tags or releases. On GitHub, these are available at `https://github.com/<owner>/<repo>/tags` and `https://github.com/<owner>/<repo>/releases` respectively.
+2. Choose the latest tag or release and write it into a string literal as the value of the `rev` attribute passed to the [fetcher](https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers).
```
infinisil, line 24:
```suggestion
2. Choose the latest tag or release and write it into a string literal as the value of the `rev` attribute passed to a Git [fetcher](https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers).
```

`rev` only makes sense for Git fetchers

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+3. Use `nix-prefetch-url` or a similar tool to fetch, unpack, and hash the files marked by the tag or release you selected.
+4. Alternatively, you can also use a tool such as `nix-init` to automatically generate a required derivation for some common project types, or use its output to help lay down a skeleton implementation of the above steps.
```
infinisil, line 26:
```suggestion
4. Alternatively, you can also use a tool such as [`nix-init`](https://github.com/nix-community/nix-init) to automatically generate a required derivation for some common project types, or use its output to help lay down a skeleton implementation of the above steps.
```

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Tags, Revisions, and Releases
+An earlier version of the `icat.nix` definition created in the {ref}`Packaging Existing Software <packaging-existing-software>` tutorial used the `master` revision of the [upstream `icat` repository](https://github.com/atextor/icat), which would be suitable for individual use but isn't good practice for submission to Nixpkgs.
+
+It is much better to use a fixed revision corresponding to a particular release version of the software so maintainers (perhaps you!) can easily check whether the package should be updated.
+
+This was already done in the previous tutorial, but as a reminder, here are the steps:
+1. Check the upstream repository for tags or releases. On GitHub, these are available at `https://github.com/<owner>/<repo>/tags` and `https://github.com/<owner>/<repo>/releases` respectively.
+2. Choose the latest tag or release and write it into a string literal as the value of the `rev` attribute passed to the [fetcher](https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers).
+3. Use `nix-prefetch-url` or a similar tool to fetch, unpack, and hash the files marked by the tag or release you selected.
```
infinisil, line 25:
```suggestion
3. Use `nix-prefetch-url` (comes with any Nix installation) or [another `nix-prefetch-*` tool](https://search.nixos.org/packages?type=packages&query=nix-prefetch) to fetch, unpack, and hash the files marked by the tag or release you selected.
```

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+4. Alternatively, you can also use a tool such as `nix-init` to automatically generate a required derivation for some common project types, or use its output to help lay down a skeleton implementation of the above steps.
+
+In the `icat` example, the source was hosted on GitHub, so `fetchFromGitHub` was used for the `src` attribute passed to `stdenv.mkDerivation`, with `rev = "v0.5"` corresponding to the latest tag, and a SHA256 hash generated by
+
+```console
+nix-prefetch-url --unpack https://github.com/atextor/icat/archive/refs/tags/v0.5.tar.gz --type sha256`
+```
+
+Here is the final version of `icat.nix` from the previous tutorial:
```
infinisil, line 34: I don't think this needs to be repeated. We should assume that users read the previous tutorial.

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+
+```console
+$ nix-build -A icat
+...
+/nix/store/g6w508vxwr3df25dnl4k3xvcr4pqxprj-icat
+```
+
+The (omitted) output indicates that the package successfully built, and the very last line reports where Nix put the build result.
+
+## Phases and Hooks
```
infinisil, line 89: This section seems to be fully covered in the other tutorial, so I don't think we need it here.

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+    cp icat $out/bin
++   runHook postInstall
+'';
+...
+```
+
+A number of other hooks are available, and when packaging for Nixpkgs you should remember to include calls to these hooks in the phases you define, to make it easy for other contributors to [override](https://nixos.org/manual/nixpkgs/stable/#chap-overrides) parts of your derivations in the future.
+
+## Package Metadata
+The Nixpkgs [contribution guidelines](https://nixos.org/manual/nixpkgs/stable/#part-contributing) require all packages in Nixpkgs to have a `meta` attribute in their derivation, which contains information like a description of the package, the homepage of the project it belongs to, the software license, which platforms the package can be built for, and a list of Nixpkgs maintainers for the package.
```
infinisil, line 111:
```suggestion
The Nixpkgs [package contribution guidelines](https://github.com/NixOS/nixpkgs/tree/master/pkgs#readme) require all packages in Nixpkgs to have a `meta` attribute in their derivation, which contains information like a description of the package, the homepage of the project it belongs to, the software license, which platforms the package can be built for, and a list of Nixpkgs maintainers for the package.
```

This was moved in the most recent release.

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+This eliminates the need to decide upon a particular category if a package fits into multiple, and it also eliminates the requirement of [updating `all-packages.nix` when using the category-based hierarchy](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#versioning), and since our case isn't bound by any naming limitations, let's use the two-letter prefix system:
+
+```console
+$ mkdir nixpkgs/pkgs/by-name/ic/icat
+```
+
+You should rename `icat.nix` when you copy it to its new home, to keep to the convention of package derivation files [being named `package.nix`](https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#example) when using the two-letter prefix system:
+
+```console
+$ mv icat.nix nixpkgs/pkgs/applications/graphics/icat/default.nix
```
infinisil, line 155:
```suggestion
$ mv icat.nix nixpkgs/pkgs/by-name/ic/icat/package.nix
```

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+  imgcat = callPackage ../applications/graphics/imgcat { };
+
+  img-cat = callPackage ../applications/graphics/img-cat { };
+...
+```
+
+:::{note}
+Attributes in `all-packages.nix` are arranged *approximately* in alphabetical order, either by attribute name or by the directory containing their corresponding `default.nix`.
+:::
+:::
```
infinisil, line 179: I don't think we should go into detail on the old category hierarchy. In almost all cases it's not necessary anymore. We can just link to this instead: https://github.com/NixOS/nixpkgs/tree/master/pkgs#category-hierarchy

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+
+  buildInputs = [ imlib2 xorg.libX11.dev ];
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/bin
+    cp icat $out/bin
+    runHook postInstall
+  '';
+}
```
infinisil, line 64:
```suggestion
stdenv.mkDerivation (finalAttrs: {
  pname = "icat";
  version = "0.5";

  src = fetchFromGitHub {
    owner = "atextor";
    repo = "icat";
    rev = "v${finalAttrs.version}";
    sha256 = "0wyy2ksxp95vnh71ybj1bbmqd5ggp13x3mk37pzr99ljs9awy8ka";
  };

  buildInputs = [ imlib2 xorg.libX11.dev ];

  installPhase = ''
    runHook preInstall
    mkdir -p $out/bin
    cp icat $out/bin
    runHook postInstall
  '';
})
```

The `version` variable was undefined.

source/tutorials/packaging/contributing-to-nixpkgs.md
```diff
@@ -0,0 +1,222 @@
...
+...
+/nix/store/vr2vk8z8839l5j6gra0qlyxrh5sarmh4-icat
+```
+
+The build completed as expected, and produced a `result` symbolic link in the current directory.
+
+It's good practice to confirm that the program actually runs:
+
+```console
+$ nix-shell --run
```
infinisil, line 198:
```suggestion
$ wget https://raw.githubusercontent.com/NixOS/nixos-artwork/master/logo/white.png
$ result/bin/icat white.png
```

infinisil (commented):
This is the first time taking a closer look at this. I question whether nix.dev is the right place for this. Because this will have the same problems as described in https://github.com/NixOS/nixpkgs/issues/244056 and https://github.com/NixOS/nix/issues/9583, which is why there's a push to move contributor docs into the source code directly.

And in particular, this tutorial covers a lot of topics that are already in the Nixpkgs contributor docs, a lot of which was revised to some degree in https://github.com/NixOS/nixpkgs/pull/245243:

- [A guide on how to make a PR](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#how-to-create-pull-requests)
- [A guide to contributing a package](https://github.com/NixOS/nixpkgs/tree/master/pkgs#quick-start-to-adding-a-package)
- [A section on meta attributes](https://github.com/NixOS/nixpkgs/tree/master/pkgs#meta-attributes), though not as nicely as written as here.
- Various other sections on some of the topics covered here

I'd prefer the contribution docs in the source code to be improved rather than a new tutorial written.
This way we can just link to a good contribution doc entry-point on GitHub from the Packaging existing software tutorial.

source/tutorials/packaging/contributing-to-nixpkgs.md
bzm3r: @infinisil I'm not convinced that's a fair assumption, just due to the reality of how people work. However, I do think that we should instead insert a short hyperlink rather than repeating it verbatim.

source/tutorials/packaging/contributing-to-nixpkgs.md
bzm3r: @infinisil Sounds good, I'll convert this into a short hyperlink.

source/tutorials/packaging/contributing-to-nixpkgs.md
bzm3r: @infinisil Should we just give it some placeholder like `0.1` then? In order to keep the focus on the content rather than on how `mkDerivation` is implemented? (I'd need to chew on this a bit myself before I can totally understand why your change works. I believe it has to do with how `mkDerivation` is defined in the first place, [in that it can take both a function or attribute set as an argument](https://github.com/NixOS/nixpkgs/blob/3a16c6447466f4034c2d75fe7014477142c9513e/pkgs/stdenv/generic/make-derivation.nix#L600), so that it is ultimately something like a fixed point function? Totally new to me, as I never dove into the source code on this point until trying to figure out why your change would work.)

bzm3r:
@infinisil Thanks so much these were all very useful suggestions. I've merged almost all of them, except for the ones that will require a bit more manual work on my part, or the ones that I have asked for further clarification on.

> I'd prefer the contribution docs in the source code to be improved rather than a new tutorial written.

I'm not sure I am a fan of this idea, but it's a confusing feeling because, at first glance, I felt myself being pulled in two different directions:, I think this is the way it *should* be done "ideally" (a central, single source of information), and on the other hand, I don't want @proofconstruction to feel that their work was a wasted effort. 

*Especially because this content did significantly improve my understanding of things. *

I will put in more thought into this, but I think the true ideal might actually lie in some sort of middle ground. Will write it out once I have a clearer picture.

bzm3r:
@proofconstruction Could I have some clarification on what you mean by: "Also, can you rebase this and submit it with the original PR as the merge target?"

If I understand you correctly:

* rebase this onto the original PR

* submit a PR to your repo that merges this work the original PR?

infinisil:
> > I'd prefer the contribution docs in the source code to be improved rather than a new tutorial written.
> 
> I'm not sure I am a fan of this idea, but it's a confusing feeling because, at first glance, I felt myself being pulled in two different directions:, I think this is the way it _should_ be done "ideally" (a central, single source of information), and on the other hand, I don't want @proofconstruction to feel that their work was a wasted effort.
> 
> _Especially because this content did significantly improve my understanding of things._
> 
> I will put in more thought into this, but I think the true ideal might actually lie in some sort of middle ground. Will write it out once I have a clearer picture.

I think I'm fine with having this for now. At least the discoverability and smaller scope of it is indeed something `CONTRIBUTING.md` can't give, _yet_. I do think we can both split up `CONTRIBUTING.md` more in the future, and maybe even proxy-render it on `nix.dev`.

proofconstruction:
> This is the first time taking a closer look at this. I question whether nix.dev is the right place for this. Because this will have the same problems as described in [NixOS/nixpkgs#244056](https://github.com/NixOS/nixpkgs/issues/244056) and [NixOS/nix#9583](https://github.com/NixOS/nix/issues/9583), which is why there's a push to move contributor docs into the source code directly.
> 
> And in particular, this tutorial covers a lot of topics that are already in the Nixpkgs contributor docs, a lot of which was revised to some degree in [NixOS/nixpkgs#245243](https://github.com/NixOS/nixpkgs/pull/245243):
> 
> * [A guide on how to make a PR](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#how-to-create-pull-requests)
> * [A guide to contributing a package](https://github.com/NixOS/nixpkgs/tree/master/pkgs#quick-start-to-adding-a-package)
> * [A section on meta attributes](https://github.com/NixOS/nixpkgs/tree/master/pkgs#meta-attributes), though not as nicely as written as here.
> * Various other sections on some of the topics covered here
> 
> I'd prefer the contribution docs in the source code to be improved rather than a new tutorial written. This way we can just link to a good contribution doc entry-point on GitHub from the Packaging existing software tutorial.

Actually yes, since [nixpkgs#245243](https://github.com/NixOS/nixpkgs/pull/245243) landed, this is no longer necessary in nix.dev. I think we should close this PR and #696 and move any useful text into the sources listed above.

PR #832, opened 2023-12-07
add FAQ entry on how to run non-nix executables

tejing1:
Adds an FAQ entry, primarily intended as a link target for the error message introduced in nixos/nixpkgs#269551.

Given that usecase, I do want to ask how much commitment to link stability is there on nix.dev? Is there perhaps a method to create a stable link that redirects to the proper location, so that external links can remain stable even if the site structure is changed?

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -25,3 +25,4 @@
 /recipes/direnv /guides/recipes/direnv 301
 /tutorials/learning-journey/sharing-dependencies /guides/recipes/sharing-dependencies 301
 /tutorials/learning-journey/packaging-existing-software /tutorials/packaging-existing-software 301
+/permalink/stub-ld /guides/faq#how-to-run-non-nix-executables 301
--- a/source/guides/faq.md
+++ b/source/guides/faq.md
@@ -49,6 +49,54 @@ See <https://github.com/nix-community/home-manager>
 
 ## NixOS
 
+### How to run non-nix executables?
+
+NixOS cannot run dynamically linked executables intended for generic Linux environments out of the box.
+This is because, by design, it does not have a global library path, nor does it follow the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) (FHS).
+
+There are a few ways to resolve this mismatch in environment expectations:
+
+- Use the version packaged in Nixpkgs, if there is one.
+  You can search available packages at <https://search.nixos.org/packages>.
+
+- Write a Nix expression for the program to package it in your own configuration.
+
+  There are multiple approaches to this:
+  - Build from source.
+
+    Many open-source programs are highly flexible at compile time in terms of where their files go.
+    For an introduction to this, see [](../tutorials/packaging-existing-software).
+  - Modify the program's [ELF header](https://en.wikipedia.org/wiki/Executable_and_Linkable_Format) to include paths to libraries using [`autoPatchelfHook`](https://nixos.org/manual/nixpkgs/stable/#setup-hook-autopatchelfhook).
+
+    Do this if building from source isn't feasible.
+  - Wrap the program to run in an FHS-like environment using [`buildFHSEnv`](https://nixos.org/manual/nixpkgs/stable/#sec-fhs-environments).
+
+    This is a last resort, but sometimes necessary, for example if the program downloads and runs other executables.
+
+- Create a library path that only applies to unpackaged programs by using [`nix-ld`](https://github.com/Mic92/nix-ld).
+  Add this to your `configuration.nix`:
+
+  ```nix
+    programs.nix-ld.enable = true;
+    programs.nix-ld.libraries = with pkgs; [
+      # Add any missing dynamic libraries for unpackaged programs
+      # here, NOT in environment.systemPackages
+    ];
+  ```
+
+  Then run `nixos-rebuild switch`, and log out and back in again to propagate the new environment variables.
+  (This is only necessary when enabling `nix-ld`; changes in included libraries take effect immediately on rebuild.)
+
+  :::{note}
+  `nix-ld` does not work for 32-bit executables on `x86_64` machines.
+  :::
+
+- Run your program in the FHS-like environment made for the Steam package using [`steam-run`](https://nixos.org/manual/nixpkgs/stable/#sec-steam-run):
+
+  ```shell-session
+  $ nix-shell -p steam-run --run "steam-run <command>"
+  ```
+
 ### How to build my own ISO?
 
 See <http://nixos.org/nixos/manual/index.html#sec-building-image>


roberth:
I would prefer to have a separate page for this, or even a dedicated "landing URL" that redirects to whichever resource is most appropriate. Such a URL would also have a clear intent, making it very unlikely to break.

In fact, I think such a decoupling would be great, and it would immediately unblock this PR if you could add a redirect in `/_redirects`.

What's more, we could get some statistics about such errors by making the URL unique.

Perhaps

```
# Stable URLs for help with diagnosing problems
# Refer to these URLs with query parameters that provide extra context
# e.g. /diagnostic/no-ld?nixos=1 to allow for later specialisation of the docs.
# Similarly, nixpkgs may refer to /diagnostics/foo?os=${stdenv.hostPlatform.parsed.kernel.name}&cpu=${stdenv.hostPlatform.parsed.cpu.name} (or some ready-made string of parameters)
/diagnostic/no-ld /guides/faq#...
```

Not sure what's the exact syntax for the `#` fragment there.

cc @fricklerhandwerk

roberth:
NB this is currently blocking https://github.com/NixOS/nixpkgs/pull/269551, and it would be unconstitutional to punish authors of new features for writing docs.

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
...
+  There are multiple approaches:
+  - Build from source.
+    For an introduction to this, see [](../tutorials/packaging-existing-software).
+  - Set the program's ELF metadata to include paths to libraries using [`autoPatchelfHook`](https://nixos.org/manual/nixpkgs/stable/#setup-hook-autopatchelfhook).
+    Do this if building from source isn't feasible.
+  - Wrap the program to run in a normal-seeming container using [`buildFHSEnv`](https://nixos.org/manual/nixpkgs/stable/#sec-fhs-environments).
+    This is a last resort, but sometimes necessary.
+    If the program downloads and runs other executables, for example.
+
+- Create a library path that only applies to unpackaged programs by using `nix-ld`.
```
infinisil, line 71:
```suggestion
- Create a library path that only applies to unpackaged programs by using [`nix-ld`](https://github.com/Mic92/nix-ld).
```

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
...
+  Note that this does not work for 32-bit executables on `x86_64` machines.
```
infinisil, line 72: This seems a bit specific, why does it not work? Is it actually due to a fundamental reason or just because of https://github.com/NixOS/nixpkgs/pull/269919?

And this would be best as a note using
```
:::{note}
Text here
:::
```

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
...
+  Add this to your `configuration.nix`, rebuild, and relogin to propagate the new environment variables:
```
infinisil, line 73:
```suggestion
  Add this to your `configuration.nix` and run `nixos-rebuild switch`:
```

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
...
+
+  ```nix
+  programs.nix-ld.enable = true;
+  programs.nix-ld.libraries = with pkgs; [
+    # Add any missing dynamic libraries for unpackaged programs
+    # here, NOT in environment.systemPackages
+  ];
+  ```
```
infinisil, line 81:
~~~suggestion
  ```
  
  Then log out and back in again.
~~~

A bit harder to miss this way.

Also why is this required at all? Ideally this wouldn't be necessary.

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
...
+
+- Run your program in the normal-seeming container made for the steam package.
+  Install `steam-run` and run your program with:
+
+  ```shell-session
+  $ steam-run <command>
+  ```
```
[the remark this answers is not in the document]
infinisil: Also we should probably have a package called something like `fhs-run` (or even a `pkgs.fhsRun.withPackages`) that doesn't depend on steam.

infinisil (changes requested):
Reviewed in the docs team meeting today :)

source/guides/faq.md
tejing1: I only know the `nix-ld` module doesn't try to install it for 32 bit, and the `pkgs.pkgsi686Linux.nix-ld` doesn't build. I didn't look into how difficult it would be to fix the issue.

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
 
 ## NixOS
 
+### How to run non-nix executables?
+
+NixOS cannot run dynamically linked executables intended for generic linux environments out of the box.
+This is because, by design, it does not have a global library path.
+Solutions include:
```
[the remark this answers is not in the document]
tejing1: This is further from the core issue, in my view. We could support this without being at all FHS-like. `nix-ld` is proof of that.

The core issue is that, unlike distros that shove all their libraries into a shared lookup path, we don't automatically inherit a serviceable library path from the other programs on the system, and would have to explicitly create one solely for non-nix executables, as `nix-ld` does.

Also, I think far more readers will immediately understand that this restriction is somewhat fundamental to the software architecture if we say there is no global library path. Talking about the FHS makes it sound like we just want to be different for not good reason, and don't care about the system being usable.

The way nix packages programs to find their libraries and other dependencies in a separate lookup path for each program should perhaps get an article in "concepts", which could be linked here.

source/guides/faq.md
```diff
@@ -49,6 +49,44 @@ See <https://github.com/nix-community/home-manager>
...
+- Use the version packaged in nixpkgs, if there is one.
+  You can search available packages at <https://search.nixos.org/packages>.
+
+- Write a nix expression for the program to package it in your own configuration.
+  There are multiple approaches:
+  - Build from source.
+    For an introduction to this, see [](../tutorials/packaging-existing-software).
+  - Set the program's ELF metadata to include paths to libraries using [`autoPatchelfHook`](https://nixos.org/manual/nixpkgs/stable/#setup-hook-autopatchelfhook).
+    Do this if building from source isn't feasible.
+  - Wrap the program to run in a normal-seeming container using [`buildFHSEnv`](https://nixos.org/manual/nixpkgs/stable/#sec-fhs-environments).
```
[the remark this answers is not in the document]
tejing1: I was trying to avoid the word "FHS" in this message. It's jargon that's basically only known in the nixos community, because we need word for what we aren't. The rest of the world doesn't need a word for it. And this message is likely to be many reader's introduction to the fact that nixos is different in this way.

source/guides/faq.md
tejing1: Reworked that so the directions for after editing `configuration.nix` are after the code block.

tejing1:
I didn't know about `_redirects`, but unfortunately it doesn't work in `devmode`, so I can't easily test whether redirection to a specific url fragment (`#foo`) in a page is possible. I suspect it isn't, though. If I am correct in that suspicion, then having a stable url through redirects would only be feasible if it's a dedicated page.

roberth:
`nix shell nixpkgs#netlify-cli -c netlify dev` may help, providing a local dev server that supports it.

infinisil:
I also tried out netlify-cli earlier. It works, but it's a bit big on dependencies, and you need to also add `deno` so it doesn't give an ugly warning. But it does work, even opens the page in a browser by default:

```
$ nix-build -A build
$ nix-shell -p netlify-cli deno --run 'netlify dev -d result'
```

And I can confirm that redirects to URLs with a `#` work just fine :)

tejing1:
Ok, then. What should the stable path be? `/landing/stub-ld`?

tejing1:
`/diagnostic` seems odd to me, since it's not really about the main purpose (as I see it), of making a stable link. Maybe just `/permalink/stub-ld`? Can't get much clearer about what it's for than that. It's a permalink for `stub-ld` to link to.

infinisil (approved):

fricklerhandwerk:
I'm really not convinced of pre-emptively creating permalinks. Every link should be a permalink: If someone decides to move a file, they should add a redirect. Eventually we should enforce this in CI and then automate creating the redirect.

infinisil:
In this case we depend on an _anchor_ though, which for nix.dev are automatically generated from the section heading. Ideally we could have persistent anchors too, but for now we don't, so I think the /permalink workaround is fine for the time being.

tejing1:
It's technically impossible to redirect an anchor, except by javascript shenanigans, since it isn't even sent to the server. So, if we wanted to separate this from the faq later, redirects couldn't do it without this kind of preparation.

_redirects
```diff
@@ -25,3 +25,4 @@
 /recipes/direnv /guides/recipes/direnv 301
 /tutorials/learning-journey/sharing-dependencies /guides/recipes/sharing-dependencies 301
 /tutorials/learning-journey/packaging-existing-software /tutorials/packaging-existing-software 301
+/permalink/stub-ld /guides/faq#how-to-run-non-nix-executables 301
```
roberth, line 28:
```suggestion
# incoming permalinks
/in/stub-ld /guides/faq#how-to-run-non-nix-executables 301
```

Brevity worth considering.

roberth (approved):

tejing1:
Thought I should avoid the idiom "bridge this gap", on further thought. It may not be readily understandable to non-native speakers.

_redirects
tejing1: I feel like that's a bit too much brevity. Having it be obvious what it's for just looking at the url is more useful than it being really short, I think.

fricklerhandwerk:
Agreed on permalinks being more accessible than JavaScipt shenanigans. I'd like to give it a last pass to ensure clarity for outside observers and then merge, if that's alright for you @tejing1. It may be a few days until I get to it though.

infinisil:
@fricklerhandwerk Unless there's incorrect information in the PR, I'd much rather prefer follow-up PRs!

tejing1:
Sooner is better, especially since it's blocking another PR, but I don't want to rush past the appropriate level of review, either. Do what needs to be done.

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
 ## NixOS
 
+### How to run non-nix executables?
+
+NixOS cannot run dynamically linked executables intended for generic linux environments out of the box.
+This is because, by design, it does not have a global library path, nor does it follow the traditional linux standard for where program files are stored on the filesystem.
+
+:::{note}
+The NixOS community uses the term "FHS" for traditional linux environments, which is short for "[Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html)".
+:::
```
fricklerhandwerk, line 59:
```suggestion
This is because, by design, it does not have a global library path, nor does it follow the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) (FHS).
```

There is nothing special about the Nix community using that abbreviation, and mentioning it should be enough.

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+
+There are a few ways to resolve this mismatch in environment expectations:
+
+- Use the version packaged in nixpkgs, if there is one.
```
fricklerhandwerk, line 63:
```suggestion
- Use the version packaged in Nixpkgs, if there is one.
```

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+  You can search available packages at <https://search.nixos.org/packages>.
+
+- Write a nix expression for the program to package it in your own configuration.
```
fricklerhandwerk, line 66:
```suggestion
- Write a Nix expression for the program to package it for your own configuration.

```

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+  There are multiple approaches to this:
+  - Build from source.
+    Many open-source programs are highly flexible at compile time in terms of where their files go.
+    For an introduction to this, see [](../tutorials/packaging-existing-software).
+  - Set the program's ELF metadata to include paths to libraries using [`autoPatchelfHook`](https://nixos.org/manual/nixpkgs/stable/#setup-hook-autopatchelfhook).
```
fricklerhandwerk, line 71:
```suggestion
  - Set the program's [ELF header](https://en.wikipedia.org/wiki/Executable_and_Linkable_Format) to include paths to libraries using [`autoPatchelfHook`](https://nixos.org/manual/nixpkgs/stable/#setup-hook-autopatchelfhook).

```

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+:::
+
+There are a few ways to resolve this mismatch in environment expectations:
+
+- Use the version packaged in nixpkgs, if there is one.
+  You can search available packages at <https://search.nixos.org/packages>.
+
+- Write a nix expression for the program to package it in your own configuration.
+  There are multiple approaches to this:
+  - Build from source.
```
fricklerhandwerk, line 68:
```suggestion
  - Build from source.

```

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+    Many open-source programs are highly flexible at compile time in terms of where their files go.
+    For an introduction to this, see [](../tutorials/packaging-existing-software).
+  - Set the program's ELF metadata to include paths to libraries using [`autoPatchelfHook`](https://nixos.org/manual/nixpkgs/stable/#setup-hook-autopatchelfhook).
+    Do this if building from source isn't feasible.
+  - Wrap the program to run in an FHS-like environment using [`buildFHSEnv`](https://nixos.org/manual/nixpkgs/stable/#sec-fhs-environments).
```
fricklerhandwerk, line 73:
```suggestion
  - Wrap the program to run in an FHS-like environment using [`buildFHSEnv`](https://nixos.org/manual/nixpkgs/stable/#sec-fhs-environments),

```

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+    This is a last resort, but sometimes necessary.
+    If the program downloads and runs other executables, for example.
```
fricklerhandwerk, line 75:
```suggestion
    This is a last resort, but sometimes necessary, for example if the program downloads and runs other executables.
```

source/guides/faq.md
```diff
@@ -49,6 +49,55 @@ See <https://github.com/nix-community/home-manager>
...
+  ```
+
+  Then run `nixos-rebuild switch`, and log out and back in again to propagate the new environment variables.
+  (This is only necessary when enabling `nix-ld`; changes in included libraries take effect immediately on rebuild.)
+
+  :::{note}
+  `nix-ld` does not work for 32-bit executables on `x86_64` machines.
+  :::
+
+- Run your program in the FHS-like environment made for the steam package using [`steam-run`](https://nixos.org/manual/nixpkgs/stable/#sec-steam-run):
```
fricklerhandwerk, line 95:
```suggestion
- Run your program in the FHS-like environment made for the Steam package using [`steam-run`](https://nixos.org/manual/nixpkgs/stable/#sec-steam-run):
```

fricklerhandwerk (approved):
Only formatting for better readability of list items.

source/guides/faq.md
tejing1: I actually think "in" is better. Users unfamiliar with the nix language often don't realize that they can locate such an expression in or alongside their `configuration.nix`, and assume they need to have a local checkout of nixpkgs in order to have their own packages.

source/guides/faq.md
tejing1: Why make this one end in a comma, when all the other options begin with a sentence saying what the approach is, and then separate sentences with commentary?

source/guides/faq.md
fricklerhandwerk: Makes sense, but the intent is certainly not evident from just that word. Fixed in a force push.

source/guides/faq.md
tejing1: We're the only ones who commonly use the term at all, though, abbreviated or not. So it almost certainly needs to be introduced as new jargon to those reading this. I'm not really comfortable using the word here without a prominent introduction to what it means.

I'm a bit on the fence about whether it's prominent enough without the note. It's certainly shorter and more straightforward, which is nice, but not everyone reads linearly, and I want the introduction of the term to catch the eye.

_redirects
tejing1: I think `permalink` is a little better, but I don't care much either way.

`topic` *might* conflict with a future desired organizational structure. Low chance, though.

source/guides/faq.md
tejing1: Yeah, alright, I'm good with it. The different coloring of the link is enough, I think.

_redirects
infinisil: I'm also for `permalink`, it expresses exactly what it's for.

PR #842, opened 2023-12-15
reorganise CODEOWNERS

fricklerhandwerk:
I'm about to reduce workload for a while and want to drastically reduce the number of notifications.
AFAICT it's not possible to unsubscribe from team review requests, but I don't want to leave the team since I'll still be active.

therefore, instead of pinging the entire team, assign subscriptions explicitly

@alejandrosame @asymmetric @henrik-ch @infinisil @olafklingt @proofconstruction please check if this arrangement makes sense for you

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,11 +11,15 @@
 # This also holds true for GitHub teams. Since almost none of our teams have write
 # permissions, you need to list all members of the team with commit access individually.
 
-* @NixOS/documentation-team
+* @NixOS/documentation-reviewers
 
 /.github/workflows/ @asymmetric
 
 /source/tutorials/file-sets.md @infinisil
+source/guides/recipes/dependency-management.md @infinisil
+source/guides/recipes/python-environment.md @alejandrosame
+source/tutorials/module-system @infinisil @asymmetric @proofconstruction
+source/tutorials/packaging-existing-software.md @proofconstruction
 
 /source/tutorials/integration-testing-using-virtual-machines.md @olafklingt
 /source/tutorials/nixos-configuration-on-vm.md @olafklingt


asymmetric (commented):
This introduces a bit of additional overhead. Also we should find a general solution that works for everyone, rather than special case you, perhaps (for example, assuming at some point we go back to just having the team in `CODEOWNERS` -- then when someone else needs to take some time off, we'd have to re-do this PR all over again).

What if instead, you configured [notification routing](https://github.com/settings/notifications/custom_routing) to have group emails skip your inbox, be automatically deleted, etc? Or more drastically, you could temporarily [ignore](https://github.com/watching) the nix.dev repo.

fricklerhandwerk:
Yeah definitely, this is why I'm asking. I don't get any GitHub emails, so notification routing doesn't work here. AFAICT one can't ignore group notifications, and certainly not for specific teams. There are not that many knobs to twiddle here it seems.

infinisil:
How about creating a documentation _subteam_ for receiving notifications, like @NixOS/documentation-notifications. We can add everybody except @fricklerhandwerk to that subteam and put the subteam in CODEOWNERS to receive all notifications.

fricklerhandwerk:
Could be @NixOS/documentation-reviewers, because that's indeed a task not everyone participates in (for good reason)

.github/CODEOWNERS
```diff
@@ -11,11 +11,15 @@
 # This also holds true for GitHub teams. Since almost none of our teams have write
 # permissions, you need to list all members of the team with commit access individually.
 
-* @NixOS/documentation-team
+* @alejandrosame @asymmetric @henrik-ch @infinisil @olafklingt @proofconstruction
```
infinisil, line 14:
```suggestion
* @NixOS/documentation-reviewers 
```

Just created https://github.com/orgs/nixos/teams/documentation-team

infinisil (changes requested):

.github/CODEOWNERS
asymmetric: Is this discoverable by people who are not in the team? How will they know to tag this team and not the -team one?

infinisil: I guess if somebody uses the teams autocompletion they'll likely pick the -reviewers one instead of the -team one.

infinisil: Now pushed this change

PR #844, opened 2023-12-19
Niv bump workflow

asymmetric:
This keeps niv sources up to date.

<del>This removes the need for the manual step.</del>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,15 +27,6 @@ Adapted from the [Contributor Covenant] and [The Carpentries Code of Conduct]:
 With the current setup, the Nix manual hosted on nix.dev does not get updated automatically with new releases.
 The following manual steps are required:
 
-- Regularly update the inputs to use the latest versions of the Nix release branches with `nix shell --run "niv update"`
-
-  To avoid long build times, make sure Nix can be fetched from the cache.
-  If it doesn't, find the latest commit that is [built by Hydra](https://hydra.nixos.org/project/nix). For example, to pin Nix 2.18:
-
-  ```bash
-  niv update nix_2-18 -r f5f4de6a550327b4b1a06123c2e450f1b92c73b6
-  ```
-
 - On each new Nix release:
 
   1. Add the latest version in [`default.nix`](./default.nix).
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -27,15 +27,6 @@ Adapted from the [Contributor Covenant] and [The Carpentries Code of Conduct]:
 With the current setup, the Nix manual hosted on nix.dev does not get updated automatically with new releases.
 The following manual steps are required:
 
-- Regularly update the inputs to use the latest versions of the Nix release branches with `nix shell --run "niv update"`
-
-  To avoid long build times, make sure Nix can be fetched from the cache.
-  If it doesn't, find the latest commit that is [built by Hydra](https://hydra.nixos.org/project/nix). For example, to pin Nix 2.18:
```
infinisil, line 33: This is kind of important, both because we don't want to render docs for branches that haven't been released yet, but also because we don't have any nix.dev cache that would cache such Nix builds for contributors.

asymmetric: You're right, plus this is unrelated to the change I've made. I'll revert.

asymmetric:
I'm actually not sure this is useful.. I thought it could help us track latest nixpkgs, but looking better, it doesn't seem possible (old problem of not having a "stable" channel).

I'll close for now.

fricklerhandwerk:
What's the problem with tracking Nixpkgs?

asymmetric:
@fricklerhandwerk I don't know if it adds much value to just periodically bump to the latest commit of pretty stable branches. We shouldn't uncover any bugs or new features like that.

What we might want is auto bump to a new nixos release, or auto add new Nix versions, but this doesn't do any of that.

fricklerhandwerk:
It would be relevant for `nixpkgs-unstable` and patch releases of Nix though, because we definitely want to catch movements there as quickly as possible.

asymmetric:
But we don't track unstable in `sources.json`. I'm not familiar with the `X.YZ-maintenance` branches of nix, but would keeping up to the Z digits really earn us much? What's Nix's versioning policy actuallyt?

fricklerhandwerk:
We're not tracking Nixpkgs yet, but once we host the manual we'll need to.

What do you mean by versioning policy? Tracking Nix maintenance branches may get us patch release notes. This is indeed lower priority because most patches don't come with a note.

asymmetric:
> We're not tracking Nixpkgs yet, but once we host the manual we'll need to.

Alright, then let's revisit this once we do that.

> What do you mean by versioning policy? Tracking Nix maintenance branches may get us patch release notes. This is indeed lower priority because most patches don't come with a note.

Oh I see, I was thinking about code, you were thinking about manuals. Either way, it seems like patch releases aren't really a priority, so that's not a strong use-case for this PR.

PR #847, opened 2023-12-21
doc: give hint regarding getting meeting link from latest commit

bzm3r:
When joining the meeting today, I used an incorrect link, as my browser served a GitHub page pointing to an older version of `maintainers/README.md`. 

For those who might be dealing with this issue in the future, including myself, I added a note+link regarding using the link from tip of master.


fricklerhandwerk:
Ah, is that because Discourse helpfully pins the link to master to a certain commit?

bzm3r:
Hmm, good question. I double checked, and it seems that Discourse, when it does have GitHub links, ends up pointing to master. So I think that's okay and it just might have been what was in my history for some reason... :shrug:

PR #848, opened 2023-12-25
updated channel version

amirreza8002:
not that important but could save some time for beginners who try to install new packages and fail (like me :p)


infinisil:
I hope somebody checked that the tutorial still works with the new channel? :)

henrik-ch:
> I hope somebody checked that the tutorial still works with the new channel? :)

If not earlier, yes it works :-).

PR #852, opened 2024-01-01
Adding a HowTo Dual boot guide

zolodev:
As this is my first contribution, I try to keep it simple in order learn the best practices to contribute to the project.
Unfortunately I could not find any templates for "How To" guides and tried to the best of my ability to add the information I thought could be good for new users like myself.

Please let me know if I can improve the documentation in any way.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -19,3 +19,5 @@
 
 /source/tutorials/integration-testing-using-virtual-machines.md @olafklingt
 /source/tutorials/nixos-configuration-on-vm.md @olafklingt
+
+/source/guides/recipes/dual-boot-NixOS.md @zolodev
--- a/source/guides/recipes/dual-boot-NixOS.md
+++ b/source/guides/recipes/dual-boot-NixOS.md
@@ -4,16 +4,16 @@ It is recommended to use Systemd boot over Grub bootloader.
 This how to is a short introduction to `systemd bootloaded` options. To get a brief understanding of how it works.
 
 ## Target audience
-A NixOS user that wants to make changes in boot, to set the `default` startup OS in Systemd boot.
+A NixOS user that wants to make changes in boot, like how to select a default OS in Systemd boot.
+Or just want to change the resolution of the menu to best suit the users needs.
 
 ## Purpose
-When dual booting, it can be useful to know how to work with systemd boot.
+When dual booting, it can be useful for the user to know how to work with systemd boot.
 
-## Steps
+## Options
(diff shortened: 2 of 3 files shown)


proofconstruction:
Hi, thanks for making a contribution!

From the description, it seems like this should include some information about common issues encountered when dual-booting NixOS, and ways to fix or get around those. Can you add something like that? Currently this is just information we could point to with a link to [the systemd-boot documentation](https://www.freedesktop.org/wiki/Software/systemd/systemd-boot/), but it'd be good to have NixOS-specific information here.

zolodev:
Thank you for the feedback. However I am no linux expert but I had problem finding this information in the documentation. 

It feels like a common issue on how to set the 'default' booting os. But finding that information is more difficult. 

I have not encountered any other common issues dual booting NixOS. But if you could point me in a direction of what information I can add, I will of course add and extend the HowTo with that information.😇

But most of the common dual boot information has already been documented elsewhere. But none mentions how to set the default booting os. That was why I thought it would be good to document it. But would it make more sense to rename the 'HowTo' to something else to clarify the intention?

And I was hoping someone could help me 😇 set the default booting os in configuration.nix file. But I do not know if that is possible. But if anyone knows how to do this, that would be something to document in this 'HowTo'.

infinisil:
I agree with @proofconstruction that this sort of information doesn't really fit here. But you're totally right that's it's hard to discover this.

Here's my suggestions: Let's link to the [upstream docs](https://www.freedesktop.org/wiki/Software/systemd/systemd-boot/) from the `boot.loader.systemd-boot.enable` options description. This is effectively the canonical place to look up documentation on specific parts of NixOS.

To do that, you'd make a PR to Nixpkgs and change this line here to include the link:

https://github.com/NixOS/nixpkgs/blob/5be2ce477a14eefe0ce37b3942c575ba80af1985/nixos/modules/system/boot/loader/systemd-boot/systemd-boot.nix#L84

> And I was hoping someone could help me 😇 set the default booting os in configuration.nix file. But I do not know if that is possible. But if anyone knows how to do this, that would be something to document in this 'HowTo'.

I don't know that myself unfortunately, and from taking a quick look at the code it doesn't seem doable at all without making some more involved patches to how NixOS works :upside_down_face:

zolodev:
Thanks @infinisil for your input. I will look into it and add the information as you suggest. ☺️

zolodev:
I have created a new PR as discussed. 
https://github.com/NixOS/nixpkgs/pull/282053

I will close this PR. 

Thanks @proofconstruction and @infinisil for the support! 👍

PR #854, opened 2024-01-05
remove no-op command in troubleshooting advice

fricklerhandwerk:
`nix-store --init` doesn't actually do anything [0] and is not even
documented in the Nix manual

[0]: https://github.com/NixOS/nix/blob/9651034dc26c009559cb571c813a2b8aa2bba3f1/src/nix-store/nix-store.cc#L736-L744

Note: I haven't tested the actual instructions, but this change at least doesn't make them more wrong

### DIFF
--- a/source/guides/troubleshooting.md
+++ b/source/guides/troubleshooting.md
@@ -36,13 +36,12 @@ This is a [known issue](https://github.com/NixOS/nix/issues/1251).
 
 It means that using a new version of Nix upgraded the SQLite schema of the [database](https://nix.dev/manual/nix/2.18/glossary#gloss-nix-database), and then you tried to use an older version Nix.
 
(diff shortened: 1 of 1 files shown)


proofconstruction:
Leading the way on "not more wrong" patches in 2024!

infinisil:
Did anybody actually test this though? In particular that FAQ entry talks about downgrading to older Nix versions, where it might have been required

fricklerhandwerk:
The last change to that was [18 years ago, for Nix 0.11](https://github.com/NixOS/nix/blame/9651034dc26c009559cb571c813a2b8aa2bba3f1/src/nix-store/nix-store.cc#L736-L744), so there's no risk any actual user would run into trouble.

infinisil:
Fair enough, but still, we should test changes to documentation, just like changes to code should also be tested

PR #857, opened 2024-01-06
global version for cross compilation tutorial nixos version

henrik-ch:
Inspired by #851

Solution found and adapted from https://github.com/sphinx-doc/sphinx/issues/4054 .

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -453,3 +453,18 @@ sitemap_url_scheme = "{link}"
 
 # Not found
 notfound_urls_prefix = "/"
+
+def ultimateReplace(app, docname, source):
+    result = source[0]
+    for key in app.config.ultimate_replacements:
+        result = result.replace(key, app.config.ultimate_replacements[key])
+    source[0] = result
+
+ultimate_replacements = {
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Hm. Not entirely sure whether this is a net improvement.

- Parameterising the version is probably good because we can do it wholesale
- Doing it only for one thing makes it inconsistent and therefore adds cognitive overhead
  - Doing it without explanation is particularly problematic and makes contributions substantially harder
- Doing it for everything and adding meta-documentation is quite a bit more work
- In any case we need more testing to be sure a change is sane. Changing stuff wholesale is bound to break things.

I'm somewhat inclined to keep it deliberately dumb and low-tech, to force going over examples for review on each change until we have automated testing.

henrik-ch:
> Hm. Not entirely sure whether this is a net improvement.

I wasn't super sure about this either; it seems to help on this piece of content, but also it makes it an outlier that works differently from everything else that we maintain.

infinisil:
I tend to agree with @fricklerhandwerk here. I'd probably prefer a linting-style solution that makes sure all tutorials consistently use the same channel.

PR #859, opened 2024-01-14
disable linkcheck for GitHub anchor links

fricklerhandwerk:
apparently it got (a lot) worse and GitHub doesn't render README files
statically any more.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,23 +99,3 @@ We would be glad to incorporate your insights.
 ## Contributor guides
 
 Please read [Contributing Documentation](https://nix.dev/contributing/documentation).
-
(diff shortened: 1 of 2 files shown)


infinisil:
Alternatively we could link to the README file specifically, like

> https://github.com/nix-community/nixos-generators/blob/master/README.md#cross-compiling

instead of

> https://github.com/nix-community/nixos-generators#cross-compiling

Though that's also just another workaround.

For now I think this PR is the best solution.

infinisil (approved):
I believe https://github.com/NixOS/nix.dev/blob/master/CONTRIBUTING.md#github-heading-anchors-fails-linkcheck isn't necessary anymore with this, so that can be removed. Otherwise this LGTM

PR #863, opened 2024-01-15
dynamically generate Nix manual version overview

fricklerhandwerk:
this removes more manuals steps from adapting the site to new releases.

@nixos/documentation-team please bikeshed whether to generate pages at all, as this introduces some mental overhead for working with the site.

depends on #830


DanielSidhion:
I looked at the changes and they seem ok, I don't know what you were referring to when you said "whether to generate pages at all" (what's the alternative? Not have any links to the nix manuals from nix.dev?).

However, I've been thinking. Is it possible to grab the nix version directly from each channel during CI? By that I mean, can we do a few "niv add <channel>" in CI and then for each channel, run `nix --version` and grab the version from there, so we (ideally) don't need to do any manual intervention at all?

default.nix
```diff
@@ -9,48 +9,80 @@ let
...
+      nix-releases = {
+        latest = "2.19";
+        rolling = "2.18";
+        stable = "2.18";
+        prev-stable = "2.13";
+      };
+      nixpkgs-releases = {
+        stable = "23.11";
+        prev-stable = "23.05";
+      };
```
infinisil, line 25: These here could be avoided by calling the `niv` entries `nix-latest`, `nix-rolling`, ..., `nixpkgs-stable`, `nixpkgs-prev-stable` instead. That way we never need to manually edit a file to update the versions, but can just do

```
niv update nix-latest --branch release-2.20`
```

infinisil (approved):
> I looked at the changes and they seem ok, I don't know what you were referring to when you said "whether to generate pages at all" (what's the alternative? Not have any links to the nix manuals from nix.dev?).

Emphasis is on _generate_. The alternative would be to not merge this PR, which then requires manually updating the Nix versions in the `.md` file whenever stable/unstable changes.

I think it's good to have this, automation is key!

default.nix
fricklerhandwerk: Reworked it. It's a bit less of a pain now, but still quite a pain.

default.nix
```diff
@@ -3,65 +3,128 @@
...
-          cp -Rf ${manual version}/share/doc/nix/manual/* $out/manual/nix/${version}
-        '';
-        # add upstream page redirects of the form `<from> <to> <status>`, excluding comment lines and empty
-        redirects = version: ''
-          sed '/^#/d;/^$/d;s#^\(.*\) \(.*\) #/manual/nix/${version}\1 /manual/nix/${version}\2 #g' ${src version}/doc/manual/_redirects >> $out/_redirects
+  nix-dev =
+    let
+      # Various versions of the Nix manuals, grep for (nix-manual)= to find where they are displayed.
+      # XXX: With the current setup, the Nix manual hosted on nix.dev does not get updated automatically with new releases.
+      # The following manual steps are required:
```
infinisil, line 17: This would be better in `CONTRIBUTING.md` to replace the section that was previously there.

infinisil (changes requested):
Oh just noticed a problem with this: If we update Nix versions, we'd be invalidating old version-specific links like `https://nix.dev/manual/nix/2.18/...`. So we should never do minor version updates for niv entries, but rather add new ones.

fricklerhandwerk:
We could also link to `stable` uniformly, or even expand such links to the actual version at build time. The rabbit hole of content generation is as deep as a Turing tar pit.

default.nix
fricklerhandwerk: Done

CONTRIBUTING.md
```diff
@@ -30,31 +30,29 @@ The following manual steps are required:
...
-     For example, to add Nix 2.19:
-
-     ```bash
-     niv add nixos/nix -n nix_2-19 -b 2.19-maintenance
-     ```
-
-  2. Reference the latest version in [`source/reference/nix-manual.md`](./source/reference/nix-manual.md).
+  ```bash
+  niv update nix-latest -b 2.20-maintenance
+  ```
```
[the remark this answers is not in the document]
infinisil: I'm afraid https://github.com/NixOS/nix.dev/pull/863#discussion_r1473411187 wasn't a good suggestion after all

fricklerhandwerk:
It works now and is even somewhat readable:

```
$ ls $(nix-build -A build --no-out-link)/manual/nix
2.10  2.11  2.12  2.13  2.14  2.15  2.16  2.17  2.18  2.19  2.20  2.21  2.4  2.5  2.6  2.7  2.8  2.9
```

Extra goodie: The Nix manual is again (optionally) available as single page (ping @ehmry)!

image

infinisil:
I'm afraid the code felt very cobbled together overall.. I kind of refactored the entire code now :sweat_smile:. Please check the commit message for more details, but this PR looks good to me now. I tested it locally with `netlify dev -d result`, including the redirects. I'm also pretty sure that all Nix versions are cached now, I don't have to build any anymore.

> Extra goodie: The Nix manual is again (optionally) available as single page (ping @ehmry)!

I reverted that for now, though I left it in the code: https://github.com/NixOS/nix.dev/blob/4daecbe3930134041d197e01509b52038692a8d2/default.nix#L54-L56

This feels way too hacky for me, it should be discussed a bit more. And there's no reason this has to be part of this PR, let's not make it bigger than it already is!

infinisil (approved):
Whoops fat fingers, though it looks like auto-merge doesn't do much when somebody still requested changes, I guess I'll have to approve :)

Issue #579, opened 2023-05-30
First steps tutorial - Nix language


zmitchell:
As part of the tutorial series being worked on by the Learning Journey WG we need a tutorial going over the Nix language. We recognize that there's an existing tutorial on nix.dev ([Nix language basics][lang_basics]). The existing guide is very thorough and we don't want to supplant that guide, but for the purposes of a quick introduction we want a short guide motivated by explaining a real piece of Nix code such as a functioning `shell.nix`.

[lang_basics]: https://nix.dev/tutorials/first-steps/nix-language

- [ ] Prepare an outline of the tutorial
- [ ] Prepare a draft of the tutorial
- [ ] Submit the draft as a PR

## Tutorial contents
- Provide the reader with a functioning `shell.nix` that builds a Python environment
- This example should have enough complexity to illustrate basic language features:
    - Function definitions (the whole file is a function definition)
    - Function application
    - Passing anonymous functions as arguments via `python.withPackages`
    - Variables, lists, attribute sets, etc
    - Navigating attribute sets via the dot operator
    - etc

roberth:
> for the purposes of a quick introduction we want a short guide motivated by explaining a real piece of Nix code such as a functioning `shell.nix`.

We have another tutorial for that, and editing a file isn't particularly inviting. I think in this tutorial we should teach readers how to learn the language and the library. For this I think the `nix repl` is a great vehicle.

To make this nicely hands-on, we could start with

`nix repl <nixpkgs>`

and then introduce concepts in a way they can follow along.

This way we can both teach the fundamental principles, and teach them how to experimentally write solutions for their configuration needs.

 - language basics
     - bindings in the repl vs `let`
          - start with this. The repl is the first instance of a scope they'll see, and we can take it from there
     - (as listed in the issue description)

 - discovering functions
     - `lib.strings.<TAB>`, etc
     - noogle

 - finding documentation
     - Nixpkgs manual
     - doc via source with ctrl+click in IDE terminal (e.g. vscode)
        ```
        nix-repl> lib.fix
        «lambda @ /home/user/h/nixpkgs/lib/fixed-points.nix:19:9»
        ```
        ctrl+click the file:line string
     - `:doc` for primops, or maybe that's redundant if we also mention...
     - noogle

 - generate a config file
     - `pkgs.formats`
     - almost an end goal? While they could use the generated file from the CLI, they'll want to use it in e.g. a NixOS module instead.

fricklerhandwerk:
[...]

PR #867, opened 2024-01-21
Elaborate on `let` bindings

rayhem:
The original documentation says let bindings are for "repeated use", though one can run

```nix
a = 3;
a + a
```

just fine without the binding. I think it's more appropriate to explain that `let` limits how far a value's name can reach. This is mentioned later, though, so I'm kind of puzzled on how to improve the language here and make it obvious *why* a let binding is a useful construction right up front.

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -388,24 +388,6 @@ in
 2
 ```
 
-Attempting to use the value outside of the binding produces an error:
-
-```{code-block} nix
-:class: expression
-a
-```
-
-```{code-block}
-:class: value
-error: undefined variable 'a'
-
-       at «string»:1:1:
-
-            1| a
-             | ^
-```
-
-
 :::{dropdown} Detailed explanation
 
 Assignments are placed between the keywords `let` and `in`.
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -371,7 +371,7 @@ error: undefined variable 'one'
 
 Also known as “`let` expression” or “`let` binding”
 
-`let` expressions allow assigning names to values for scoped use.
+`let` expressions allow assigning names to values for repeated and scoped use.
 
 Example:
 
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -371,7 +371,7 @@ error: undefined variable 'one'
 
 Also known as “`let` expression” or “`let` binding”
 
-`let` expressions allow assigning names to values for repeated use.
+`let` expressions allow assigning names to values for scoped use.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


source/tutorials/nix-language.md
```diff
@@ -380,14 +380,32 @@ Example:
...
+:class: value
+error: undefined variable 'a'
+
+       at «string»:1:1:
+
+            1| a
+             | ^
+```
+
+
```
henrik-ch, line 408: I feel that this section is a bit redundant, as it's only showing that we can't refer to a variable that hasn't been defined.

source/tutorials/nix-language.md
```diff
@@ -371,7 +371,7 @@ error: undefined variable 'one'
 
 Also known as “`let` expression” or “`let` binding”
 
-`let` expressions allow assigning names to values for repeated use.
+`let` expressions allow assigning names to values for scoped use.
```
henrik-ch, line 374:
```suggestion
`let` expressions allow assigning names to values for repeated and scoped use.
```

henrik-ch (commented):
Welcome to nix.dev, @rayhem! It's great to see new users and contributors joining our community.

Thank you for your suggestions regarding the language tutorial. This section of the site has been crafted with a lot of care, and we're always mindful about maintaining its conciseness to respect our readers' time and attention. While we might not make significant changes to this part right now, your input is still valued and definitely contributes to our ongoing discussions.

Please continue to explore nix.dev and don't hesitate to share more issues and PRs. Your perspective is important to us, and there's a lot of room for improvement and new ideas.  Your engagement is what helps our community grow and thrive. Looking forward to seeing more of your contributions!

source/tutorials/nix-language.md
rayhem: I agree. Do you think it would be appropriate to add commentary along the lines of "`let` is similar to curly brace scoping in other languages" with information on the similarities and differences? Or is that too much for this section of the tutorial?

infinisil:
> The original documentation says let bindings are for "repeated use", though one can run
> 
> ```nix
> a = 3;
> a + a
> ```
> 
> just fine without the binding.

That only works in the nix repl (although only if you remove the `;`), it's not part of the Nix language per se! Underneath, nix repl roughly translates those statements into a Nix expression like this:

```nix
let
  a = 3;
in
a + a
```

fricklerhandwerk:
@rayhem Sorry for chiming in late. Indeed I think it would be too much to go into that sort of depth here. The tutorial is supposed to be a first, deliberately superficial introduction to the most widely used patterns. And yet some readers report it hits them like a wall of text. So while there needs to be a place for these different angles on the language, it would probably need to be a separate tutorial: https://github.com/NixOS/nix.dev/issues/579 (And there probably needs to be an even more condensed first glance at the language.)

Since the change now seems to have boiled down to almost nothing, I'll close this PR. This is not to dismiss your work! Feel free to add your feedback to other issues as we compile ideas for further tutorials and guides.

PR #870, opened 2024-01-26
Fix unquoted URL in "Nix language basics"

MasonM:
I'm a newbie, so I might be missing something, but the use of an unquoted URL here seems like it goes against the ["Best Practice" page](https://nix.dev/guides/best-practices#urls), which says to always quote URLs.


henrik-ch (approved):
LGTM

PR #871, opened 2024-01-26
remove email subscription form

fricklerhandwerk:
historically it still points to Cachix and not to a community-owned instance.
at the moment there is no regular community newsletter, which would certainly include updates about documentation.
we could add it back once such a thing exists, but probably it's better to manage that in one place.

@nixos/marketing-team

Related: https://github.com/NixOS/nixos-weekly/issues/170


henrik-ch (approved):
Checked and LGTM

PR #872, opened 2024-01-26
front page: Who is Nix for?

fricklerhandwerk:
For at least two years there were ongoing debates about the target
audience of the Nix ecosystem. [Nix maintainers did not manage to
converge on a statement][0]. The marketing team is explicitly [focusing
on software developers][1], and the documentation team was supposed to
align with that [from the very beginning][2].

What was missing so far was what we mean by "software developers".
Arguably, it encompasses a particular mind set about how to deal with
computers and enough time to learn things.

And while for documentation it matters most *what* the software is doing
rather than *who* it is made for (or by), the ecosystem is a community
effort, and [people matter][3]. So far, we haven't really identified the
(eventual) boundaries of this community (at least I don't know of any
serious attempt), which also plays into the definition of a target
audience. Yet, in practice, not every contribution, not every question
or comment is treated equally, and this has reasons in our implicit
assumptions about who belongs.

This is an attempt to draw such a boundary that is not arbitrary, and
neither to narrow yet
Projectsnor too wide. The goal is, as always, to help
Nix beginners set realistic expectations for their journey.

The list of occupations/interests who may benefit from Nix is based on
the [2022][4] and [2023][5] community surveys.

[0]: https://github.com/NixOS/nix/pull/7156
[1]: https://github.com/NixOS/nixos-homepage/blob/76d42a052dce7917007b5553dd08ea2569d04bc7/community/teams/marketing.tt#L89
[2]: https://discourse.nixos.org/t/2022-06-15-documentation-team-meeting-notes-1/20004
[3]: https://discourse.nixos.org/t/zurich-23-05-zhf-hackathon-and-workshop-report/29093#ux-workshop-21
[5]: https://discourse.nixos.org/t/nix-community-survey-2023-results/33124
[4]: https://discourse.nixos.org/t/2022-nix-survey-results/18983

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -70,7 +70,7 @@ The following illustrate of what can be achieved with the Nix ecosystem:
 
 ## Who is Nix for?
 
-Nix is a tool for software developers.
+Nix is primarily a tool for software developers.
 
 You don't have to be a professional and you don't need formal education in informatics to greatly benefit from Nix.
 For our purposes here, you are a software developer if you manipulate character strings to make computers do exactly what you want.
--- a/source/index.md
+++ b/source/index.md
@@ -68,6 +68,28 @@ The following illustrate of what can be achieved with the Nix ecosystem:
 - Atomic upgrades and rollbacks.
 
 
+## Who is Nix for?
+
+Nix is a tool for software developers.
+
+You don't have to be a professional and you don't need formal education in informatics to greatly benefit from Nix.
+For our purposes here, you are a software developer if you manipulate character strings to make computers do exactly what you want.
+
+However, experience with complex software projects helps understanding why Nix is useful, knowing some informatics helps appreciating how it works, and either one is a good starting point to [help improve it](how-to-contribute).
+
+You probably won't want to go back to a world without Nix if you're a:
+
+- Full-stack or back-end developer
+- Test engineer
+- Embedded systems developer
+- DevOps engineer
+- System administrator
+- Data scientist
+- Natural scientist
+- Student of a technical field
+- Open source software enthusiast
+
+
 ```{toctree}
 :hidden:
 
--- a/source/index.md
+++ b/source/index.md
@@ -70,12 +70,11 @@ The following illustrate of what can be achieved with the Nix ecosystem:
 
 ## Who is Nix for?
 
-Nix is primarily a tool for software developers.
+Nix is a tool for people who both need computers to do exactly as intended, repeatably, far into the future, and who are familiar with command line interfaces and plain text editors.
 
-You don't have to be a professional and you don't need formal education in informatics to greatly benefit from Nix.
-For our purposes here, you are a software developer if you manipulate character strings to make computers do exactly what you want.
-
-However, experience with complex software projects helps understanding why Nix is useful, knowing some informatics helps appreciating how it works, and either one is a good starting point to [help improve it](how-to-contribute).
+You don't have to be a professional software developer and you don't need formal education in informatics to greatly benefit from Nix.
+However, experience with complex software projects and knowing some informatics helps with appreciating why it's useful and how it works.
+And it helps with learning how to use it effectively and [how to make improvements](how-to-contribute).
 
 You probably won't want to go back to a world without Nix if you're a:
 
--- a/source/contributing/how-to-contribute.md
+++ b/source/contributing/how-to-contribute.md
@@ -1,3 +1,4 @@
+(how-to-contribute)=
 # How to contribute
 
 The Nix ecosystem is developed by many volunteers and a few paid developers, maintaining one of the largest open source software distributions in the world.


source/index.md
```diff
@@ -68,6 +68,28 @@ The following illustrate of what can be achieved with the Nix ecosystem:
 - Atomic upgrades and rollbacks.
 
 
+## Who is Nix for?
+
+Nix is a tool for software developers.
```
DanielSidhion, line 73:
```suggestion
Nix is primarily a tool for software developers.
```

I think adding "primarily" helps keep the door open to different audiences who may be willing enough to learn and use Nix, while not distracting from the focus on software developers.

source/index.md
```diff
@@ -68,6 +68,28 @@ The following illustrate of what can be achieved with the Nix ecosystem:
...
+
+You don't have to be a professional and you don't need formal education in informatics to greatly benefit from Nix.
+For our purposes here, you are a software developer if you manipulate character strings to make computers do exactly what you want.
```
DanielSidhion, line 76: I think this sentence doesn't help narrow down the audience.

I'm coming from this with mostly an assumption (roughly reinforced by answers in the Nix community surveys): that there are primarily two types of audience that come to Nix/NixOS. The first audience is people who actually write code and build software (I'm not using the term "software developer" here to avoid confusion with the current version of the PR). The second audience is people who are Linux enthusiasts - they use a Linux OS, know how to perform certain tasks on the terminal and are usually a bit more tech-savvy, but they are "script kiddies" at best when it comes to actual understanding about how software works.

With this assumption in mind, "manipulate character strings to make computers do exactly what you want" can be applied to both types of audience, but as I interpreted it from the other sources you listed, the focus is only on the first type of audience, not the second.

```suggestion
For our purposes here, you are a software developer if you write or build code for computers.
```

I feel this suggestion is excluding some part of the audience that is intended to be included, but I am not sure about that. I made this suggestion to start from the other end: very narrow, and hoping that we can identify other audiences that we want to include and keep adjusting the sentence.

source/index.md
```diff
@@ -68,6 +68,28 @@ The following illustrate of what can be achieved with the Nix ecosystem:
 - Atomic upgrades and rollbacks.
 
 
+## Who is Nix for?
+
+Nix is a tool for software developers.
+
+You don't have to be a professional and you don't need formal education in informatics to greatly benefit from Nix.
```
DanielSidhion, line 75: This sentence should come after the next one: first clarify what we mean by software developer, then make this statement that you don't have to be a professional and so on...

DanielSidhion (commented):
Thanks for this PR! This audience focus is something that has been in my mind for a few weeks now, and I wrote in one of the comments my main assumption of what the two current types of audience Nix has. While I'd like to eventually be able to cover at least both kinds of audience, I agree that focusing on software developers is a good initial focus.

btw - you may want to update the link to the 2022 nix survey, because it's pointing to something else I think. I found this link to be more helpful: https://discourse.nixos.org/t/2022-nix-survey-results/18983

source/index.md
fricklerhandwerk: Yeah, I was shuffling around those sentences until they stopped making sense to me. You're probably right.

source/index.md
fricklerhandwerk: Fully agreed that there's a producer <-> consumer spectrum, but I really didn't want to go down that road here. 

My way of phrasing it is certainly peculiar. It's more to accommodate a sentiment expressed in https://github.com/NixOS/nixos-homepage/issues/828: I want the community to be welcoming to nerds of all sorts, and especially beginners. If you had asked me 10 years ago, I would not have identified with your definition. Beginners don't *write* code. They use it, they configure existing software (manipulating what isn't or what we usually don't consider code), sometimes read code, and then maybe start changing that code. The cool thing about Nix is that, while certainly you'll end up writing code, in the beginning it doesn't really feel like it.

The other thing I was trying to capture is that what you really need is a particular mind set (and a lot of time to learn things) rather than a particular occupation (which "writing code" is) or qualification. We eventually meet at what ends up being programming. It's just that at the boundaries there are enough people who don't actually want to program - but still do because they have to deal with the fact that software sucks or doesn't do what they want. Natural scientists are a good example. All the anecdotes I know  amount to "we write code [that breaks] all the time but would rather do `$SCIENCE`".

All this clearly excludes people who won't use the shell or a plain text editor.

That was my take at "not too wide, not too narrow".

infinisil: > All this clearly excludes people who won't use a command line interface or a plain text editor.

This sounds much more straightforward, how about just mentioning this instead?

fricklerhandwerk:
```suggestion
For our purposes here, you are a software developer if you use a command line interface and a plain text editor.
```

You mean like this?

source/index.md
```diff
@@ -68,6 +68,28 @@ The following illustrate of what can be achieved with the Nix ecosystem:
 - Atomic upgrades and rollbacks.
 
 
+## Who is Nix for?
+
+Nix is primarily a tool for software developers.
```
infinisil, line 73: Shouldn't this be a PR to nixos.org instead? nix.dev is for learning about Nix, but this section doesn't teach anything, it's just marketing. If somebody went onto nix.dev, we should assume that they want to learn about Nix, so I don't think this is necessary or useful here.

And if your intention is to make Nix platform _contributors_ be more focused on improving the developer story, I also don't think this would achieve that. Nix is a community of various individuals and companies, which will do whatever they're most interested in (which often isn't focused on the developer story).

source/index.md
DanielSidhion: I see what you were going for now. From the PR description, I thought you wanted to closely align with the marketing team's audience focus (which seems to be closer to the definition I suggested).

If the idea is to keep it broader, why not remove the term "software developer" completely? It's bound to cause confusion (probably similar to mine).

My suggestion would be to shorten this to something like: "Nix is a tool for people comfortable with a command line interface and a plain text editor."

source/index.md
fricklerhandwerk: > Shouldn't this be a PR to nixos.org instead?

In an ideal world, maybe? In practice, I find that code unmaintainable and refuse to work with it.

> nix.dev is for learning about Nix, but this section doesn't teach anything, it's just marketing.

This is very clearly not marketing, because it doesn't advertise anything, it doesn't try to convince anyone of anything. The purpose is, as said in the commit message, to set expectations and prepare people for what they'll deal with. The phrase "going back to a world without Nix" is only a measure to avoid sounding academic. The technically correct statement would be "According to community surveys 2022 and 2023, people who identify with these descriptions very probably find Nix useful".

> And if your intention is to make Nix platform contributors be more focused on improving the developer story, I also don't think this would achieve that.

While that's not my intention, I strongly doubt it's true what you say. [The first impression matters a lot](https://en.wikipedia.org/wiki/Priming_(psychology)), and even if this particular piece is below the fold, whoever cares enough to read it will carry that message around with them for a long time to shape their thinking.

infinisil: > This is very clearly not marketing

Fair yeah, but still, setting expectations seems like a main landing-page kind of deal. And you refusing to touch nixos.org is clearly something gone wrong that we should prioritise addressing instead. This is definitely worth a @NixOS/marketing-team ping.

infinisil: I guess it shouldn't hurt to _also_ set expectations for nix.dev specifically, but so far this PR seems to set expectations for Nix in general.

source/index.md
fricklerhandwerk: > Nix is a tool for people comfortable with a command line interface and a plain text editor.

Hm, I agree it's more straightforward and descriptive. But also completely lacks the part that I find most important, and that's about the spirit of it all, the "why does it matter". Why would you bother going through pages of dry text for hours on end? Your suggestion makes it sound like this is yet another command line tool. I'm getting flashbacks from the vision statement discussion...

Okay, which group of people would care about the following?

- Makes computers *actually* do (well, almost) exactly the same thing every time (it's cached `execve()` in a sandbox on immutable files)
- Allows expressing process execution and file system operations in terms of declarative programs
- Exposes and allows managing all the complexity of the open source software ecosystem

If not forced by necessity, probably it's perfectionist control-freaks. Maybe that's the uncomfortable truth that prevents us from saying it out loud. If there's a constructive variant of it, I'd still take that.

source/index.md
fricklerhandwerk: I'm being deliberately generic here, because we don't have that sort of writing anywhere. I'd rather have it here now and move it around later. I've been repeating myself and trying to come up with things that help orient people for almost two years now, getting bogged down with concerns on every step. I'm all in favor of being considerate with far-reaching decisions, but *come on!*

fricklerhandwerk:
Closed all comments after re-thinking the phrasing. Thanks @DanielSidhion for the first feedback round.

DanielSidhion (approved):

source/index.md
DanielSidhion: I'm all in favor of getting more content written as well! For now, I'd rather have redundant information in different places than try to keep it in a single place (as long as that information won't go out of date soon). It's been my experience that people find nix through multiple different links, and often don't visit all the other links (e.g. they find combinations of the nix pills, manuals, nix.dev, wiki, and multiple other sites, but often don't find all of those), so having repeating information (which doesn't go out of date soon) is better IMO at the moment.

Properly organising all the content under a single structure takes way more time (but we're moving there, e.g. with the nix.dev -> docs.nixos.org rename and so on), and when we get there we can start cutting the redundant content.

My only concern is making sure what we're adding is:
1. Not confusing for the intended audience
2. Improving on the current content

I think the latest version of this PR fits both criteria, so I'm approving!

fricklerhandwerk:
Updated the links - there were typos in the commit message/PR description.

PR #876, opened 2024-01-27
update information flow diagram

fricklerhandwerk:
we probably don't need it here in particular, but originally when the documentation team was new, it was useful to understand and communicate what's going on. it can probably stay here until we either find a better place or feel like it's in the way.

until then, a small update for clarity and consistency.


maintainers/information-flow.md
```diff
@@ -0,0 +1,53 @@
...
+flowchart LR
+  X --> |A| Y
+```
+
+Information resources – communication platforms – are distinguished by mode of interaction:
+
+*Streams* can be appended to:
+
+- The amount of information in them grows without bounds.
+- Only recent items are relevant.
```
DanielSidhion, line 20: Minor thing: is this true for RFCs as well? They're categorised as a stream.
I think RFCs should be considered artifacts. I'm interpreting RFCs as each individual RFC instead of the "RFC repository". In that sense, they're more like artifacts.

DanielSidhion (approved):

maintainers/information-flow.md
fricklerhandwerk: I bounced back and forth. I'm inclined to consider RFCs throwaway, because ideally (and this is just parroting comments from other people) the knowledge gets merged into living code and documentation. That they aren't in practice seems rather like a slight dysfunction of processes.

infinisil (approved):
Pushed to fix CI (trailing newline). While it's not perfect, I don't think we can make a perfect diagram, so let's just merge this for now, we can improve it afterwards if necessary.

PR #879, opened 2024-01-28
move shell.nix tutorial up

fricklerhandwerk:
it doesn't make sense that we continue with reproducibility becuase we
haven't seen lookup paths yet.


infinisil (approved):
After a close look, this makes sense!

PR #881, opened 2024-01-28
move automatic environments up, add next steps

fricklerhandwerk:
automatically entering the shell is the next logical step, especially if
we have turned it on and off multiple times in a row.


infinisil (approved):
LGTM!

PR #888, opened 2024-01-29
Update packaging-existing-software.md

nbelakovski:
As it was written, I thought we'd have to do more work to supply fetchzip, but the next invokation of nix-build seems to call fetchzip without any definition errors so I assume the error is in the documentation.


fricklerhandwerk:
Yes, thanks!

PR #901, opened 2024-02-06
declarative shells tutorial: more generic example commands

### LINKED ISSUE #892: Git shell hook breaks if not a Git repo
The guidance says users should:

```
Exit the shell by typing exit or pressing Ctrl+D, then start it again with nix-shell to observe the effect.
```

But the shell hook `git status` causes an error:

```
Admin:~/environment/nix $ nix-shell
fatal: not a git repository (or any of the parent directories): .git
```

The guidance should either include a friendly shell hook (e.g. the `cowsay` example from the first tutorial), or should explain to the user the expected message.

fricklerhandwerk:
fixes #892

@aynsof thanks for reporting!

@nixos/documentation-team what do you think? We discussed a long time ago https://github.com/NixOS/nix.dev/issues/498#issuecomment-1507689468 to use developer-oriented tools in the examples, but this seems to keep failing due to ever so slightly elevated requirements for them to actually work.
I'm now inclined to keep examples deliberately toy-like to avoid obscuring the actual thing we want to teach.
Admittedly, this particular change adds lots of character noise, but I couldn't come up with something more elegant (brief but powerful) in the short amount of time I gave it.
Maybe someone has a better suggestion?

### DIFF
--- a/source/tutorials/first-steps/declarative-shell.md
+++ b/source/tutorials/first-steps/declarative-shell.md
@@ -98,7 +98,7 @@ Packages defined in the `packages` attribute will be available in `$PATH`.
 
 You may want to automatically export certain environment variables when you enter a shell environment.
 
-Set `MANPATH` and `MANPAGER` such that `man cowsay` in the shell environment is displayed with colors.
+Set `GREETING` so it can be used in the shell environment:
 
 ```diff
  let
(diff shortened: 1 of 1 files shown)


henrik-ch (approved):
👍 This works as expected as far as I can see, and I consider it an improvement as it's simpler and requires less unix cli knowledge. (Beginners may not know what is part of nix and not or struggle with the overhead of learning about man pages at the same time as they learn nix)

aynsof:
@fricklerhandwerk thank you, this is so much clearer for newcomers!

PR #902, opened 2024-02-06
use `nixShellNoCC` in tutorials and guides

fricklerhandwerk:
in our examples we never actually use the compiler toolchain stuff from
`mkShell`, so we may as well not burden users with large downloads.

It's not as slick, but arguably better practice. @infinisil I've seen you using it that way, what do you think?

Thinking further, maybe we could add (and backport) an alias to it in Nixpkgs that has a more precise name that will be easier to phase out of in a better world where Nix allows actual environments (or entire shells) and Nixpkgs provides them?

### DIFF
--- a/source/guides/recipes/dependency-management.md
+++ b/source/guides/recipes/dependency-management.md
@@ -38,7 +38,7 @@ Add niv to the development environment for your project to have it readily avail
    build = pkgs.hello;
  in {
    inherit build;
-+  shell = pkgs.mkShell {
++  shell = pkgs.mkShellNoCC {
 +    inputsFrom = [ build ];
 +    packages = with pkgs; [
 +      niv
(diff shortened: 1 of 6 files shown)


henrik-ch:
Going through the dependency management tutorial, it fails in this step:

[nix-shell:/niv-tutorial/old]# nix-build
error: attribute 'mkShellNoCC' missing

       at /niv-tutorial/default.nix:7:11:

            6|   inherit build;
            7|   shell = pkgs.mkShellNoCC {
             |           ^
            8|     inputsFrom = [ build ];

I didn't look into this any further, but I assume that the mkShellNoCC is a more recent addition to nixpkgs.

infinisil (approved):
Sounds fine to me!

Issue #553, opened 2023-05-12
Deploy site previews for PRs


zmitchell:
**Observations**

PRs often change the layout, presentation, or content of the site.

**Problem**

It is much easier to review PRs when you can visually inspect the result of a change.

**Approaches**

- One approach is requiring the PR author to also attach a screenshot in the PR body, but this is a burden and is likely to be forgotten as it's a nonstandard PR workflow and can't be automated easily.
- Since we're using Cloudflare Pages, which has a Preview Deployments feature, we can render a preview deployment on each push to a PR branch.
    - Publishing the Preview URL to a comment on the PR likely requires editing the GitHub Actions for the site.
    - More info:
        - [Cloudflare Preview Deployments](https://developers.cloudflare.com/pages/platform/preview-deployments/)
        - [Cloudflare Pages Action](https://github.com/cloudflare/pages-action)

**Will the contributor need help?**
Yes. This would require making changes to the Cloudflare Pages account, which very few people have access to. The contributor working on this could submit the changes to the GitHub Actions, then ping another maintainer when it's time to edit the Cloudflare Pages account.

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

zmitchell:
Ha, apparently this has been enabled very recently.

henrik-ch:
Work to enable site previews is under way with #479 .

fricklerhandwerk:
@asymmetric @yukiisbored, after @delroth moved nix.dev to Netlify the previews of PRs for external branches ceased working again. That was actually very useful. Can one of you please dig into it to get it back for us? That would be absolutely awesome.

delroth:
Hmm, it worked on my fork when I tested it originally... It looks like some issue with GH actions and permissions, when running in a PR context the action is not finding the deployment token and skips the deployment.

I'm not sure when I'll have time to debug this - I don't think the issue is specific to Netlify or needs any specific access/permissions to fix, so if anyone happens to know about GH actions and wants to have a look this would be welcome. Otherwise I'll add this to my TODO list, but I'm pretty busy in the next 7-14 days.

(I think the original setup with Cloudflare side-stepped that problem with a separate deployment workflow. This could likely be re-introduced here? Not sure.)

PR #904, opened 2024-02-06
actually fix mutable shortlinks

fricklerhandwerk:
now tested with:

    nix build && nix-shell -p netlify-cli --run "cd result; netlify dev"


infinisil:
If you create PRs from branches in this repository (instead of your fork), you get the auto-deployments and don't have to test locally :)

fricklerhandwerk:
Would be great to have deployments from forks as  well: https://github.com/NixOS/nix.dev/issues/553

And also, doing things locally is faster and better.

PR #912, opened 2024-02-08
use a text-only configuration for the VM tutorial

fricklerhandwerk:
@nh2 @olafklingt follow-up on #768

with the right flags it can be started in the current terminal.
when the VM stops, it's ungarbled with `reset`.

found this when trying to test a VM configuration without all the graphical hassle:
https://nixos.wiki/wiki/Cheatsheet#Building_a_service_as_a_VM_.28for_testing.29


nh2:
Ah, very good to know!

olafklingt:
Yes i think this makes sense. Thank you!

olafklingt (approved):

PR #914, opened 2024-02-11
Replace mobile Wikipedia links with desktop ones

BOHverkill:
Wikipedia automatically redirects mobile users to the mobile version when clicking a desktop link, but doesn't do the same the other way around. This also introduces consistency, as some Wikipedia links use the desktop version and some the mobile version.


henrik-ch:
Thank you for paying attention and cleaning this up.

PR #921, opened 2024-02-23
Fixing typo in `integration-testing-using-virtual-machines` and `packaging-existing-software`

ibleedicare:
Hello !

This PR fix two typo I found in the documentation

PR #925, opened 2024-02-26
Error in git command to copy

Talleyrand-34:
The command was git "glone" instead of git clone


fricklerhandwerk:
Thanks, fixed in https://github.com/NixOS/nix.dev/pull/921 already!

PR #935, opened 2024-03-08
migrate and update `callPackage` tutorial

### LINKED ISSUE #651: Package parameters and overrides (with `callPackage`)
**Proposal**

`callPackage` is used everywhere in Nixpkgs, but it's quite magical. It's very easy to use, so there are few instructions needed (rather a few good examples highlighting patterns *where* it makes particular sense to use it), but it requires understanding the rationale and inner workings.

**Proposal**

Import https://summer.nixos.org/blog/callpackage-a-tool-for-the-lazy/ (source: https://github.com/NixOS/nixos-summer/blob/main/content/blog/callpackage-a-tool-for-the-lazy.md) to nix.dev, and expand it to provide the motivation for why it exists at the beginning, and possibly some more details on `newScope` and friends at the end.

**Willing to help?**

Can help with reviews.

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

fricklerhandwerk:
closes #651

@NobbZ

### DIFF
--- a/source/tutorials/callpackage.md
+++ b/source/tutorials/callpackage.md
@@ -1,0 +1,224 @@
+---
+title: callPackage, a tool for the lazy
+date: 2022-09-08
+authors:
+  - Norbert Melzer
+  - Valentin Gagarin
+  - Matthias Meschede
+---
+
+The Nix package manager ships with a a special-purpose programming language for creating packages and configurations: the Nix language.
+It is used to build the Nix package collection, known as [`nixpkgs`](https://github.com/nixos/nixpkgs) – the largest, most up-to-date open source software distribution in the world.
+
+Being purely functional, the Nix language allows declaring custom functions to abstract over common patterns.
+One such pattern is parametrization of package definitions, that is, builds which can vary by their dependencies or settings.
+
+`nixpkgs` is a sizeable software project on it's own, with coding conventions and idioms that have emerged over the years.
+It has [established a convention](https://github.com/NixOS/nixpkgs/pull/9869) of composing parameterized packages with automatic settings through a function named [`callPackage`](https://github.com/NixOS/nixpkgs/commit/fd268b4852d39c18e604c584dd49a611dc795a9b).
+
+This article shows how to use it and why it's beneficial.
+
+## Basic examples
+
+Before even discussing the benefits, let's see how it actually gets used.
+
+Given are the files `hello.nix` and `default.nix`:
+
+```nix
+# default.nix
+let pkgs = import <nixpkgs> {}; in
+pkgs.callPackage ./hello.nix {}
+```
+
+`default.nix` produces a *derivation* from the contents of `hello.nix`.
+A derivation is what the Nix package manager calls a description of how to obtain a desired build result. Build results are often packaged executables, but can be arbitrary files. Derivation is also the data type of the corresponding expression in the Nix language.
+
+```nix
+# hello.nix
+{ writeShellScriptBin }:
+writeShellScriptBin "hello" ''
+  echo "hello, world!"
+''
+```
+
+`hello.nix` declares a function which takes as argument an attribute set with one element `writeShellScriptBin`. `writeShellScriptBin` is also a function, which happens to return a derivation. The build result in this case is an executable shell script with the contents `echo "hello world"` named `"hello"`.
+
+Building `default.nix` with `nix-build` produces the build result `./result/bin/hello`, and running this script will nicely greet you.
+
+As you can see, the argument `writeShellScriptBin` gets filled in automatically when the function in `hello.nix` is evaluated.
+Explaining in detail how this happens is not in the scope of this blogpost.
+This automatic filling of attributes is what `callPackage` is responsible for.
+It passes attributes that exist in the `pkgs` attribute set to the called function, simply matching by name.
+
+It may appear cumbersome to create the extra file `hello.nix` for the package in such a simple setup.
+We have done so because this is exactly how [`nixpkgs`](https://github.com/nixos/nixpkgs) is organized: every package is a file that declares a function.
+This function takes as arguments the package's dependencies.
+
+If you continue reading, you will see the benefits of this pattern!
+
+## 1. Benefit: parametrized builds
+
+Let's change the `default.nix`.
+
+Now it does not produce a single derivation any more, but an attribute set with the attribute `hello` containing the original derivation:
+
+```nix
+# default.nix
+let pkgs = import <nixpkgs> { }; in
+{
+  hello = pkgs.callPackage ./hello.nix { };
+}
+```
+
+When we build it with `nix-build -A hello` (accessing the attribute `hello` with the `-A` flag), the outcome will be the same as before.
+
+We also change `hello.nix` to add an additional parameter `audience` with default value `"world"`:
+
+```nix
+# hello.nix
+{ writeShellScriptBin
+, audience ? "world"
+}:
+writeShellScriptBin "hello" ''
+  echo "hello, ${audience}!"
+''
+```
+
+Building this will still yield the same output as before.
+
+Things get more interesting when we alter `default.nix` another time to make use of this new argument.
+Let's call `hello.nix` with the `audience` attribute set to `people`.
+
+Note how we pass the parameter `audience` in the second argument to `callPackage`, which is passed on to the function defined in `hello.nix`:
+
+```nix
+# default.nix
+let pkgs = import <nixpkgs> { }; in
+{
+  hello = pkgs.callPackage ./hello.nix { };
+  people = pkgs.callPackage ./hello.nix { audience = "people"; };
+}
+```
+
+Building via `nix-build -A people` will now yield a script that prints `hello,
+people`.
+
+We could use the very same syntax to also overwrite the automatically discovered
+arguments like `writeShellScriptBin`, though that doesn't make sense here.
+
+This pattern is used widely in `nixpkgs`:
+For example, functions which represent Go programs often have a parameter `buildGoModule`, and it is common to see an expression like `callPackage ./go-program.nix { buildGoModule = buildGo116Module; }` to enforce a certain Go compiler version.
+`nixpkgs` is therefore not simply a huge library of pre-configured packages, but a collection of functions that can be used to generate custom configurations of packages and even of whole ecosystems (e.g. "All Python packages using my custom interpreter") on the fly without duplicating code.
+
+## 2. Benefit: overrides
+
+`callPackage` adds more convenience by adding an attribute to the derivation it returns: the `override` function.
+
+This means that, as a consequence of handling builds with `callPackage`, we can also change the value of these function arguments _after_ the fact, using the derivation's `override` function.
+
+Consider this new `default.nix`, where we added a third attribute `folks`:
+
+```nix
+# default.nix
+let pkgs = import <nixpkgs> { }; in
+rec {
+  hello = pkgs.callPackage ./hello.nix { };
+  people = pkgs.callPackage ./hello.nix { audience = "people"; };
+  folks = hello.override { audience = "folks"; };
+}
+```
+
+Note that the resulting attribute set is now recursive (by the keyword `rec`), that is, attribute values can refer to names from within the same attribute.
+
+Here we take the `hello` derivation and call its `override` attribute as a function, passing the attribute set `{  audience = "folks"; }`. `override` passes `audience` to the original function in `hello.nix` - to be precise, *overrides* whatever arguments have been passed in the original `callPackage` that produced the derivation `hello`.
+
+Building and running the `folks` attribute with `nix-build -A folks` will again produce a new version of the script.
+It will print, as you may expect, `hello folks`.
+
+All the other parameters will remain the same as they have been when `hello` was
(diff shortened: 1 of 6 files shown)


NobbZ (approved):
No strict blockers on my opinion only a handfull of suggestions, most related to changes that appear out of scope for this PR.

source/tutorials/packaging-existing-software.md
```diff
@@ -5,7 +5,7 @@ myst:
     "keywords": "Nix, packaging"
 ---
 
-(packaging-existing-software)=
+(packaging-tutorial)=
```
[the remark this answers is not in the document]
fricklerhandwerk: This is somewhat unrelated, because preserving the URL means we have to add a redirect from summer.nixos.org. @delroth (as our infra liaison) where would that be defined?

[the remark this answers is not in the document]
NobbZ: I am not talking about preserving the URL of the [original post written during SoN '22](https://summer.nixos.org/blog/callpackage-a-tool-for-the-lazy/), but about the URL of [packaging existing software tutorial on nix.dev](https://nix.dev/tutorials/packaging-existing-software) whichs URL you change here in this hunk.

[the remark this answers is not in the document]
fricklerhandwerk: Oh no, that's just an internal identifier for cross-references. Changing an URL would mean moving the file, and we'd always put in a redirect in that case.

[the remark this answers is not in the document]
NobbZ: I misunderstood the purpose of that alias.

Thanks for clarifying!

NobbZ (approved):

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+Nix ships with a a special-purpose programming language for creating packages and configurations: the Nix language.
+It is used to build the Nix package collection, known as {term}`Nixpkgs`.
+
+Being purely functional, the Nix language allows declaring custom functions to abstract over common patterns.
+One of the most prominent patterns in Nixpkgs is parametrisation of package recipes.
+
+## Overview
+
+Nixpkgs is sizeable software project on it's own, with coding conventions and idioms that have emerged over the years.
+It has [established a convention](https://github.com/NixOS/nixpkgs/pull/9869) of composing parameterised packages with automatic settings through a function named [`callPackage`](https://github.com/NixOS/nixpkgs/commit/fd268b4852d39c18e604c584dd49a611dc795a9b).
```
infinisil, line 24: The PR there didn't establish the convention. Even before that PR, `callPackage` was used for the vast majority of packages, see [the file before the PR](https://github.com/Mathnerd314/nixpkgs/blob/c545bc80f68a44d0d0e1cc85290d74d1d57c09fa/pkgs/top-level/all-packages.nix).

Here's the actual commit that introduced `callPackage` as a convention: https://github.com/nixos/nixpkgs/commit/d17f0f9cbca38fabb71624f069cd4c0d6feace92

```suggestion
It has [established a convention](https://github.com/nixos/nixpkgs/commit/d17f0f9cbca38fabb71624f069cd4c0d6feace92) of composing parameterised packages with automatic settings through a function named [`callPackage`](https://github.com/NixOS/nixpkgs/commit/fd268b4852d39c18e604c584dd49a611dc795a9b).
```

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+{ writeShellScriptBin }:
+writeShellScriptBin "hello" ''
+  echo "Hello, world!"
+''
+```
+
+:::{dropdown} Detailed explanation
+`hello.nix` declares a function which takes as argument an attribute set with one element `writeShellScriptBin`.
+[`writeShellScriptBin`](https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-writeShellScriptBin) is a function that happens to exist in Nixpkgs, a [build helper](https://nixos.org/manual/nixpkgs/unstable/#part-builders) that returns a derivation.
+The derivation output in this case is an executable shell script with the contents `echo "hello world"` named `"hello"`.
```
infinisil, line 58:
```suggestion
The derivation output in this case contains an executable shell script in `$out/bin/hello` that prints "Hello world" when run.
```

The derivation output would be the entire `$out` directory, not just `$out/bin/hello`.

And it doesn't just contain that string, it also contains a shebang before that.

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+
+```{code-block} nix
+:caption: default.nix
+let
+  pkgs = import <nixpkgs> { };
+in
+pkgs.callPackage ./hello.nix { }
+```
+
+Realise the derivation in `default.nix` and run the executable that is produced:
```
infinisil, line 71: I don't think the terms "realise" and "derivation" work here. The Nix file contains a Nix expression that first needs to be instantiated before it's a derivation.

```suggestion
Build `default.nix` and run the executable that is produced:
```

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+
+```shell-session
+$ nix-build -A hello-people
+$ ./result/bin/hello
+Hello, people!
+```
+
+This pattern is used widely in Nixpkgs:
+For example, functions which represent Go programs often have a parameter `buildGoModule`.
+It is common to find expressions like `callPackage ./go-program.nix { buildGoModule = buildGo116Module; }` to enforce a certain Go compiler version.
```
infinisil, line 151:
```suggestion
It is common to find expressions like `callPackage ./go-program.nix { buildGoModule = buildGo116Module; }` to change the default Go compiler version.
```

Not enforced, can be overridden with `.override`.

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+Nixpkgs is therefore not simply a huge library of pre-configured packages, but a collection of functions – package *recipes* – for customising packages and even entire ecosystems (for example "All Python packages using my custom interpreter") on the fly without duplicating code.
+
+# Overrides
+
+`callPackage` adds more convenience by adding an attribute to the derivation it returns: the `override` function.
+This means that, as a consequence of parameterising builds with `callPackage`, you can also change the value of these parameters _after the fact_ using the returned derivation's `override` function.
```
infinisil, line 157: Doesn't flow very well here, the explanation of what the function does should come earlier.

Maybe just

```suggestion
`callPackage` adds more convenience by allowing parameters to be customised _after the fact_ using the returned derivation's `override` function.
```

It's a bit odd to use terms "after" and "customise" when there's no ordering of operations and nothing being changed, but I think it's fine.

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+-{
++rec {
+   hello = pkgs.callPackage ./hello.nix { };
+   hello-people = pkgs.callPackage ./hello.nix { audience = "people"; };
++  hello-folks = hello.override { audience = "folks"; };
+ }
+```
+
+:::{note}
+The resulting attribute set is now recursive (by the keyword `rec`).
```
infinisil, line 175: Would be good to link to https://nix.dev/guides/best-practices#recursive-attribute-set-rec here

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+Pass the parameter `audience` in the second argument to `callPackage`:
+
+```{code-block} diff
+:caption: default.nix
+ let
+   pkgs = import <nixpkgs> { };
+ in
+ {
+   hello = pkgs.callPackage ./hello.nix { };
++  hello-people = pkgs.callPackage ./hello.nix { audience = "people"; };
```
infinisil, line 134: This is generally not recommended because any changes to the arguments to `hello` won't influence `hello-people`. It's also conceptually the easiest if each file is only imported from one location. This also leads to the question later when `.override` should be preferred over using another `callPackage`.

I'd say this is best avoided by just passing audience to `hello` instead:

```suggestion
-  hello = pkgs.callPackage ./hello.nix { };
+  hello = pkgs.callPackage ./hello.nix { audience = "people"; };
```

This might warrant moving the introduction of `-A` further down.

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+
+Building `hello-folks` attribute and running the resulting executable will again produce a new version of the script:
+
+```shell-session
+$ nix-build -A hello-folks
+$ ./result/bin/hello
+Hello, folks!
+```
+
+A real-world example is the [`neovim`](https://search.nixos.org/packages?show=neovim) package recipe, which has has overrideable arguments such as `extraLuaPackages`, `extraPythonPackages`, or `withRuby`.
```
infinisil, line 191: Important to point out is that these parameters are only discoverable by looking at the source file, which people can get to by clicking the "Source" link on search.nixos.org

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+  a = pkgs.callPackage ./a.nix { };
+  b = pkgs.callPackage ./b.nix { inherit a; };
+  c = pkgs.callPackage ./c.nix { inherit b; };
+  d = pkgs.callPackage ./d.nix { };
+  e = pkgs.callPackage ./e.nix { inherit c d; };
+}
+```
+
+:::{note}
+In this example `inherit a;` can be considered equivalent to `a = a;`.
```
infinisil, line 219:
```suggestion
Here, `inherit a;` is equivalent to `a = a;`.
```

There's some cases where it isn't like in `a: rec { inherit a; }`, but here it is equivalent.

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+:::
+
+Previously declared derivations are passed as arguments to other derivations through `callPackage`.
+
+In this case you have to remember to manually specify all arguments required by each package in the respective {term}`Nix file` that are not in Nixpkgs.
+If `./b.nix` requires an argument `a` but there is no `pkgs.a`, the function call will produce an error.
+This can become quite tedious quickly, especially for larger package sets.
+
+Use `lib.callPackageWith` to create your own `callPackage` based on an attribute set.
```
infinisil, line 228: Might it not be better to go straight to `lib.makeScope`?

source/tutorials/callpackage.md
```diff
@@ -0,0 +1,283 @@
...
+This relieves you from dealing with them manually, and precludes configuration errors that may only surface late into a lengthy build process.
+
+Of course this small example is still manageable in the original form.
+And the implicitly recursive variant can obscure the structure for software developers not familiar with lazy evaluation, making it harder to read for them than it was before.
+But this benefit really pays off for large constructions, where it is the amount of code that would obscure the structure, and where manual modifications would become cumbersome and error-prone.
+
+## Summary
+
+Using `callPackage` does not only follow Nixpkgs conventions, which makes your code easier to follow for experienced Nix users.
+It also gives you some benefits for free:
```
infinisil, line 270:
```suggestion
Using `callPackage` does not only follow Nixpkgs conventions, which makes your code easier to follow for experienced Nix users, but it also gives you some benefits for free:
```

infinisil (changes requested):
Pretty nice overall!

source/tutorials/callpackage.md
fricklerhandwerk: Uhh thanks for the dig-up!

source/tutorials/callpackage.md
fricklerhandwerk: I thought so too until I tried, and it makes things really hairy. I'd rather avoid it, because it would take forever to get right, and the tutorial already provides a lot of value.

source/tutorials/callpackage.md
fricklerhandwerk: We haven't really converged on a robust terminology here. In the Nix manual we distinguish between [derivation](https://hydra.nixos.org/build/253249891/download/1/manual/glossary.html#gloss-derivation) (which is rather vaguely defined) and store derivation (the thing that needs to be instantiated). Also "to build" something isn't well defined (although arguably it is the specific act of "running the `builder` executable of a derivation"), while "to realise" is (and is broader than "to build"). 

In short: I'd rather leave that as is, at least for now.

source/tutorials/callpackage.md
fricklerhandwerk: I've thought about it, but would rather skip it here:
- It's besides the point. Adding more comments will distract from the actual learning objective. The note is merely to remind novice readers what `rec` is about so they can fully grasp the example.
- Referring to a piece that says "`rec` bad" from a place where we very deliberately use `rec` for clarity and small diff seems weird and counterproductive to building trust in the information we provide.
- Thinking about "`rec` bad" made me doubt this should be here on nix.dev to begin with. It seems it's more of a Nixpkgs convention, where it makes sense if you have huge files and the need to separate interface from implementation. But generally, using `rec` judiciously can greatly increase readability in my opinion.

fricklerhandwerk:
Linkcheck fails due regressions on nixos.org, I opened an issue https://github.com/NixOS/nixos-homepage/issues/1348.

PR #938, opened 2024-03-10
Update packaging-existing-software.md

bew:
Few things I noticed while reading this tutorial, and a few rephrasing to help reading. 

:warning: There is also a difference between the last two examples that is not explained, where the last code example depends on `xorg.libX11.dev` but earlier examples only depends on `xorg.libX11`

/cc @fricklerhandwerk

### DIFF
--- a/source/tutorials/packaging-existing-software.md
+++ b/source/tutorials/packaging-existing-software.md
@@ -427,7 +427,7 @@ But the important bit for this tutorial is `fatal error: X11/Xlib.h: No such fil
 
 ## Finding packages
 
-Determining from where to source a dependency is currently a somewhat involved, because package names don't always correspond to library or program names.
+Determining from where to source a dependency is currently somewhat involved, because package names don't always correspond to library or program names.
 
 You will need the `Xlib.h` headers from the `X11` C package, the Nixpkgs derivation for which is `libX11`, available in the `xorg` package set.
 There are multiple ways to figure this out:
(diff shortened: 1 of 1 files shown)


olafklingt:
lgtm

source/tutorials/packaging-existing-software.md
```diff
@@ -446,21 +446,28 @@ In case all else fails, it helps to become familiar with searching the [Nixpkgs
...
 $ nix-shell -p git ripgrep
+[nix-shell:~]$ # git & rg are available in this shell!
+```
+
+Since the Nixpkgs repository is huge, it can be faster to only clone the latest revision to avoid waiting a long time for a full clone:
+
+```console
 [nix-shell:~]$ git clone https://github.com/NixOS/nixpkgs --depth 1
+...
+[nix-shell:~]$ cd nixpkgs/
```
bew, line 464: I also noticed that all prompts below this line are actually wrong, since we did a `cd nixpkgs/` they should be like:
```diff
-[nix-shell:~]$ some commands....
+[nix-shell:~/nixpkgs]$ some commands...
```

Do you want me to change these as well?

source/tutorials/packaging-existing-software.md
```diff
@@ -480,7 +487,7 @@ Since `rg` is case sensitive by default,
 Add `-i` to make sure you don't miss anything:
 
 ```
-[nix-shell:~]$ rg -i "libx11 =" pkgs
+[nix-shell:~]$ rg -i "libx11 =" pkgs/
```
fricklerhandwerk, line 490:
```suggestion
[nix-shell:~]$ rg -i "libx11 =" pkgs
```

source/tutorials/packaging-existing-software.md
```diff
@@ -496,7 +503,7 @@ pkgs/servers/x11/xorg/overrides.nix
 
 ### `nix-locate`
 
-Consider using `nix-locate` from the [`nix-index`](https://github.com/nix-community/nix-index) tool to find derivations that provide what you need.
+Another option is to use `nix-locate` from the [`nix-index`](https://github.com/nix-community/nix-index) tool to find derivations that provide what you need.
```
fricklerhandwerk, line 506:
```suggestion
To search derivations on the command line, use `nix-locate` from the [`nix-index`](https://github.com/nix-community/nix-index).
```

fricklerhandwerk (approved):
@bew thanks a lot for the fixes! I reverted a few small things, but feel free to make new PRs if you think those should still be discussed.

source/tutorials/packaging-existing-software.md
```diff
@@ -496,7 +504,7 @@ pkgs/servers/x11/xorg/overrides.nix
 
 ### `nix-locate`
```
fricklerhandwerk, line 505:
```suggestion
### Local derivation search
```

fricklerhandwerk:
> There is also a difference between the last two examples that is not explained, where the last code example depends on `xorg.libX11.dev` but earlier examples only depends on `xorg.libX11`

It seems to make no difference, and it builds and runs just as well without it, so I removed it.

source/tutorials/packaging-existing-software.md
```diff
@@ -446,21 +446,28 @@ In case all else fails, it helps to become familiar with searching the [Nixpkgs
...
+...
+[nix-shell:~]$ cd nixpkgs/
 ```
 
-To narrow down results, specify which subdirectory you want to search:
+To narrow down results, specify which subdirectory you want to search (`pkgs` folder in nixpkgs holds all packages):
 
 ```console
-[nix-shell:~]$ rg "x11 =" pkgs
+[nix-shell:~]$ rg "x11 =" pkgs/
```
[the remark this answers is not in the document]
bew: My reasoning was that it makes it explicit for the reader that it is the path to a directory, not just a dumb identifier.
`¯\_(ツ)_/¯`

bew:
Thanks for the review & quick merge 👌

I like your suggestions, makes the flow of sentences simpler.
I think we could explain more about why we search `something =` (why `=` ? 🤔), and the benefit of `nix-locate` (I haven't tried it yet actually)

Issue #926, opened 2024-02-26
Add a section to for a NixOS graphical configuration in the VM tutorial


fricklerhandwerk:
## Problem

An obvious next step after getting into the subject with a lightweight VM is to prototype a graphical setup before getting into bare-metal installation. Apparently it's entirely non-obvious how to actually do that, given you need to "somehow know" the three (!) configuration options that have to be set to make it work, compounded by the existence of many display managers and window managers, and the choice between X and Wayland.

## Proposal

Add a section at the end of the tutorial that shows a working graphical VM configuration, including how to invoke it, and point to search.nixos.org/options for more... options.

PR #940, opened 2024-03-10
nixos-configuration-on-vm.md: graphical vm

olafklingt:
aims to solve #926


olafklingt:
right now this is basically a draft. 
i just wrote down what i know.
Before I continue I would like to know if those two examples make sense in the context of this tutorial. Each example introduces another hurdle and is maybe confusing.

olafklingt:
I think the additions to the tutorial are done. I hope it doesn't add confusion.

infinisil:
Note that we worked on this together in todays docs team meeting. Instead of making review comments we just pushed a commit instead, which makes it a lot faster.

@fricklerhandwerk also wanted to do a pass to make the wording in that tutorial more consistent.

source/tutorials/nixos/nixos-configuration-on-vm.md
```diff
@@ -186,14 +174,6 @@ This command builds the attribute `vm` from the `nixos-23.11` release of NixOS,
 
   Here we set `nixpkgs` to refer to a [specific version of Nixpkgs](ref-pinning-nixpkgs) and set `nix-config` to the `configuration.nix` file in the current directory.
 
-:::{admonition} NixOS
-On NixOS the `$NIX_PATH` environment variable is usually set up automatically, and there is also [a convenience command for building virtual machines](https://nixos.org/manual/nixos/stable/#sec-changing-config).
-You can use the current version of `nixpkgs` to build the virtual machine like this:
-```shell-session
-nixos-rebuild build-vm -I nixos-config=./configuration.nix
-```
```
olafklingt, line 188: Why do you suggest to remove this command/section?

olafklingt (commented):
great, i just have one question above.

source/tutorials/nixos/nixos-configuration-on-vm.md
infinisil: `nixos-rebuild` is really intricate with how it works. When I tested it, it failed to do what one would expect because there was an `/etc/nixos/flake.nix` file..

To keep this tutorial more focused on the use case of building a VM just using Nix, we thought it made sense to avoid any potential confusion with `nixos-rebuild`.

olafklingt: oh yes flakes ... ok.

olafklingt:
something went wrong when you updated the pr ... i don't think you reviewed the latest version.

PR #943, opened 2024-03-17
fix: minor grammar correction

cybardev:
Very minor grammar issue. Noticed while reading the docs.


DanielSidhion:
Thank you for the fix!

PR #951, opened 2024-04-01
point wiki links to official wiki

JulienMalka:


DanielSidhion (approved):
Thanks! We'll definitely want to change the description on that section as well, but I'm merging this for now to get the official link out.

PR #954, opened 2024-04-06
correct the number of flake users

Mic92:
The statistic in
https://discourse.nixos.org/t/nix-community-survey-2023-results/33124 does also include N/A responses in the percentage, so the 58% is a bit misleading as the conclusion would be that 42% are still not using flakes. I dropped the 1/3 statement for now, because I couldn't find any information about that, but I can re-add it if this percentage is actually correct. I also dropped the 2022 results as it didn't seem to have the explicit question in it.

### DIFF
--- a/source/concepts/flakes.md
+++ b/source/concepts/flakes.md
@@ -68,7 +68,7 @@ In particular:
 This led to a situation where the stable interface was only sparsely maintained for multiple years, and repeatedly suffered breakages due to ongoing development.
 Meanwhile, the new interface was adopted widely enough for evolving its design without negatively affecting users to become very challenging.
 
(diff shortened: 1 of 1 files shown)


hsjobeki:
Lgtm. However notable that it is based on the assumption that the N/A people are distributed in the same way as the respondents.

I included the image here and did the calculation landing at the same 84% 👍

image

Mic92:
I also requested the raw numbers for this question of the survey here:

[Survey_2023_Nix_Community_Survey_2023.pdf](https://github.com/NixOS/nix.dev/files/14895849/Survey_2023_Nix_Community_Survey_2023.pdf)

This pdf removes `N/A` if you are too lazy to calculate it yourself:

[Survey_2023_Nix_Community_Survey_2023-2.pdf](https://github.com/NixOS/nix.dev/files/14895852/Survey_2023_Nix_Community_Survey_2023-2.pdf)

olafklingt (approved):

PR #955, opened 2024-04-07
Improve Documentation: Exclude Prompt Characters from Copied Shell Commands

andylizf:
Prevent the inclusion of the dollar sign and preceding characters when copying shell commands from the documentation by wrapping them in a `console` code block.

### DIFF
--- a/source/tutorials/first-steps/ad-hoc-shell-environments.md
+++ b/source/tutorials/first-steps/ad-hoc-shell-environments.md
@@ -37,7 +37,7 @@ these 3 derivations will be built:
 
 Within the Nix shell, you can use the programs provided by these packages:
 
-```
+```console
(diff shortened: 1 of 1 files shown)


source/tutorials/first-steps/ad-hoc-shell-environments.md
```diff
@@ -37,7 +37,7 @@ these 3 derivations will be built:
 
 Within the Nix shell, you can use the programs provided by these packages:
 
-```
+```console
```
infinisil, line 40:
~~~suggestion
```shell-session
~~~

To match the surrounding code blocks

andylizf: Maybe single line commands need to be wrapped in "console"?

PR #956, opened 2024-04-09
deduplicate CONTRIBUTING.md

fricklerhandwerk:
all the information is alreaady in the contributing guide on the website

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,36 +2,18 @@
 
 nix.dev is a community effort to collect, create, and maintain world-class learning resources for Nix.
 
-We strongly encourage everyone interested to contribute by asking informed questions or directly proposing changes.
+We strongly encourage everyone interested to participate:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,36 +2,18 @@
...
-
-## Code of conduct
-
-Adapted from the [Contributor Covenant] and [The Carpentries Code of Conduct]:
-
-- Use welcoming and inclusive language
-- Show empathy and respect towards other people
-- Be respectful of different viewpoints and experiences
-- Give and gracefully accept constructive criticism
-- Focus on what is best for the community
```
infinisil, line 20: There's no code of conduct on the contributing page (which is notably also linked to from nixos.org). We do have https://github.com/NixOS/.github/blob/master/CODE_OF_CONDUCT.md though, which is accessible from all the repos, so I think it's fine

fricklerhandwerk: I'll add that in a follow-up PR, since the contributor guides need a bit of clean-up as well.

PR #957, opened 2024-04-09
split out a page for documentation resources

fricklerhandwerk:
this looks like a large change, but it simplifies a few things:
- separate page for documentation resources, with overall less text
- updated information on the official NixOS Wiki
- communication channels are now subsumed under "how to get help"
- a few wording corrections while at it


olafklingt:
#959 should fix lint check

PR #959, opened 2024-04-09
Update accept.txt

olafklingt:
let vale accept NixOS Wiki


infinisil:
Nice! How about adding yourself as codeowner of that file? That way it lessens the notification load for other team members :)

PR #960, opened 2024-04-12
refactor: rename test.nix to default.nix for consistency

ben-christian:
Bit of a nit pick but will keep the example consistent


infinisil (approved):
Thanks!

PR #963, opened 2024-04-15
highlight background of collapsed texts

olafklingt:
I usually think that text collapsed behind a details element is difficult to recognize as such. 
This is a proposal to highlight those elements with a blue background color.


infinisil (approved):
Looking good!

PR #964, opened 2024-04-17
update nix version in docs and link

eflanagan0:
I prefer seeing versions align with the current stable of a tool.


olafklingt:
There is no automatic update of versions because there is no automatic testing for that. 
The current version of nix on NixOS Stable is 2.18.1 .
As the tutorial suggests different ways to install nix (outside of NixOS) someone else should decide if the change is in line with the Tutorial.

PR #965, opened 2024-04-17
replace unoffical wiki links with offical ones

jopejoe1:
We now have an unofficial wiki page, so we should link that one instead of the unofficial ones


olafklingt:
thank you

PR #968, opened 2024-04-20
Minor improvements

pdietl:
Add more error checking to the `map` and `geocode` script
Rename `map` to `map.sh` and `geocode` to `geocode.sh` in hopes that
when downloading, Google Chrome doesn't automatically add a `.txt` extension.

The default zoom of 2 results in the `center` option not showing anything
different. So, we change the default zoom to 10.

We also clear up the fact that using an option type of `nullOr <blah>` does not
set a default value of `null` for you.


fricklerhandwerk:
@infinisil I did the history rewrite.

PR #972, opened 2024-04-24
fix syntax error

stablejoy:
Update packaging-existing-software.md - fix syntax error in skeleton derivation example. 
Remove `;` from
```
{ stdenv }:
stdenv.mkDerivation {	};
```
This gives `error: syntax error, unexpected ';', expecting end of file`

Removing the semicolon:
```
nix-instantiate --eval --expr '{ stdenv }: stdenv.mkDerivation {}'
<LAMBDA>
```


infinisil (approved):
Thanks!

PR #974, opened 2024-05-04
`lib` unused binding in hello.nix

stablejoy:
We apply `lib.fakeSha256` to get the right hash. But, upon removing `lib.fakeSha256`, the `lib` in the argument set becomes an unused binding. Should we delete it from the arguments then? Should there be something written in the tutorial that explains it? Also later the `hello.nix` is copied into `icat.nix` where the unbinding issue remains.

### DIFF
--- a/source/tutorials/packaging-existing-software.md
+++ b/source/tutorials/packaging-existing-software.md
@@ -306,7 +306,7 @@ Navigate to the project's [Tags page](https://github.com/atextor/icat/tags) to f
 In this case, the latest release tag is `v0.5`.
 
 As in the `hello` example, a hash must also be supplied.
-This time, instead of using an empty string and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command.
+This time, instead of using the empty string and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command.
 
 You need the SHA256 hash of the *contents* of the tarball (as opposed to the hash of the tarball file itself).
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-existing-software.md
```diff
@@ -96,7 +95,7 @@ stdenv.mkDerivation {
 
   src = fetchzip {
     url = "https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz";
-    sha256 = lib.fakeSha256;
+    sha256 = " ";
```
fricklerhandwerk, line 98:
```suggestion
    sha256 = "";
```

source/tutorials/packaging-existing-software.md
```diff
@@ -309,7 +306,7 @@ Navigate to the project's [Tags page](https://github.com/atextor/icat/tags) to f
 In this case, the latest release tag is `v0.5`.
 
 As in the `hello` example, a hash must also be supplied.
-This time, instead of using `lib.fakeSha256` and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command.
+This time, instead of using an empty string and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command.
```
fricklerhandwerk, line 309:
```suggestion
This time, instead of using the empty string and letting `nix-build` report the correct one in an error, you can fetch the correct hash in the first place with the `nix-prefetch-url` command.
```

fricklerhandwerk (approved):
Nice, thanks a lot!

PR #988, opened 2024-05-26
proxy development version of the Nix manual

fricklerhandwerk:
this avoids exposing an ugly URL and allows us to move things
around later

Made with @roberth at #ZurichZHF

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -33,3 +33,5 @@
 /tutorials/module-system/module-system /tutorials/module-system/deep-dive 301
 
 /permalink/stub-ld /guides/faq#how-to-run-non-nix-executables 301
+/manual/nix/unstable https://hydra.nixos.org/job/nix/master/build.x86_64-linux/latest/download 200
+/manual/nix/development https://hydra.nixos.org/job/nix/master/build.x86_64-linux/latest/download 200
--- a/source/reference/nix-manual.md
+++ b/source/reference/nix-manual.md
@@ -17,7 +17,7 @@ This page is pre-processed before rendering with Sphinx. For details:
 ```{toctree}
 :hidden:
 
-Nix pre-release (development) <https://hydra.nixos.org/job/nix/master/build.x86_64-linux/latest/download>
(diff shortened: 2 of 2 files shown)


_redirects
```diff
@@ -33,3 +33,5 @@
 /tutorials/module-system/module-system /tutorials/module-system/deep-dive 301
 
 /permalink/stub-ld /guides/faq#how-to-run-non-nix-executables 301
+/manual/nix/unstable https://hydra.nixos.org/job/nix/master/build.x86_64-linux/latest/download 200
+/manual/nix/development https://hydra.nixos.org/job/nix/master/build.x86_64-linux/latest/download 200
```
infinisil, line 37: Testing this locally, the proxying doesn't actually work, it does a full redirect

fricklerhandwerk: We can reasonably assume this may work live, and if it doesn't it's at least not a degradation. In any case this change is a hard requirement for https://github.com/NixOS/nixos-homepage/pull/1446

infinisil:
This doesn't work! https://nixos.org/manual/nix/stable goes to https://nix.dev/manual/nix/2.18/stable/, which is a 404

fricklerhandwerk:
@infinisil @roberth apparently proxying doesn't work in the live depolyment either, it's just a 302 redirect.

```shell
$ curl -I https://nix.dev/manual/nix/development
HTTP/2 302
age: 1
cache-status: "Netlify Edge"; fwd=miss
content-type: text/html; charset=utf-8
date: Sun, 26 May 2024 19:01:29 GMT
location: https://hydra.nixos.org/build/261304987/download
netlify-vary: query
server: Netlify
set-cookie: hydra_session=98cabe9914f21a207f36894eb80485db413012de; path=/; expires=Sun, 02-Jun-2024 19:01:29 GMT; HttpOnly; SameSite=Lax
strict-transport-security: max-age=31536000
x-nf-request-id: 01HYV4EYRE3Z5GR8J5T6WM09TB
content-length: 332
```

PR #997, opened 2024-06-11
offer the Nix manual as a single page again

fricklerhandwerk:
@infinisil


infinisil (commented):
Hacky but sure :laughing: 

Though do keep in mind that this increases the surface of what any potential future doc tooling is expected to support.

PR #1001, opened 2024-06-17
Niv ? Npins?

jd1t25:
Hi , I am new and still learning, 
Following the wiki , during "nix-build" it throws error

in default.nix 
`sources ? import ./nix/sources.nix,`
`nix/sources.nix': No such file or directory`

before that we run npin
`nix-shell -p npins --run "npins init --bare; npins add github nixos nixpkgs --branch nixos-23.11"`

tree npins
```bash
npins
├── default.nix
└── sources.json
```

which clearly doesn't what sources.nix

after researching on its github , i found its alternative niv 
then reading its wiki and running this 
`nix-shell -p niv --run "niv init; niv add NixOS/nixpkgs -n nixpkgs -b nixos-23.11"`

tree niv
```bash
nix
├── sources.json
└── sources.nix
```
Either use niv or npin
( Or i mess up something big , and this pr just a mistake)

### DIFF
--- a/source/tutorials/working-with-local-files.md
+++ b/source/tutorials/working-with-local-files.md
@@ -99,7 +99,7 @@ Then create a `default.nix` file with the following contents:
 :caption: default.nix
 {
   system ? builtins.currentSystem,
-  sources ? import ./nix/sources.nix,
+  sources ? import ./npins/default.nix,
 }:
 let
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Ah yes, thanks a lot - I missed that spot when changing things from niv to npins.

PR #1004, opened 2024-06-19
JSON convert in modules error

jd1t25:
Hi again, 
while continuing following and learning , under Section modules 

when you run script "./eval.bash" it shows
```
{ pkgs ? import <nixpkgs> { } }:
  ^
 error: cannot convert a function to JSON
```

can we convert function to json in nix? ( did i miss something )
so i just declared pkgs inside   "let .... in" block 

```
let
  pkgs = import <nixpkgs> {};
....
in
```

after executing it shows 
```
while evaluating attribute 'name'
error: cannot convert a function application to JSON
```

after reading wiki , i followed with    **--strict**
```
nix-shell -p jq --run "nix-instantiate --eval --json --strict | jq"
```

now it does show the expected output.
```
{
  "name": "Boaty McBoatface"
}
```

Please check on your end , i did double check about the error and also there might be better solution than this

### DIFF
--- a/source/tutorials/module-system/a-basic-module/default.nix
+++ b/source/tutorials/module-system/a-basic-module/default.nix
@@ -7,4 +7,4 @@ let
     ];
   };
 in
-  result.config
+result.config
--- a/source/tutorials/module-system/a-basic-module/default.nix
+++ b/source/tutorials/module-system/a-basic-module/default.nix
@@ -1,5 +1,5 @@
-{ pkgs ? import <nixpkgs> { } }:
 let
+  pkgs = import <nixpkgs> {};
   result = pkgs.lib.evalModules {
     modules = [
       ./options.nix
@@ -7,4 +7,4 @@ let
     ];
   };
 in
-result.config
+  result.config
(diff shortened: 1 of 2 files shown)


source/tutorials/module-system/a-basic-module/default.nix
```diff
@@ -1,10 +1,10 @@
...
+  pkgs = import <nixpkgs> {};
   result = pkgs.lib.evalModules {
     modules = [
       ./options.nix
       ./config.nix
     ];
   };
 in
-result.config
+  result.config
```
djacu, line 10: nit: leave the white space before `result.config` out. Which formatter are you using? `nixfmt-rfc-style` was made official back in March I believe.

```suggestion
result.config
```

djacu (dismissed):
@proofconstruction @fricklerhandwerk Other than my nit I think this is good. I can't think of another way to make the `default.nix` file work without using `--apply` and that adds a lot of complexity to the shell script.

fricklerhandwerk:
Oh damn, `nix-instantiate` doesn't do autocall like `nix-build` and `nix-shell`.

fricklerhandwerk:
This sucks. But we don't have to change the default argument, because that's still useful if you end up building derivations. Just the command line invocation becomes cumbersome and introduces redundancy:

```bash
nix-shell -p jq --run "nix-instantiate --eval --json --strict --arg pkgs 'import <nixpkgs> {}' | jq"
```

fricklerhandwerk (approved):

PR #1009, opened 2024-06-28
dependency-management.md: update niv references

AndersonTorres:
1. A small fixup/typo over an older reference to `niv`;
2. `niv` is not exactly unmaintained - the most recent commit is from less than 2 months ago. Let's rewrite this in a more neutral manner.


fricklerhandwerk (approved):
Thanks!

Issue #131, opened 2021-04-13
mkShell examples should use nativeBuildInputs  not buildInputs


arianvp:
buildInputs only works for legacy reasons as far as  I understand.

Furthermore benefit of `nativeBuildInputs` is that it makes  bash-completion work for things installed with nix-shell on `nixpkgs-unstable`

eyeinsky:
@arianvp Could you expand why this is so? Recently while debugging an issue found this https://nixos.wiki/wiki/Development_environment_with_nix-shell#No_GSettings_schemas_are_installed_on_the_system where it uses `buildInputs`.

arianvp:
You won't get autocomplete and man pages with buildInputs

eyeinsky:
Other than that, will it "just work"? If I'm not mistaken using `buildInputs` will be often used by those who reuse their default.nix as shell.nix, and, khm, don't have the knowledge to know the difference.

arianvp:
But for default.nix it's also wrong. Applications needed for your build need to go into nativeBuildInputs otherwise cross compilation does not work.  Libraries that are linked against go into buildInputs

hsjobeki:
Fixed in nix.dev; we use packages, which is mapped to nativeBuildInputs:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/mkshell/default.nix#L41C3-L41C21

PR #1020, opened 2024-07-14
fix mkShellNoCC to use `packages` instead of `buildInputs`

willbush:
According to:
https://github.com/NixOS/nix.dev/blob/cdd059f7e5627b68a137d4de3ea98ab100bdb5a9/source/tutorials/first-steps/declarative-shell.md?plain=1#L84

>The `packages` attribute argument to `mkShellNoCC` is simply an alias for `nativeBuildInputs`.

I believe https://github.com/NixOS/nix.dev/issues/131 can be closed now. I think this is one of the last remaining instances where mkShell* uses `buildInputs`.

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1936,7 +1936,7 @@ let
   message = "hello world";
 in
 pkgs.mkShellNoCC {
-  buildInputs = with pkgs; [ cowsay ];
+  packages = with pkgs; [ cowsay ];
   shellHook = ''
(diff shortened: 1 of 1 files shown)


hsjobeki (approved):
lgtm.

Imagine we have actual api documentation. We'd instead link to https://nixos.org/manual/nixpkgs/stable/#sec-pkgs-mkShell. Explain the concept here, link to the function interface documentation.

willbush:
@hsjobeki can #131 be closed?

PR #1021, opened 2024-07-15
Add requirements.txt

nolash:
I wanted to merely clone and build offline docs. 

To do so, I had to make a virtualenv and manually install the requirements.

I've `pip freeze`'d those requirements and figured adding them to the repo could be of help to someone else who seeks to do the same.


fricklerhandwerk:
Hi @nolash, I'm not exactly sure what exactly you want to do and how this will help. At least the Nix manual index page is already generated by Nix, and there may be more of that in the future. So whatever you would build in a vanilla Python environment will be incomplete. Is there something you need but can't do with `nix-build` right now?

nolash:
@fricklerhandwerk the contribution was to allow building documentation for nix without nix. As in, without the thing being documented.

Running vanilla docs takes less than a minute on a medium shitty line. nix-build has already ran 5 minutes here downloading without budging.

This is what I built vanilla: https://holbrook.no/~lash/tmp/nixdevbuild/install-nix.html#install-nix

I was thinking merely as a courtesy.

fricklerhandwerk:
Yeah, evaluating all the Nix versions to get the manuals takes forever, and a more correct thing to do would be pulling the build output directly from Hydra. But Hydra doesn't expose the store paths in an API and I'm not gonna scrape the Hydra webpage just for this. If you just want to have a shortcut, maybe make the manual replacements optional with a parameter? There's really no need to duplicate information that will move around anyway.

fricklerhandwerk:
@nolash in #1035 building the manuals by default is disabled, which achieves the same result but without redundancy.

PR #1025, opened 2024-07-24
Add first build information in section of working with local files

YSawc:
~- In union working section of [Working with local files](https://nix.dev/tutorials/working-with-local-files#union-explicitly-exclude-files), code difference is not clearly written so fix it.~
- add first build information in section of [Working with local files](https://nix.dev/tutorials/working-with-local-files#union-explicitly-exclude-files)

### DIFF
--- a/source/tutorials/working-with-local-files.md
+++ b/source/tutorials/working-with-local-files.md
@@ -380,7 +380,7 @@ Create a file set containing a union of the files to exclude (`fs.unions [ ... ]
 +      ]);
 ```
 
-This will work as expected.
+This will work as expected:
 
 ```
 $ nix-build
--- a/source/tutorials/working-with-local-files.md
+++ b/source/tutorials/working-with-local-files.md
@@ -366,17 +366,31 @@ One way to fix this is to use [`unions`](https://nixos.org/manual/nixpkgs/stable
 
 Create a file set containing a union of the files to exclude (`fs.unions [ ... ]`), and subtract it (`difference`) from the complete directory (`./.`):
 
-```{code-block} nix
+```{code-block} diff
 :caption: build.nix
-  sourceFiles =
-    fs.difference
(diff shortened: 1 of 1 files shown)


source/tutorials/working-with-local-files.md
```diff
@@ -366,17 +366,31 @@ One way to fix this is to use [`unions`](https://nixos.org/manual/nixpkgs/stable
 
 Create a file set containing a union of the files to exclude (`fs.unions [ ... ]`), and subtract it (`difference`) from the complete directory (`./.`):
 
-```{code-block} nix
+```{code-block} diff
```
fricklerhandwerk, line 369: We don't always have to make it a difference. There's no firm rule, but I tried making it a diff where it's a small change - in order to highlight that - and have a full file, like here, where it's essentially a new thing.

source/tutorials/working-with-local-files.md
```diff
@@ -366,17 +366,31 @@ One way to fix this is to use [`unions`](https://nixos.org/manual/nixpkgs/stable
...
++      ./.
++      (fs.unions [
++        (fs.maybeMissing ./result)
++        ./default.nix
++        ./build.nix
++        ./npins
++      ]);
+```
+
+This will work as expected.
```
fricklerhandwerk, line 383:
```suggestion
This will work as expected:
```

source/tutorials/working-with-local-files.md
YSawc: [11c9f26](https://github.com/NixOS/nix.dev/pull/1025/commits/11c9f2643881fde0e4446d53a1493822434a8529)
Ok, I remove code defferences and change the description of PR.

fricklerhandwerk (approved):
Thanks for taking the time! Every little fix makes the next person's journey a bit easier

PR #1027, opened 2024-08-10
Clarify wording around `nix-shell --run`

toraritte:
This may seem pedantic, but I found the part saying that `--run` "_executes the given Bash command_" command misleading, especially if someone new to Linux shells would read it.

This suggestion is far from being perfect, but anything else I came up with seemed either too long-winded and/or out of place.

Thank you for these great write-ups!

### DIFF
--- a/source/tutorials/first-steps/ad-hoc-shell-environments.md
+++ b/source/tutorials/first-steps/ad-hoc-shell-environments.md
@@ -179,7 +179,7 @@ git version 2.33.1
 
 There are three things going on here:
 
-1. `--run` executes the given command within the resulting Nix shell (which is a [Bash](https://www.gnu.org/software/bash/) sub-shell) and exits when it's done.
+1. `--run` executes the given [Bash command](https://www.gnu.org/software/bash/manual/bash.html#Shell-Commands) within the environment created by Nix, and exits when it's done.
 
    You can use this with `nix-shell` whenever you want to quickly run a program you don't have installed on your machine.
 
--- a/source/tutorials/first-steps/ad-hoc-shell-environments.md
+++ b/source/tutorials/first-steps/ad-hoc-shell-environments.md
@@ -179,7 +179,7 @@ git version 2.33.1
 
 There are three things going on here:
 
-1. `--run` executes the given [Bash](https://www.gnu.org/software/bash/) command within the Nix shell and exits when it's done.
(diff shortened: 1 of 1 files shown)


source/tutorials/first-steps/ad-hoc-shell-environments.md
```diff
@@ -179,7 +179,7 @@ git version 2.33.1
 
 There are three things going on here:
 
-1. `--run` executes the given [Bash](https://www.gnu.org/software/bash/) command within the Nix shell and exits when it's done.
+1. `--run` executes the given command within the resulting Nix shell (which is a [Bash](https://www.gnu.org/software/bash/) sub-shell) and exits when it's done.
```
fricklerhandwerk, line 182:
```suggestion
1. `--run` executes the given [Bash command](https://www.gnu.org/software/bash/manual/bash.html#Shell-Commands) within the environment created by Nix, and exits when it's done.
```

More of a source of confusion would be that there really isn't such thing as a "Nix shell" - all Nix does is create an environment to spawn Bash with. There's a page in Bash documentation that expands on the concept of a command. But other than the statement is correct: That flag takes a Bash command.

fricklerhandwerk (changes requested):

source/tutorials/first-steps/ad-hoc-shell-environments.md
toraritte: Wow, I always struggled to understand the concept of "Nix shell" and now I learned something new! I think adding your comment into a collapsible `<detail>` section would be enlightening. Thank you!

fricklerhandwerk: I don't feel comfortable with explaining away a misconception that was built up somewhere else. The idea here is to provide a good approximation to the truth, compactly. Ideally nix.dev is the first encounter for Nix users, and with what we give them they will be able to decipher the rest of the ecosystem.

toraritte: Understandable. This closed PR will be here if someone else falls into the same trap as me. Thanks again!

fricklerhandwerk:
@toraritte You correctly pointed out a lack of clarity there, and that led to a nice small improvement! There were some issues with broken links, which is why I couldn't merge it faster.

fricklerhandwerk (approved):

PR #1028, opened 2024-08-10
Update declarative-shell.md

toraritte:
I presume that at one point there was only one package that was referred from `pkgs` and then another one was added later. Or I misunderstood something and the sentence **is** correct, so perhaps there should be a note there what the one is referring to... Thanks again!


fricklerhandwerk (approved):

fricklerhandwerk:
Good catch, thanks a lot!

PR #1057, opened 2024-09-09
hotfix nix install URL: set to 2.18.5

roberth:


roberth:
Done the right way, and to 2.23; all the way back to 2.18 was unnecessary.

PR #1060, opened 2024-09-13
add admonitions for macOS 15 Sequoia update

abathur:
Same premise/language as https://github.com/NixOS/nix/pull/11487

The impending release of macOS 15 Sequoia will break many existing nix installs on macOS, which may lead to an increased number of people who are looking to try to reinstall Nix without noticing the open/pinned issue (NixOS#10892) that explains the problem and outlines how to migrate existing installs.

These admonitions are a short-term measure until we are over the hump and support volumes dwindle.

### DIFF
--- a/source/install-nix.md
+++ b/source/install-nix.md
@@ -18,6 +18,14 @@ On Arch Linux, you can alternatively [install Nix through `pacman`](https://wiki
 
 ::::{tab-item} macOS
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Gah, this linkchecker seems to annoy many upstreams, but no time to fix it now. Force-merging to get it in.

PR #1061, opened 2024-09-16
update Nix releases

fricklerhandwerk:


proofconstruction:
Do we have a script or any other infra for doing this automatically or is this a manual process?

fricklerhandwerk:
It's a script but I currently run it manually once in a while: https://github.com/NixOS/nix.dev/blob/master/CONTRIBUTING.md#updating-reference-manuals

It needs a bit of attention still because it for some reason includes all the `backport-*` branches. But then it could be scheduled to run once week or something. But ideally it would get triggered builds succeeding.

PR #1062, opened 2024-09-17
`docker load` with Nix Flakes

peterbecich:
Very useful to load a Nix-built Docker image directly into Docker.

This documents a small difference about Nix Flake-built Docker image directly into Docker: `--print-out-paths`

### DIFF
--- a/source/tutorials/nixos/building-and-running-docker-images.md
+++ b/source/tutorials/nixos/building-and-running-docker-images.md
@@ -101,6 +101,12 @@ $ docker load < $(nix-build hello-docker.nix)
 Loaded image: hello-docker:y74sb4nrhxr975xs7h83izgm8z75x5fc
 ```
 
+If you use Nix Flakes and have defined a package `helloDocker`, the equivalent would be:
+```shell-session
+$ docker load < $(nix build .#helloDocker --print-out-paths)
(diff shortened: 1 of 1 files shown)


peterbecich:
@fricklerhandwerk please review this, i.m.o. this change makes it easier to follow the documentation while using a Nix Flake

fricklerhandwerk:
Hi @peterbecich, thanks for your suggestion. I won't merge this one for the following reasons:

- nix.dev only documents stable interfaces, which means, flakes are not a supported use case. This was decided a long time ago, and you can find the rationale in [meeting notes from 2022 on Discourse](https://discourse.nixos.org/c/dev/documentation/25) or in closed issues here on GitHub. The conditions on which that decision was based have not changed, so there is no reason yet to re-open the discussion.
- We aim to provide exactly one known-good way to do things to reduce cognitive load on learners and maintenance effort for contributors. Conditionals like "if you use flakes" just don't fit in there.
- In this particular case I don't see any benefit in highlighting an alternative way to do things, as it has the exact same effect as what is stated in the previous sentence – with the only exception that it assumes you've done something that is not taught in the tutorials. Which is another aspect: we aim to make the curriculum self-contained; to achieve the objectives in the tutorial series, you shouldn't have to know anything except for what is provided or referenced.


I also have personal opinions for why not to promote experimental features and flakes in particular, happy to discuss those in other venues.

PR #1064, opened 2024-09-21
typo: no folder named 'nix' in this example

proudline:
Should be './npins' folder instead of './nix' for coherence


fricklerhandwerk (approved):
Good catch, thanks!

PR #1065, opened 2024-09-30
Link nixpkgs library in module-system tutorial

ttrssreal:
Hi,
This looks like it's supposed to be a MD link. Is that right?
I've just linked it to `nixpkgs/lib`


fricklerhandwerk (approved):
Oh yes, thanks!

PR #1070, opened 2024-10-31
Added extra context in the nested string interpolation example 

ellyxir:
the string + operator was not previously introduced and it wasn't clear that you can have more than just a variable within ${}
im doing a video series for learning the nix programming language and when i saw this part of the language tutorial, i had a hard time parsing what i was seeing. perhaps this could make it more clear. thank you!

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -803,8 +803,7 @@ error: cannot coerce an integer to a string
             5|
 ```
 
-Interpolated expressions can be arbitrarily nested.
-
+Interpolated expressions can be arbitrarily nested. Note that the `+` sign in the following example is the concatenation operator which is taking a string as its second argument.
 (This can become hard to read, and we recommend to avoid it in practice.)
 
 Example:
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1767,10 +1767,11 @@ For historical reasons, some of the functions in `pkgs.lib` are equivalent to [`
 So far we have only covered what we call *pure expressions*:
 declaring data and transforming it with functions.
 
-In practice, describing derivations requires observing the outside world.
+In practice, describing derivations requires observing the outside world. Recall that derivations are
+precise descriptions of how contents of existing files are used to derive new files.  We discuss [](derivations)futher in the tuturial.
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Yeah the tutorial is deliberately not going into depth about operators, because they're largely unsurprising -- except of course for string/path concatenation.

Thanks for pointing it out.

Instead of prefixing the note,
I suggest moving it into a collapsible "detailed explanation" block like it's done elsewhere, and link to the right place in the Nix manual for details on that particular operator.

Could you please do that? If not, I'll keep a reminder to finish it up.

ellyxir:
Yes I'd be happy to but won't get to it until Monday. Hope that's ok!

ellyxir:
I've moved it to a detailed explanation block and added a little more detail. Also linked to manual and the functional libraries section because standard operators and functions are mentioned there. Let me know if this is ok. Thank you!

source/tutorials/nix-language.md
```diff
@@ -822,6 +822,26 @@ in
 "no no no"
 ```
 
+:::{dropdown} Detailed explanation
+The `+` sign in the above expression is the concatenation operator. It is one of the many [built-in operators][operators] (`+`, `==`, `&&`, etc.) These operators can be used within `${ }`.
```
fricklerhandwerk, line 826:
```suggestion
Any Nix expression where the value can be represented as a string can be used within `${ }`.

The `+` sign in the above expression is the [string concatenation operator](https://nix.dev/manual/nix/latest/language/operators#string-concatenation), which takes two strings and produces a new string.

The expression in the example is deliberately confusing in order to demonstrate that arbitrarily nested string interpolations are possible, but tend to be hard to read.

It denotes a string that contains the interpolation of concatenating the value of `a` with a string that starts with a space and is followed by another interpolated string.
That second interpolated string is again the result of concatenating the value of `a` and yet another string that starts with a space and is followed by an interpolation of `a`.
```

fricklerhandwerk (approved):
Thanks a lot @ellyxir. This prompted me to fix [a small discoverability issue](https://github.com/NixOS/nix/pull/11793) in the Nix manual.

I expanded a bit on the explanation.

PR #1073, opened 2024-11-05
add `nix-serve-ng` as an alternative to set up HTTP caches

fricklerhandwerk:
@Gabriella439 I did consider making it the default in the tutorial instead of adding yet another alternative to the list. What do you think?


Gabriella439:
I think it's fine to mention it as an alternative

PR #1078, opened 2024-11-12
updated pinning-nixpkgs.md to have info on how to get the commits and channel names to use

ellyxir:
i went through first steps, then the language basics. at the end of language basics it had some next steps which took me back to declarative shell evironments. first time around, i didnt pay attention to specifics because i didn't know the language yet. this time around, i expected to understand it fully. 

i noticed the fetchTarball url https://github.com/NixOS/nixpkgs/tarball/nixos-24.05 and also the url in the detailed description https://github.com/NixOS/nixpkgs/tarball/2a601aafdc5605a5133a2ca506a34a3a73377247.

i had no idea how to create these. the detailed description has a link to ref-pinning-nixpkgs but this doesn't talk about how to find the commit to use. 
i thought it would be useful to add that information.

The next section towards-reproducibility-pinning-nixpkgs does discuss pinning to specific commits which would have been helpful to link to, but i think adding this to the manual would help more paths because even the towards-reproducibility-pinning-nixpkgs document references the manual.

### DIFF
--- a/source/reference/pinning-nixpkgs.md
+++ b/source/reference/pinning-nixpkgs.md
@@ -77,3 +77,8 @@ Specifying remote Nix expressions, such as the one provided by Nixpkgs, can be d
     pkgs = import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/nixos-22.11.tar.gz") {};
   in pkgs.stdenv.mkDerivation { ... }
   ```
+## Finding specific commits and releases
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Great find, thanks a lot! I continue in https://github.com/NixOS/nix.dev/pull/1080 since I can't modify the source branch for some reason.

PR #1087, opened 2024-11-26
Update callpackage.md

ellyxir:
removed extra "has", gramatical error perhaps


fricklerhandwerk (approved):
Thanks for reading carefully!

PR #1094, opened 2025-01-06
Add missing word in callpackage.md

agustinmista:
I believe there's a missing `set` at the end of this note for it to make sense:

```
The resulting attribute set is now recursive (by the keyword rec). That is, attribute values can refer to names from within the same attribute [set].
```


fricklerhandwerk (approved):
Oh yes, thanks!

PR #1099, opened 2025-01-22
update Nixpkgs and Nix releases

fricklerhandwerk:


roberth:
You might want to switch to https://nixos.org/manual/nixpkgs/stable/#tester-lycheeLinkCheck, which is not flaky

PR #1100, opened 2025-01-26
doc: fix broken links

dybucc:


source/tutorials/nix-language.md
```diff
@@ -315,8 +315,8 @@ Nix language data types *without functions* work just like their counterparts in
 - List elements are separated by white space.[^list-whitespace]
 :::
 
-[^attrnames]: Details: [Nix manual - attribute set](https://nix.dev/manual/nix/stable/language/values.html#attribute-set)
-[^list-whitespace]: Details: [Nix manual - list](https://nix.dev/manual/nix/stable/language/values.html#list)
+[^attrnames]: Details: [Nix manual - attribute set](https://nix.dev/manual/nix/2.24/language/syntax#attrs-literal)
+[^list-whitespace]: Details: [Nix manual - list](https://nix.dev/manual/nix/2.24/language/syntax#list-literal)
```
fricklerhandwerk, line 319:
```suggestion
[^attrnames]: Details: [Nix manual - attribute set](https://nix.dev/manual/nix/stable/language/syntax#attrs-literal)
[^list-whitespace]: Details: [Nix manual - list](https://nix.dev/manual/nix/stable/language/syntax#list-literal)
```

No need to pin the version (if we do it, we should do it globally, but that needs better tooling).

fricklerhandwerk (approved):
Thanks for catching it. The actual problem is that Netlify doesn't seem to do what it promises with regards to combining "pretty URLs" with redirects -- the redirects should work even for `.html` URLs but don't.

PR #1103, opened 2025-02-01
Clarify the lambda description

oleorhagen:
Update the wording of the lambda description in the function language description.

### DIFF
--- a/source/tutorials/nix-language.md
+++ b/source/tutorials/nix-language.md
@@ -1111,7 +1111,7 @@ Each of them is explained in the following, and here is an overview:
   { a, b, ... }@args: a + b + args.c
   ```
 
-Functions have no names.
+When functions have no names.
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Oh but "Functions have no names." is a full sentence. We could be slightly more specific and say that "Functions in the Nix language have no names.", but that always seemed redundant to me, since all of this is clearly about the Nix language.

source/tutorials/nix-language.md
```diff
@@ -1111,7 +1111,7 @@ Each of them is explained in the following, and here is an overview:
   { a, b, ... }@args: a + b + args.c
   ```
 
-Functions have no names.
+When functions have no names.
```
fricklerhandwerk, line 1114:
```suggestion
Functions in the Nix language have no names.
```

fricklerhandwerk (approved):

PR #1105, opened 2025-02-17
Update continuous-integration-github-actions.md

fzakaria:
Document how Nix MagicCache is now deprecated.

### DIFF
--- a/source/guides/recipes/continuous-integration-github-actions.md
+++ b/source/guides/recipes/continuous-integration-github-actions.md
@@ -74,6 +74,8 @@ you should see status checks appearing on commits and PRs.
 
 ## Caching builds using GitHub Actions Cache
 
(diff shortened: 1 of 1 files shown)


jfly (approved):

fzakaria:
Probably want to remove the whole section but I'll leave that for others to decide; at least more now it's good to recommend not to use it.

fricklerhandwerk:
I suggest we remove the article and point the URL to a tombstone with the explanation, as we do with release notes about breaking changes in Nixpkgs.

@fzakaria do you happen to have a source to support the claim about what GitHub may or may not want, so we can link to that?

fzakaria:
@fricklerhandwerk i'm not associated with DetSys; I just linked to their own article how they are deprecating MagicNix and the rationale behind it.

(Unfortunate because it truly was _magic_ and _free_)

I suggest landing this as is and following up with the removal etc..
For now, no one should be using it as it's been deprecated February 1st onwards.

fzakaria:
CC @grahamc

fricklerhandwerk:
Superseded by #1118

PR #1108, opened 2025-03-05
Add and simplify nix usage bullet points on the front page

wamirez:


fricklerhandwerk:
I'd prefer if we used the wording exactly [as in the original](https://github.com/ngi-nix/ngipkgs/blob/06dbb4f838d1e59cd035affcddd5f240f4cc7699/README.md#L30-L32), as that was painstakingly composed (thanks to @cleeyv's patience and diligence) and made a few things clear about what we're actually doing. We can link the terms Nix/Nixpkgs/NixOS to the glossary (which may need an update in its own right). And possibly those three bullet points should comprise the first section, while the existing points would be a more application-specific second section.

PR #1119, opened 2025-03-16
Add note about xz requirement for tarball in install-nix.md

thadguidry:
Since we're providing for download `.tar.xz` binary tarballs, we assume that systems have xz-utils or similar for handling that compression format.  We should make a note of this small requirement at the beginning of the Install Nix page.

Ref for WSL :
```
PS C:\Users\thadg> wsl
thad@MarsOrBust:/mnt/c/Users/thadg$ curl -L https://nixos.org/nix/install | sh -s -- --daemon
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
100  4267  100  4267    0     0    880      0  0:00:04  0:00:04 --:--:--  5216
sh: you do not have 'xz' installed, which I need to unpack the binary tarball
```

Fixed after running `sudo apt install xz-utils` for my Debian distribution with WSL

### DIFF
--- a/source/install-nix.md
+++ b/source/install-nix.md
@@ -1,6 +1,8 @@
 (install-nix)=
 
 # Install Nix
+Requirements:
+ - Prior to installation, you might need to first install `xz-utils` or similar for uncompressing the Nix binary tarball (`.tar.xz`) that will be downloaded via the scripts below.
 
 :::::{tab-set}
 


jfly (commented):
Thanks! I personally not convinced it's a good idea to duplicate the dependencies (which could change over time) here vs the source of truth ([the installer itself](https://github.com/NixOS/nix/blob/e4bda20918ad2af690c2e938211a7d362548e403/packaging/installer/install.in#L86)).

That said, waiting for a whole download just to get an error message is not very nice. How about we move the checks for missing dependencies to the top of the install script instead? Then people will find out about missing dependencies immediately.

I could also see putting a more generic note here saying something like "Note: the installer may error our if you're missing dependencies".

thadguidry:
Agree! Feel free to hack on this PR, or close it with another for improvements on the script as you suggest.

PR #1126, opened 2025-04-07
update Nix releases

roberth:
I believe this is all that's needed, but `nix-build .` output looked broken, containing placeholders like `@nix-latest@`. Is that normal?


fricklerhandwerk:
@roberth Yes, you need to enable a flag in `default.nix` (which is done in CI) because otherwise evaluation takes too long to be practical for iterating on contents.

PR #1127, opened 2025-04-09
simplify note on lazy attribute access

### LINKED ISSUE #1125: tutorial: confusing advice in Note
https://github.com/NixOS/nix.dev/blob/93b9f91ace1f0f34360ccf1bd44e4496207eca27/source/tutorials/packaging-existing-software.md?plain=1#L533-L537

Instead of being helpful, this note makes it more confusing. The first part says there will be a syntax error, while the second part says to do that so less is included.

fricklerhandwerk:
closes #1125

@joyously thanks for pointing out the issue. I removed the, IMO superfluous, part that hedges against readers introducing a syntax error. Does this make things clearer?

### DIFF
--- a/source/tutorials/packaging-existing-software.md
+++ b/source/tutorials/packaging-existing-software.md
@@ -504,7 +504,7 @@ To search derivations on the command line, use `nix-locate` from the [`nix-index
 
 ### Adding package sets as dependencies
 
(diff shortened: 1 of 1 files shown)


joyously:
Yes, as long as that's accurate, it sounds good. Thanks.

PR #1129, opened 2025-04-13
Add some documentation about the official nix formatter

jfly:
Ping @NixOS/nix-formatting, since I can't see to add the team as a reviewer on this PR.

We now have some decent tooling around this, but it's not very discoverable for people new to nix. We (the formatting team) believe `nix.dev` is a good place for this documentation to live.

I copied this text (with slight modifications) from `nix eval --raw nixpkgs#nixfmt-tree.meta.longDescription`.

### DIFF
--- a/source/guides/recipes/autoformatting.md
+++ b/source/guides/recipes/autoformatting.md
@@ -1,0 +1,56 @@
+(autoformatting)=
+# Autoformatting
+
+[`nixfmt`](https://github.com/NixOS/nixfmt) is the official Nix autoformatter.
+Official tooling currently does not use nixfmt out of the box. Subscribe to
+[NixOS/nix PR #11252](https://github.com/NixOS/nix/pull/11252) for updates on that effort.
+
+Because `nixfmt` doesn't support formatting whole directory trees, you need
+additional tooling such as `treefmt`. The `nixfmt-tree` package provides a
+`treefmt` pre-configured to run `nixfmt` on all nix files in your project. Just
+add it to your shell:
+
+```nix
+mkShell {
+  packages = [ pkgs.nixfmt-tree ];
+}
+```
+
+Note: this assumes you're project is in a git repository, and you wish to treat
+the entire repo as your project to be formatted.
+
+If you need to configure any [treefmt options], or enable formatting other
+(non-nix) files, you can use `treefmt.withConfig`:
+
+[treefmt options]: https://treefmt.com/latest/getting-started/configure/#global-options
+
+```nix
+pkgs.treefmt.withConfig {
+  runtimeInputs = [
+    pkgs.nixfmt-rfc-style
+    pkgs.ruff
+  ];
+
+  settings = {
+    # Customize detection of the root of the project.
+    tree-root-file = "flake.nix";
+
+    # Configure nixfmt for .nix files.
+    formatter.nixfmt = {
+      command = "nixfmt";
+      includes = [ "*.nix" ];
+    };
+
+    # And for .py file.
+    formatter.ruff = {
+      command = "ruff";
+      options = [ "format" ];
+      includes = [ "*.py" ];
+    };
+  };
+}
+```
+
+This can get a little tedious.
+[treefmt-nix](https://github.com/numtide/treefmt-nix) has a big library of
+preconfigured formatters, and provides a `check` derivation you can use in CI.
--- a/source/guides/recipes/index.md
+++ b/source/guides/recipes/index.md
@@ -11,4 +11,5 @@ Managing remote sources <./dependency-management.md>
 Python development environment <./python-environment.md>
 post-build-hook.md
 continuous-integration-github-actions.md
+autoformatting.md
 ```


MattSturgeon (commented):
Thanks for tackling this!

source/guides/recipes/autoformatting.md
```diff
@@ -0,0 +1,44 @@
...
+The easiest way to leverage `nixfmt` in your project is with the `nixfmt-tree`
+wrapper. Here's how:
+
+- For `nix fmt` to format all Nix files, add this to the `flake.nix` outputs:
+
+  ```nix
+  formatter.''${system} = nixpkgs.legacyPackages.''${system}.nixfmt-tree;
+  ```
+
+- The same can be done more efficiently with the `treefmt` command,
```
fricklerhandwerk, line 19: Please show exactly one way of doing things, and since official documentation is committed to stable features (and you already point out that it's more efficient) let's use the one where you add the package to your shell environment.

source/guides/recipes/autoformatting.md
```diff
@@ -0,0 +1,44 @@
...
+
+For more flexibility, you can customize this package using
+```nix
+nixfmt-tree.override {
+  settings = { /* treefmt config */ };
+  runtimePackages = [ /* List any formatters here */ ];
+}
+```
+
+Alternatively you can switch to the more fully-featured [treefmt-nix](https://github.com/numtide/treefmt-nix).
```
fricklerhandwerk, line 44: Rather than propose ever more alternative ways of doing things, I'd appreciate (and review) a guide to set up a pre-commit hook with treefmt-nix.

source/guides/recipes/autoformatting.md
MattSturgeon: > Please show exactly one way of doing things

I'm not sure I agree we should limit ourselves to only documenting a single approach to things.

`nix fmt` and `nix-shell --run treefmt` both have their own pros and cons, I'd expect users would want to be aware of each.

source/guides/recipes/autoformatting.md
MattSturgeon: While they both relate to formatting, and both produce a configured instance of treefmt, `pkgs.nixfmt-tree`, `pkgs.treefmt.withConfig`, and `treefmt-nix` all exist for different reasons and suit different use cases.

Users who _just_ want to configure formatting nix files in their project shouldn't need to download or configure treefmt-nix; there's a solution built into nixpkgs.

Users who want a more complex formatting setup, where they end up configuring several treefmt options, may benefit from using either treefmt-nix or `pkgs.treefmt.withConfig`.

source/guides/recipes/autoformatting.md
fricklerhandwerk: Realty is, we have to balance cognitive load against factual completeness. This place is about managing cognitive load, reference documentation is for completeness. I'd appreciate  discussion in reference documentation and issue threads on the costs and benefits of various approaches, but the motto of nix.dev is "getting things done" -- not "agonizing over the options". We're free to agonize in these behind the scenes exchanges, but we must not spill that on beginners.

source/guides/recipes/autoformatting.md
fricklerhandwerk: Right, and then we'd need to ask what's the purpose of any given article, and whether this information may be better placed elsewhere, such as is Nixpkgs contributor docs.

This sort of problem appears all of the time here, and sometimes one first needs to write an entirely guide to to write the guide one originally wanted to write. Otherwise it becomes just as unfocused and confusing as we still have an unfortunate reputation for.

source/guides/recipes/autoformatting.md
infinisil: Tbh I'm not sure if `nix fmt` has a benefit other than it being a shorter command, but we should recommend `direnv` anyways, in which case `treefmt` has the exact same length. I'd be fine with dropping `nix fmt`, though open about other benefits I'm missing!

infinisil:
From the formatting team meeting today:

- Discussed changes to make, there are legitimate reasons for keeping documentation of multiple options, but we can remove some of the options, in particular:
  - `nix fmt` -> Use a shell with `treefmt` instead
  - `nixfmt-tree.override` -> Use `treefmt.withConfig` instead
  - `treefmt --ci` -> Use `treefmt-nix` instead
- In general briefly mention when one would want to use the next step
- @jfly will also check about the possibility of (moving treefmt-nix into Nixpkgs](https://discourse.nixos.org/t/formatting-team-meeting-2025-03-18/61868/2) (potentially combining it with `treefmt.withConfig`), which would remove one more way of doing things
  - But this might hurt the development speed of treefmt-nix considerably

jfly (commented):
I've updated this PR with a second, more streamlined version of the docs.

source/guides/recipes/autoformatting.md
```diff
@@ -0,0 +1,55 @@
...
+
+```nix
+mkShell {
+  packages = [ pkgs.nixfmt-tree ];
+}
+```
+
+Note: this assumes you're working in a git repository with a [`.git/index` file
+at the
+root](https://github.com/NixOS/nixpkgs/blob/bbcdf758d61b6095693d00dea8a75061bf89b264/pkgs/by-name/ni/nixfmt-tree/package.nix#L33-L34).
```
MattSturgeon, line 21: :+1: for mentioning, however while it's currently correct, it will soon be outdated.

https://github.com/numtide/treefmt/pull/571 changed treefmt's default behaviour, so now it will dynamically ask `git` for the root of the workspace.

With that change, `tree-root-file` is no longer needed in most cases. Therefore, `nixfmt-tree` will remove its `tree-root-file = ".git/index"` definition once the next treefmt version is released.

It will _still_ be correct to say something like:

> this assumes you're project is in a git repository, and you wish to treat the entire repo as your project to be formatted

, however it will no longer be necessary to mention `.git/index` specifically or the implications that has for git worktrees.

source/guides/recipes/autoformatting.md
```diff
@@ -0,0 +1,55 @@
...
+
+If you need to change the root detection, or configure treefmt to format other (non-nix)
```
MattSturgeon, line 23: Maybe better to be more general and less specific regarding treefmt options?

```suggestion
If you need to configure any [treefmt options], or enable formatting other (non-nix)
```

We can also use this as an opportunity to link to the treefmt config docs:

```md
[treefmt options]: https://treefmt.com/latest/getting-started/configure/#global-options
```

source/guides/recipes/autoformatting.md
```diff
@@ -0,0 +1,55 @@
...
+      options = [ "format" ];
+      includes = [ "*.py" ];
+    };
+  };
+}
+```
+
+This can get a little tedious.
+[treefmt-nix](https://github.com/numtide/treefmt-nix) has a big library of
+preconfigured formatters, and provides a `check` derivation you can use in CI.
```
MattSturgeon, line 55: If we're mentioning the `check` derivation, should we include a (simple) example of using it?

If flakes are acceptable in nix.dev guides, maybe adding the drv to a flake's `checks` output would be a suitable example?

Such an example may be a good opportunity to "sneak in" defining `formatter` and `devshell` outputs, although not if that ends up making the example less simple.

source/guides/recipes/autoformatting.md
```diff
@@ -0,0 +1,55 @@
+(autoformatting)=
+# Autoformatting
+
+[`nixfmt`](https://github.com/NixOS/nixfmt) is the official Nix autoformatter.
+It is not yet wired up to any official tooling. Subscribe to
```
MattSturgeon, line 5: Very minor, but the wording feels backwards to me... The onus should be on the tooling to wire up; it is out of nixfmt's hands what tooling decides to use it.

I'm also struggling to phrase it, but maybe:

```suggestion
Official tooling has yet to be integrated with nixfmt. Subscribe to
```

or:

```suggestion
Official tooling currently doesn't use nixfmt out of the box. Subscribe to
```

MattSturgeon (commented):
LGTM overall. Minor comments, thoughts & suggestions:

source/guides/recipes/autoformatting.md
jfly: I agree it would be nice to include a simple example of using the `check` derivation, but that requires either using npins or flakes to pull in treefmt-nix. It gets messy kind of fast, and I'm lightly leaning towards "just send them to treefmt-nix to figure this out", as I've done here.

I'm very ambivalent on this, though. I'm going to leave this unresolved, perhaps someone with strong opinions will show up.

MattSturgeon (approved):
Looks good from my perspective :grin: 

Thanks @jfly

Issue #1135, opened 2025-05-04
Issue on page /tutorials/packaging-existing-software.html - nix-build output different from what is suggested by tutorial, potentially confusing for newcomers


prophetofxenu:
Hello, I was trying to follow along with the tutorial at [Packaging existing software with Nix](https://nix.dev/tutorials/packaging-existing-software) but got a different output when trying to discover the real hash to be used in hello.nix. I'm using NixOS 24.11 with Plasma 6.

From the tutorial, I am expecting an error message similar to this:
```
error:
...
       … while evaluating attribute 'src' of derivation 'hello'

         at /home/nix-user/hello.nix:9:3:

            8|
            9|   src = fetchzip {
             |   ^
           10|     url = "https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz";

       error: hash mismatch in file downloaded from 'https://ftp.gnu.org/gnu/hello/hello-2.12.1.tar.gz':
         specified: sha256:0000000000000000000000000000000000000000000000000000
         got:       sha256:0xw6cr5jgi1ir13q6apvrivwmmpr5j8vbymp0x6ll0kcv6366hnn
       error: 1 dependencies of derivation '/nix/store/8l961ay0q0ydfsgby0ngz6nmkchjqd50-hello-2.12.1.drv' failed to build
```

But the actual output is:
```
error: hash mismatch in fixed-output derivation '/nix/store/9v2iw49lql...-source.drv':
        specified: sha256-AAAAAAAA....=
              got: sha256-1kJjhtlsAkp....=
error: 1 dependencies of derivation '/nix/store/38bg5g9i...-hello-2.12.1.drv' failed to build
```
(full hashes omitted because I'm manually copying from a VM I don't have the clipboard setup in.)

There are a few things I found confusing in this:
1. The format of the hashes is different. The ones in the tutorial are in hex, while the real ones are base 64. While the steps are basically the same (copy the real hash to hello.nix), the tutorial omits the `sha256:` part. I had to include it (with a - instead to match the format) to get the derivation working.
2. The error messages are different as well (`hash mismatch in file downloaded from <url>` vs `hash mismatch in fixed-output derviation '/nix/store/9v2...-source.drv`). The first suggests the downloaded file had a different hash than what was expected, while the second suggests an intermediary artifact was what had the different hash.
[...]

PR #1136, opened 2025-05-04
Update nix-build output in packaging-existing-software.md, update port in CONTRIBUTING.md

prophetofxenu:
* Replace nix-build output in tutorial for packaging existing software. Previous output does not match current output, and uses different hash formats. Output and hash has been updated.
* Update the port in CONTRIBUTING.md the local server runs at from 5500 to 8080 which is what it was when I ran with no changes.

Resolves issues raised in #1135


jfly (approved):
:shipit:

PR #1137, opened 2025-05-05
Fix LaTeX build

holzingk:
With those changes, `make pdflatex` produces a well-formatted PDF document, see the [nixpkgs-cookbook.pdf](https://github.com/user-attachments/files/20040807/nixpkgs-cookbook.pdf) as an example I just built.

You might consider including the resulting PDF in the CI workflow and link it from the deployed website.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -256,11 +256,11 @@ html_sidebars = {
 
 latex_elements = {
     # The paper size ('letterpaper' or 'a4paper').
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Thanks! PR to add it to the site appreciated! Not sure who'd read it as a PDF but why not, maybe there's a place in the footer for it.

PR #1139, opened 2025-05-06
Publish PDF

holzingk:
I had to update pinned nixpkgs, as it did not build with TexLive 2022.

Some content annotations are not transformed very well in LaTex output. For example, there is a
[bug](https://github.com/executablebooks/sphinx-design/issues/107) in sphinx-design which prints dropdowns just as subsections without any visual highlighting. In general it is readable and better typeset compared to the PDF obtained by printing to PDF.

Also, there is now full texlive which is quite a huge dependency.


default.nix
```diff
@@ -16,21 +16,24 @@ let
...
-        sphinx
-        sphinx-book-theme
-        sphinx-copybutton
-        sphinx-design
-        sphinx-notfound-page
-        sphinx-sitemap
-        pkgs.perl
+      nativeBuildInputs = [
+        nix-dev-python-pkgs
+        pkgs.texlive.combined.scheme-full
```
Mic92, line 36: Indeed downloading multiple gigabyte for a single pdf on each ci run is too slow.
https://github.com/rgri/tex2nix/ could help to cut it down.

holzingk:
Rebased on current master and now using tex2nix to obtain a minified texlive distribution. Thanks @Mic92 for the great tool.

fricklerhandwerk (approved):
Pretty cool! Thanks a lot @holzingk

PR #1141, opened 2025-05-10
Package manager install on Debian, Ubuntu and Alpine

ngaro:
Documents installation with package manager on popular distributions. This is tested on the newest stable version of each of those:
- Debian 12
- Ubuntu 24.04
- Alpine 3.21

### DIFF
--- a/source/install-nix.md
+++ b/source/install-nix.md
@@ -14,7 +14,10 @@ Install Nix via the recommended [multi-user installation]:
 $ curl -L https://nixos.org/nix/install | sh -s -- --daemon
 ```
 
-On Arch Linux, you can alternatively [install Nix through `pacman`](https://wiki.archlinux.org/title/Nix#Installation).
+On most distributions you can also install it with the package manager:
+* On Arch Linux the different types of installation are documented [on the Arch Linux wiki](https://wiki.archlinux.org/title/Nix#Installation).
+* On Debian and Ubuntu (and distributions based on these like Mint, MX, ...) you can use `apt install nix`
+* On Alpine you can use `apk add nix`
 
 ::::
 


ngaro:
Don't merge this PR just yet. Although it installs fine, it doesn't configure all settings. The tutorials will not work if some extra config is not done.

source/install-nix.md
```diff
@@ -14,7 +14,10 @@ Install Nix via the recommended [multi-user installation]:
 $ curl -L https://nixos.org/nix/install | sh -s -- --daemon
 ```
 
-On Arch Linux, you can alternatively [install Nix through `pacman`](https://wiki.archlinux.org/title/Nix#Installation).
+On most distributions you can also install it with the package manager:
+* On Arch Linux the different types of installation are documented [on the Arch Linux wiki](https://wiki.archlinux.org/title/Nix#Installation).
+* On Debian and Ubuntu (and distributions based on these like Mint, MX, ...) you can use `apt install nix`
```
mweinelt, line 19: Also nix-setup-systemd for a more complete and useful install.

hansat299:
I would like to make some remarks about the part of this chapter which refers to installing nix on a linux distribution using the distribution package manager.

About Arch Linux:
Arch Linux uses the package manager pacman and refers to the package nix and you can install it using *pacman -S nix*. The other options are the nix supported install script, *Lix* and *Zero to Nix*. The first one is already mentioned above on the page. In my opinion I would not include the last two in this chapter because it is a bit out of scope.

The nix package is multi-user because it contains the installation of the nix-daemon.

About Debian/Ubuntu
The Debian nix package is for a local user. For multi-user installation, Debian offers the *nix-setup-systemd*, for which Debian will install the *nix* package as well

About Alpine:
The Alpine *nix* package is multi-user because it contains the nix-daemon installation. 

I would advise to add the minimum configuration for nix because the documents is also mend for beginners. You could format the information with a drop down. The configurations are:
- Add the user into the nix-users group ...command..
- Define the nix-channel ...command...

For more detailed installation notes, I advise to link this paragraph to https://wiki.nixos.org/wiki/Nix_Installation_Guide and visa versa. So that the page references reader to related observations outside nix.dev.

ngaro:
I am just starting with Nix. Odds are that I will make mistakes when my instructions become more detailed.
Can someone with experience complete this ?

fricklerhandwerk:
Thanks @ngaro @hansat299 for your contributions. I'm a bit hesitant to pick this up as is, because starting yet another collection of notes won't be an improvement over the pile of words in the wiki or anywhere else, but simply add another place with potential for divergent information that someone needs to keep up to date. What we'd actually need here is something along the lines as what @imincik added for NGIpkgs: Instructions ([example](https://ngi.nixos.org/project/Blink/#demo)) that are [automatically tested](https://github.com/ngi-nix/ngipkgs/blob/8b5e99e6ca4b9ff8e83ab9542911e1a65b390108/.github/workflows/test-demo.sh) against the [actual installers](https://github.com/ngi-nix/ngipkgs/blob/8b5e99e6ca4b9ff8e83ab9542911e1a65b390108/.github/workflows/test-demo.yaml) on those operating systems.

Now that code will probably be rather hard to read because it's built for a particular purpose, but the general idea would still apply here. It would need a bit of effort to adapt to nix.dev, and while upstreaming the work in NGIpkgs is definitely planned, it's not a primary goal this year and I don't want to make promises I can't keep.

If you're interested in having this, I'd strongly encourage you to work on top of that material and would gladly merge anything that has green tests (i.e. can be reviewed by eyeballing).

hansat299:
1) The initiative of adding  something to nix.dev are missing stones I met installing nix on Debian. I would like to share those to others for which they need to take care of if they install nix.
2) I agree partly with your hesitation, but do you want to wait one, two years to let the starter wait to discover or ask about missing stones? This would stop progress of nix.dev with something a starter or medium user needs.
3) I am not familiar with nix package building. If something is not working with a nix/nixos project. I would like to share solutions in a wiki or other form to help other users to proceed. So working with ngipkgs is a step too far for me.

hsjobeki:
@ngaro what is the state of this PR? Do you want to finish it? I think this is still worth adding. Had just this use case a couple days ago. Instructions in the form of code copy & paste would have been invaluable

PR #1142, opened 2025-05-24
nix: update to 2.29

Mic92:


Mic92:
Should be rather uncontroversial and we are lagging behind the release.

PR #1143, opened 2025-05-24
Automatically nix manual

Mic92:
based on https://github.com/NixOS/nix.dev/pull/1142


fricklerhandwerk:
Thanks! I don't see why we should run an update every day though, seems very wasteful. Before the automation I ran the command manually around each release and it was fine. We may as well do a workflow_dispatch and add a line to the Nix release instructions to hit that button on GitHub. Once we migrate Nixpkgs/NixOS manuals here, we can do the same for stable and only bump unstable every 24h.

Mic92:
Do you think so? It may just run a minute or so. Probably peanuts compared to what we do in nixpkgs. I got complaints about the missing manual in less then 24h.

fricklerhandwerk:
Just because we're likely burning embarrassing amounts of ancient phytoplankton with Nixpkgs doesn't mean that it's alright or that we have to do any more of that... We probably can't simply make it pull-based, but should be easy enough to add `gh repo clone nixos/nix.dev && cd nix.dev && gh workflow run ...` to the release script, no?

PR #1144, opened 2025-05-24
Bump nixpkgs

Mic92:


Mic92:
Wasn't aware of this modification. I did notice the API change but thought this was because of the version bump. Creating a symlink might also solve deduplicate the files. However not modifying those files also doesn't seem to be too bad.

fricklerhandwerk:
Symlink it is then. :)

PR #1147, opened 2025-05-30
Add hydra proxy header with well-known X-From

mweinelt:
This has now become required to bypass anubis on hydra.nixos.org.

Thanks, @adamcstephens, for preparing this.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -37,7 +37,5 @@
 
 /permalink/stub-ld /guides/faq#how-to-run-non-nix-executables 301
 /manual/nix /reference/nix-manual 200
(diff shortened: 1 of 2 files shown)


adamcstephens:
Just so we’re clear, I haven’t tested this. Unfortunately it doesn’t look like we get a preview deploy from opening this PR.

mweinelt:
Looks sufficiently like https://docs.netlify.com/routing/redirects/rewrites-proxies/#custom-headers-in-proxy-redirects and https://github.com/NixOS/nixos-homepage/blob/3b757750aecb1234332090cab831990c54baa617/core/public/netlify.toml#L6

jfly (approved):

mweinelt:
Looks like we lost two redirect rules.

PR #1148, opened 2025-05-31
Fix download link for PDF

holzingk:


fricklerhandwerk (approved):
Good catch. It only worked for the frontpage...

PR #1153, opened 2025-06-05
Fix search container margin causing unintentional element overlap

cowboycodr:
Before ([URL](https://nix.dev/search.html?q=nix#n)):
image

After (with proposed changes):
image

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -3,8 +3,8 @@
 }
 
 div.highlight-shell-session + div.highlight-none pre {
-  background-color: var(--pst-color-preformatted-text), 1 !important;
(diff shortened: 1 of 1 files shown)


cowboycodr:
Note: these screenshots were both tested on Firefox, and I haven't given the website a try with Chromium. So if anyone is using a Chromium-based browser let me know if you're getting different results.

fricklerhandwerk:
Thanks for the fix! Could you please remove the unrelated formatting changes from the diff?

PS: `nix-shell -p chromium`

cowboycodr:
I've removed the unrelated formatting changes, and also tested it on chromium. I think we're good to go!

PR #1154, opened 2025-06-06
no longer proxy to hydra from netlify

Mic92:
The infra team would like to have the freedom to have longer maintainace periods of hydra if necessary. Also we want to reduce the load on the service. This commit makes nix.dev mirror the manual as snapshots. During the build process we download the tarballs from github releases


nix/development-manual.json
```diff
@@ -0,0 +1,6 @@
+{
+  "version": "2025.01.06",
+  "sha256": "0000000000000000000000000000000000000000000000000000000000000000",
```
Mic92, line 3: This needs to be updated. Once we have the green light.

Mic92:
For context at the moment the development manual is also broken as a result of adding anubis in front of hydra.

Mic92:
cc @mweinelt

PR #1159, opened 2025-06-26
nixos-configuration-on-vm: show `initialHashedPassword`

raboof:
Instead of showing people `initialPassword` and then telling them not to use it, just show `initialHashedPassword` directly.

### DIFF
--- a/source/tutorials/nixos/nixos-configuration-on-vm.md
+++ b/source/tutorials/nixos/nixos-configuration-on-vm.md
@@ -52,11 +52,10 @@ To be able to log in, add the following lines to the returned attribute set:
   };
 ```
 
-Additionally, you need to specify a password for this user.
-For the purpose of demonstration only, you specify an insecure, plain text password by adding the `initialPassword` option to the user configuration:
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
The reason we did it that way it's because it's much easier to follow manually. More outside interactions we can't control here and more visual noise. Fresh learners are very sensitive to that, and a typo and another round of interactions risks breaking the flow and distracting even further. The tutorial is not supposed to smoothly transfer to production uses, we need many more steps to get people there.

raboof:
> The tutorial is not supposed to smoothly transfer to production uses, we need many more steps to get people there.

That's completely reasonable of course

> The reason we did it that way it's because it's much easier to follow manually. More outside interactions we can't control here and more visual noise. Fresh learners are very sensitive to that, and a typo and another round of interactions risks breaking the flow and distracting even further. 

That's true, though the current flow of "Here is how you X but actually don't do it that way" is also somewhat off-putting for new learners.

I don't have a good "best of both worlds" suggestion though, so I'll close this for now.

fricklerhandwerk:
Well we could emphasize just that, that it's for demonstration purposes in order to focus on the essentials, and that working with secrets properly is more involved.

PR #1170, opened 2025-09-02
Add recipe: Emit an evaluation warning only once

roberth:
_A technique to reduce warning noise, even for ones with a lot of prosaic context._

This isn't a particularly beginners' _topic_, but I don't see why we shouldn't have recipes for things you don't immediately run into. I hope it's quite accessible anyway.

Real world example
- https://github.com/NixOS/nixpkgs/pull/439189

### DIFF
--- a/source/guides/recipes/emit-a-warning-only-once.md
+++ b/source/guides/recipes/emit-a-warning-only-once.md
@@ -1,0 +1,205 @@
+(emit-a-warning-only-once)=
+# Emit an evaluation warning only once
+
+The Nix language provides a `warn` function that lets expression authors provide feedback to their code's callers/users.
+Since such warnings are tied to the flow of evaluation, it may happen that the warning you wish to add is trigger too many times.
+This document shows a technique to solve that problem, and documents its requirements and limitations.
+
+## Working principle
+
+Nix evaluates the root value of an expression file only once.
+We will exploit this property to emit fewer warnings.
+
+## Example
+
+Suppose we have the following little library:
+
+```nix
+{ lib, pkgs }:
+
+{
+  makeWidgetScript = { widgetType, useLegacyMethod ? true }:
+    pkgs.writeScriptBin "make-widget" ''
+      #!${pkgs.runtimeShell}
+      echo "Just made a" ${lib.escapeShellArg widgetType} ${lib.optionalString (!useLegacyMethod) "efficiently"}
+    '';
+}
+```
+
+Since we want widget production to be efficient, we want to phase out `useLegacyMethod`, but without causing unnecessary disruption, so we add a warning:
+
+```nix
+{ lib, pkgs }:
+
+{
+  makeWidgetScript = args@{ widgetType, useLegacyMethod ? true }:
+    lib.warnIf
+      (useLegacyMethod &&
+        # not explicitly provided by the caller
+        !(args ? useLegacyMethod))
+      ''
+        ACME Inc is in the process of phasing out the legacy widget production method,
+        but `makeWidgetScript` is still invoked without disabling the legacy method.
+        If this is intentional, and you are not yet ready to upgrade to the efficient method,
+        you may specify `useLegacyMethod = true;` to temporarily prolong your use during
+        the phase-out.
+      ''
+      (pkgs.writeScriptBin "make-widget" ''
+        #!${pkgs.runtimeShell}
+        echo "Just made a" ${lib.escapeShellArg widgetType} ${lib.optionalString (!useLegacyMethod) "efficiently"}
+      '');
+}
+```
+
+This will cause each `makeWidgetScript` invocation to emit a sizable message, every time it is invoked in the old way.
+
+Now let's use a top level scope to reduce our warnings to just one.
+
+We'll start by adding a `let` binding to the top of the file.
+The bindings in it will only be evaluated once, as the evaluator will cache the value it leads up to.
+
+```nix
+let
+  widgetScriptWarning = <...>; # TODO
+in
+{ lib, pkgs }:
+
+<...> # The rest of the code
+```
+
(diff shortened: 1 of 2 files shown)


source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
...
+      else x: x
+    )
+    pkgs.writeScriptBin "make-widget" ''
+      #!${pkgs.runtimeShell}
+      echo "Just made a" ${lib.escapeShellArg widgetType} ${lib.optionalString (!useLegacyMethod) "efficiently"}
+    '';
+}
+```
+
+## Best of both worlds
```
hsjobeki, line 129: I would drop this. Most people don't need this.

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
...
+# Emit an evaluation warning only once
+
+The Nix language provides a `warn` function that lets expression authors provide feedback to their code's callers/users.
+Since such warnings are tied to the flow of evaluation, it may happen that the warning you wish to add is trigger too many times.
+This document shows a technique to solve that problem, and documents its requirements and limitations.
+
+## Working principle
+
+Nix evaluates the root value of an expression file only once.
+We will exploit this property to emit fewer warnings.
```
hsjobeki, line 11: No future tense please. We are going to add a technical writing guide. Use present tense to describe how things happen.

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
+(emit-a-warning-only-once)=
+# Emit an evaluation warning only once
+
+The Nix language provides a `warn` function that lets expression authors provide feedback to their code's callers/users.
+Since such warnings are tied to the flow of evaluation, it may happen that the warning you wish to add is trigger too many times.
+This document shows a technique to solve that problem, and documents its requirements and limitations.
```
hsjobeki, line 6: Pure meta-commentary. I would cut it. Replace with a one-sentence statement of the technique itself or drop entirely. The document works without it.

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
+(emit-a-warning-only-once)=
+# Emit an evaluation warning only once
+
+The Nix language provides a `warn` function that lets expression authors provide feedback to their code's callers/users.
+Since such warnings are tied to the flow of evaluation, it may happen that the warning you wish to add is trigger too many times.
```
hsjobeki, line 5: Simplify and fix typo
```suggestion
Since such warnings are tied to the flow of evaluation the warning may fire more times than intended.
```

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
...
+```nix
+let
+  widgetScriptWarning = <...>; # TODO
+in
+{ lib, pkgs }:
+
+<...> # The rest of the code
+```
+
+Unfortunately, this means we don't have access to the `lib.warn` family of functions, but fortunately we can usually rely on `builtins.warn` nowadays (more on that later).
```
hsjobeki, line 70: Don't hedge a fact that has a precise answer.  builtins.warn was introduced in a precise version. This sentence adds vagueness.

As a user i would not know if i can use builtins.warn or not.

Sugggestion: drop the `nowadays (more on that later)` and state facts instead.

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Finding call sites
+
+If the evaluator does not produce an unrelated warning first, a quick method to find a call site is to pass `--show-trace --abort-on-warn` to your Nix command invocation.
+Otherwise, `--debugger-on-warn` is more suitable, as it lets you resume evaluation after warnings you ignore.
+
+## `builtins.warn` availability
+
+`builtins.warn` was introduced in Nix 2.23.
+Most users have upgraded far beyond 2.22, but if your code is in the upgrade path for users who may not have, use this polyfill:
```
hsjobeki, line 157: Just state the polyfill.

```suggestion
builtins.warn was introduced in Nix 2.23. If you support older versions, use this polyfill:
```

Most users have upgraded far beyond 2.22, statement Is not necessary/verified. 
Depends who your users are. Most of open source users? Most nixpkgs users? The sentence works without that claim.

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
...
+      (pkgs.writeScriptBin "make-widget" ''
+        #!${pkgs.runtimeShell}
+        echo "Just made a" ${lib.escapeShellArg widgetType} ${lib.optionalString (!useLegacyMethod) "efficiently"}
+      '');
+}
+```
+
+This will cause each `makeWidgetScript` invocation to emit a sizable message, every time it is invoked in the old way.
+
+Now let's use a top level scope to reduce our warnings to just one.
```
hsjobeki, line 56: Remove filler intro
```suggestion
Add a top-level ... 
```

source/guides/recipes/emit-a-warning-only-once.md
```diff
@@ -0,0 +1,205 @@
...
+
+```nix
+let
+  warn = builtins.warn or builtins.trace;
+  # ...
+in
+# ...
+```
+
+The `. or` operator will take care of the potentially missing `warn`.
```
hsjobeki, line 167: will take is future tense. `. or` typo
```suggestion
The `or` operator takes care of the potentially missing `warn`.
```

hsjobeki (changes requested):
Overall nice document that teaches the important bits and pieces.

The main issues are passive voice, meta-commentary, hedging language, and a few consistency problems.

Not sure if we need the "best of both worlds" section, since it seems a bit nieche

PR #1179, opened 2025-10-10
treewide: nixfmt

philiptaron:
Also, update to latest Nixpkgs to get nixfmt 1.1.


philiptaron:
Yeah, I just had 10 minutes. It might come in the future.

jfly:
> Yeah, I just had 10 minutes.

My hope is that someday this will only take 10 minutes to do :)

PR #1191, opened 2025-12-18
Improvement of Note on Fetchers

franzhusch:
I wanted to make this change to emphasize the difference between the builtin fetchers and the nixpkgs fetchers.


franzhusch:
Thanks for the feedback @hsjobeki. I have chosen to highlight the additional hash argument in the nixpkgs fetchers. That will also make them recognizable and is probably the most apparent difference for anyone not too deep into the details. Feel free to give me additional feedback if you think something could be better :)

franzhusch:
Your suggestions read much better @hsjobeki . I will take the tips and try to write better next time.

PR #1192, opened 2025-12-21
Link to Nixcademy lazy evaluation article

mightyiam:


mightyiam:
It's not a link to malware. Let people decide for themselves what amount of marketing they're willing to put up with in the content that they consume. Don't make this decision on their behalf. That's being neutral. That's my two cents.

fricklerhandwerk:
> Let people decide for themselves what amount of marketing they're willing to put up with in the content that they consume.

This is not how this here is intended to work. The [contributing guide clearly says to respect others cognitive resources](https://nix.dev/contributing/documentation/style-guide). And the reason it was written that way is that we have too much noise, not too few strings of characters that match certain patterns. I agree with what @hsjobeki wrote and suggest to close this.

I would also agree that there are discussions worth having with project leadership about finding forms for honoring prolific community members or giving businesses opportunities to advertise relevant services beyond the "commercial support" page, but let's not do it here.

thilobillerbeck:
Second this. I think we have to make clear separations on corporate involvement and community content. While I certainly agree people should reflect on their consumed information, we should fill resources dedicated to clear communication inside the community with such content only.

Additionally, mixing in content from parties with a financial incentive without a upfront proposed rule-set (e.g. if someone gives the foundation x amount of money, they can do a blog post), then complaints will come up. Other companies will ask themselves: "hey, we can do that as well" and over time things will get more overloaded and harder to navigate.

It's great what Jacek is doing with his work and his involvement, but I just think documentation is not the right context to link to from.

tfc:
> This is a new kind of external reference. Most of our current external references are purely educational.

With all due respect:

- ["Using Cachix you’ll never have to waste time building a derivation twice, and you’ll share built derivations with all your developers."](https://nix.dev/guides/recipes/continuous-integration-github-actions.html) direct advertisement of commercial nonfree cachix product
- ["Cachix: Nix binary cache as a service"](https://nix.dev/tutorials/nixos/binary-cache-setup) advertisement in the binary cache guide for nonfree commercial cachix product
- [Alternatives to setting up distributed builds with 100% FOSS: nixbuild.net, hercules-ci.org, garnix.io](https://nix.dev/tutorials/nixos/distributed-builds-setup#alternatives) direct advertisement of products that are all commercial and nonfree
- ["Take a look at Arion, a docker-compose wrapper with first-class support for Nix."](https://nix.dev/tutorials/nixos/building-and-running-docker-images) link to arion docs which are hosted by hercules-ci.org which directly advertise the commercial and nonfree hercules CI.
- [Provisioning-over-SSH article links to disko repo which is hosted by nix-community but still directly advertises for commercial numtide company openly calling to action to contact them for commercial help](https://nix.dev/tutorials/nixos/provisioning-remote-machines#provisioning-remote-machines-tutorial)
- ["Experimental does not mean unstable (Graham Christensen, September 2023)"](https://nix.dev/concepts/flakes) reference to determinate systems blog, commercial!
- [It may be more convenient to use a software like Etcher to flash the image to your SD card if you are on a system where it’s available."](https://nix.dev/tutorials/nixos/installing-nixos-on-a-raspberry-pi) reference to website that advertises to use the commercial balena platform with more text than actually describing ether on that page

My opinion is that these products and companies and the hard-working people behind them are great and provide value for the community with both their services and the free software & docs without forcing anyone to anything.

So does the free Nixcademy content.

If you reject this PR, please consider also removing those links. This is not fair.

hsjobeki:
i might have viewed this to restrictive indeed. The problem is that the topic was never properly discussed afaik.

From my understanding most external links in nix.dev (apart from those to the nix and nixpkgs manuals) should not be here. If they contribute something substantial, a PR of the same content would add it to the version control and give the community maintainence control over it. 

There is more effort involved to create a potentially bigger PR than adding a link. However accepting lots of those links, lead to documentation growing in blogs and other websites instead of in PRs to nix.dev. 

I think this is actually okay to merge after normal review since we have so many commercial links in place already. I might create an issue to discuss pruning them later.

mightyiam:
Thank you for looking into it, @hsjobeki. Is there anything I can do?

source/recommended-reading.md
```diff
@@ -22,6 +22,10 @@
 
   Introduction to writing derivations.
 
+- [What You Need to Know About Lazy Evaluation in Nix, Nixcademy](https://nixcademy.com/posts/what-you-need-to-know-about-laziness/) (Jacek Galowicz, 2024)
+
+  Understanding Nixpkgs overlays and the module system requires an understanding of lazy evaluation and fixed-point recursion. This guide explains all these, in order.
```
hsjobeki: I suggest removal of the second sentence. This doesn't really make sense here.

Add a dedicated chapter about laziness in the language section where it belongs, rather than front-loading this prerequisite claim in the reading list. Understanding Nixpkgs overlays and the module system requires understanding the language first, which is in https://nix.dev/tutorials/nix-language.html

hsjobeki (changes requested):

source/recommended-reading.md
mightyiam: I've struggled to understand what exactly I am asked to do. Here is my best guess, just pushed as an additional commit. Is this it, or was I supposed to keep some of that sentence? I read your suggestion about the addition of some chapter about laziness as what would be an ideal contribution with regards to this topic, but not as a practical request for this PR. Please correct me if I'm wrong.

hsjobeki:
@mightyiam could you squash your two commits?

tfc:
> Removal is planned as soon as we have a native chapter about lazyness

anakin

hsjobeki:
> > Removal is planned as soon as we have a native chapter about lazyness

I didn't intend to treat any external link differently. This is about the content that should be upstreamed which makes the link redundant. I want to get rid of all of the externals eventually, as mentioned. 

Meme appreciated - please continue this discussion in the issue I created:
https://github.com/NixOS/nix.dev/issues/1194

fricklerhandwerk:
I promised @tfc to clarify my comment, and took a deep breath trying to understand why three of us had a similarly pronounced reaction to the proposal. So here's a few thoughts for the archives.

As far as I can tell no one here  has a problem with *referencing* commercial material for informational purposes. The issue seems to be that we're trying to present it impartially, which raises some difficulties. For me this has always been a problem of improving the signal-to-noise-ratio and the effort of curation.

In that regard @tfc, the instances with Cachix, Arion, and Etcher clearly don't fit in. When owned by Domen it may or may not have been legitimate; now it's unmarked advertisement – editorial malpractice. The reason it's still there is that no one got to rework that material since we inherited it. The articles happen to be on those topics noone cared about maintaining so far. I'd be in favor of removing those mentions.

The mentions of (commercial) *alternatives* for upstream tools are okay, I added some of them myself. There it should be quite clear from context that there's no particular value judgement, merely their existence is noted.

Back in the day we had lengthy discussions about whether or how to recommend or introduce more or less third-party tools, and settled on that anything from nix-community is okay. How those projects present themselves is not something we here can reasonably (or, IMHO, should try to) control though, and attempts to establish a more formal relationship between nix-community and the official project evaporated. I don't have an satisfying answer to that aspect, but also I consider it a side issue of questionable value to debate here. Even the broader issue of which tools to recommend is much more tricky than may first appear, because it's really about support status and system-wide product design, and that gets into politics very quickly.


As I've noted, there's no policy for ads. When I was actively maintaining the material, the discriminator for endorsement was whether we (the loose group of contributors) found that it was distinctively useful information for a given topic and whether we could "support" the use case presented in the external material, in the sense that we could answer questions about it and confidently evaluate the merit of subsequent pull requests. I understand the desire for predictability by commercial actors though, so having a bit of a stance on those questions would be good. So thanks @hjsobeki for opening an issue for reference, even if I won't hold my breath for it to be truly resolved any time soon.

Hopefully this background information helps building empathy with who's trying to take care of things here. My personal reaction was grounded in the feeling "yet more stuff to evaluate whether it fits in". My ideal for "official" material was always to be a sort of synthesis of what the community figured out over time. It was based on the observation that there's just too many variations of essentially the same things, which are hard to navigate for beginners because there's little indication of what can be relied on. And commercial actors are in a sense adversaries in such a situation: they can produce arbitrary amounts of material of arbitrary usefulness to the upstream cause.

I admit we didn't find a cheap way to realize that vision, some way that wouldn't devolve to the pre-existing wiki way of people piling up all sorts of semi-redundant, often half-baked bits and pieces. I suspect there is no cheap way for what I consider high quality documentation.

That said, as I'm not available to do serious work here at this time anyway, my opinion shouldn't matter beyond sharing experience. Which is why I hid my initial comment and will hide this one as well. Thanks a lot again @hsjobeki for taking care of things!

PR #1197, opened 2026-01-06
clarify flakes as erring to implicit management of recursive dependencies

KiaraGrouwstra:
while i'm not sure this is the right place for this, i think defaulting to implicit management of recursive dependencies is currently a feature that sets it apart from competing dependency management systems for Nix. in my understanding, that makes this a relevant consideration in whether one might choose to adopt them.

i.e., if one wants a better `apt` offering packages that (individually) span multiple nixy repositories, flakes may well be the closest thing they can find right now.

on the other hand, if one is coding a project themselves, i.e. may want more control over their (recursive) dependencies, then this distinguishing feature in itself may be sufficient reason to opt to pass up on them, at least for the purpose of a default mode of dependency management.

### DIFF
--- a/source/concepts/flakes.md
+++ b/source/concepts/flakes.md
@@ -88,7 +88,7 @@ This overview can help get what you need from flakes while preserving compatibil
 
 [dependency management]: https://nix.dev/guides/recipes/dependency-management.html
 
-[^flake-inputs]: Nix repositories offering only flake entrypoints may be import using [`flake-inputs`].
+[^flake-inputs]: Nix repositories offering only flake entrypoints may be imported using [`flake-inputs`].
 
 ### Discoverability
 


philiptaron:
I'm not sure that I understand the title or the proposed wording. Flake inputs do indeed allow users to inject their own inputs, but as I see it, there's an unpaved road that leads to making use of that input in the derivations *from Nixpkgs* that they're likely trying to use. That usually ends in a long and somewhat complicated overlay. Am I misunderstanding here?

KiaraGrouwstra:
what i tried to say here is, while flake `inputs` allow overriding versions of recursive dependencies using the `follows` mechanism, the default, when one were not to explicitly override a version, is to use versions specified by the parent dependency.

this differs from how dependency management would work traditionally, e.g. how package derivations would specify all their dependencies as module inputs, making the caller specify inputs explicitly. that explicit approach seems facilitated by the traditional dependency pinning mechanisms such as `npins`.

i hope that helps a bit - feel free to suggest changing the wording if i wasn't communicating that intent well.

philiptaron:
So "[1000 instances of Nixpkgs](https://zimbatm.com/notes/1000-instances-of-nixpkgs)" as a positive reference rather than a negative one?

KiaraGrouwstra:
depending on your use-case, it _could_ be.

i think most current nix(os) users are rather in the category of coding up declarative set-ups rather than using nix imperatively, so i believe for most of us in the community '1000 instances of nixpkgs' is not what we want.

i agree that, given that, this might not be the right place to clarify flakes offer implicit-leaning dependency management.
would you see a place that might be better perhaps?

KiaraGrouwstra:
in retrospect, i'd consider rewording the 'why controversial' section into a broader 'why you might consider alternatives' one. i feel like controversy perhaps reflected the time of writing, but doesn't necessarily seem like the reasons one should base their decisions off of today.

philiptaron:
Yeah, this whole section (before the proposed change) reads like it was written some years ago, which it likely was.

I'm game for some broader reworking. Want to propose that?

KiaraGrouwstra:
sure - let me try and draft something up

hsjobeki (changes requested):

KiaraGrouwstra:
i tried expanding this further now - feedback welcome

KiaraGrouwstra:
@normalcea thanks for your swift feedback! i've tried to address them now.

source/concepts/flakes.md
```diff
@@ -1,112 +1,219 @@
...
-  Improvements to experimental features have low priority.
-- The [Nix documentation team](https://nixos.org/community/teams/documentation.html) focuses on improving documentation and learning materials for stable features and common principles.
-  When using flakes, you will have to rely more heavily on user-to-user support, third-party documentation, and the source code.
+[various problems]: https://wiki.lix.systems/books/lix-contributors/page/flakes-feature-freeze#bkmrk-design-issues-of-fla
+[open design questions around the `nix` command line interface]: https://github.com/NixOS/nix/issues?q=is%3Aissue+is%3Aopen+label%3Anew-cli+sort%3Areactions-%2B1-desc
+[problems with the implementation]: https://github.com/NixOS/nix/issues?q=is%3Aissue+is%3Aopen+label%3Aflakes+sort%3Areactions-%2B1-desc
+[consolidated the featureset]: https://wiki.lix.systems/books/lix-contributors/page/flake-stabilisation-proposal
```
philiptaron: I took a stab at turning these bullet points into prose. I think I captured what you were going for, but this one might be more "me" than the others.

```suggestion
Flakes were criticized [during the RFC process](https://github.com/NixOS/rfcs/pull/49) for trying to solve too many problems at once and at the wrong abstraction layer. Despite unresolved design concerns, the implementation was merged before the RFC was accepted, then the [RFC was withdrawn](https://github.com/NixOS/rfcs/pull/49#issuecomment-659379115) in order to re-assess core design concerns. There is no timeline to conclude the experiment.

This created a catch-22: flakes became widely adopted while fundamental questions remained open. [Various questions](https://wiki.lix.systems/books/lix-contributors/page/flakes-feature-freeze#bkmrk-design-issues-of-fla) around versioning, composability, and cross-compilation persist. [Problems with the implementation](https://github.com/NixOS/nix/issues?q=is%3Aissue+is%3Aopen+label%3Aflakes+sort%3Areactions-%2B1-desc) remain. Yet changing the design now risks breaking the many projects that depend on it.

The community remains divided. Nixpkgs, the largest Nix project, does not use flakes in its internal tooling. Different groups have taken divergent approaches: Determinate Systems unilaterally declared flakes stable (with proprietary extensions), while the Lix fork has [frozen](https://wiki.lix.systems/books/lix-contributors/page/flake-stabilisation-proposal) the current implementation into a de-facto "v1". This fragmentation of underlying implementations undermines the unified interface that motivated flakes in the first place.

Despite all this, flakes through the v3 CLI remain a useful and relatively ergonomic tool for composing code across disparate ecosystems. As evidenced by adoption, they are the default choice for sharing Nix code in a federated or distributed manner.
```

source/concepts/flakes.md
```diff
@@ -1,112 +1,219 @@
...
+[parameterized]: https://github.com/NixOS/nix/issues/2861
+[downgrade]: https://github.com/NixOS/nix/issues/3843
+[utility functions]: https://github.com/numtide/flake-utils
+
+Alternatives:
+
+- Use flakes as thin wrappers over existing Nix code, so code can be used in both ways.
+- Settle for exposing traditional Nix modules, and let flake users import the project with `flake = false;`.
+
+[`import`]: https://nix.dev/tutorials/nix-language#import
```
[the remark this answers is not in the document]
KiaraGrouwstra: while i realized this is based on my phrasing, i guess the wording here reminds me of some nuances to this:

> Adding a `flake.nix` exposes your project to the flake ecosystem. Users can reference it in their flake inputs

- non-flake projects can in fact be sourced, the difference being that for projects with a `flake.nix`, you'd get the flake's `outputs` over just the project path, while for flake projects additionally using `inputs` such (recursive) dependencies would be sourced as well.

> build it with `nix build`, run it with `nix run`

- the v3 cli (`nix build`, `nix run`) would support loading arbitrary nix files using `-f <path-to-nix-file>`

> or add it to a registry.

rechecking this, i think what you can add is not confined to locations containing a `flake.nix`. (adding references to the relevant JSON configuration by NixOS's `nix.registry` further seems to require no experimental features - tho actually making use of the registry will.)

> blueprint

hm, this is making me realize my split into four sequential-ish sections was maybe still less granular than the actual considerations involved.

for the last section (flake-only) i had been thinking essentially of the dendritic pattern, which makes modules very hard to consume [through](https://github.com/fricklerhandwerk/flake-inputs/) anything but the flake entrypoint.

blueprint tho uses modules with flakes' module arg `inputs`, which is not nearly as bad as the dendritic pattern for vanilla consumers, but does still make for a gap with vanilla modules.

(as an additional consideration there, that technically isn't even exactly about adding another section in-between dependency management and flake-only, in the sense the `self` parameter would be available before use of flakes' `inputs` for dependency management as well.)

maybe this also feels a bit awkward in the sense that for compatibility blueprint would be closer to such later sections, while at the same time i understand that it's on-topic to the `system` section you added it for here.

> Flake users can also import non-flake projects directly with `flake = false` in their inputs.

i'm probably nitpicking here, but i feel the new phrasing here could be misread such as to imply that being able to import both types of entrypoints is specific to opting in to flakes (to e.g. the point of enabling the experimental features). given e.g. [`flake-inputs`](https://github.com/fricklerhandwerk/flake-inputs/) i think the potential for such a reading could end up misleading, regardless of intent.

KiaraGrouwstra:
@philiptaron 

> I let my fingers fly this morning. I'm mostly just trying to re-write in order to understand.

thank you!

i tried to kinda add some of my considerations now - hopefully that might clarify a bit over the original wording.

as some background, i used flakes for a few years, then last year redid my nixos config following [jade's blog post](https://jade.fyi/blog/pinning-nixos-with-npins/) using npins.
going by the tool itself, that switch felt like a relief, not having to deal with exploding dependencies anymore. at the same time tho, documentation-wise it's felt a bit more uphill, as there have definitely been a few libraries that considered flakes the default and only option to the point you would end up having to search more to find what you need - or worse, were just [harder](https://github.com/sodiboo/niri-flake/issues/1089) without.

for work we struggled with that more so still, having ended up with a project that partly used `flake-parts`, had a [dependency](https://github.com/nixops4/nixops4/issues/113) using that, while also using `npins`, making for a situation where we had mixed versions of nixpkgs, while the dependency made it hard to get out. [`flake-inputs`](https://github.com/fricklerhandwerk/flake-inputs/) and a [way to generate (`flake-parts`-enabled) flakes without experimental features](https://git.fediversity.eu/fediversity/fediversity/src/commit/24e4c01806b3ccf8057aeb32cf71209df6407180/mkFlake.nix) spun out of that, while escaping from all that is [still](https://git.fediversity.eu/fediversity/fediversity/pulls/583) looking [uphill](https://git.fediversity.eu/fediversity/fediversity/pulls/553), months later.

conversely, i think doing vanilla would not make it hard to pick up flakes: in fact, it would rather get one closer to the thin-wrapper type of combination that i think could actually make for the golden mean here, getting one the compatibility with the schema for discoverability, as well as with the v3 cli for caching in CI, potentially without having to get into the mess of exploding dependencies.

i kind of hope we could move closer to that golden mean collectively - and if i've sounded bearish in my writing, that's kind of the experiences underlying that.

source/concepts/flakes.md
```diff
@@ -1,112 +1,219 @@
...
+[`nix.registry`]: https://search.nixos.org/options?channel=unstable&show=nix.registry&query=registry
+[eta-reduction]: https://wiki.haskell.org/Eta_conversion
 
-Flakes add the following behavior to Nix:
+## Should I use flakes in my project?
 
-1. A `flake.nix` file enforces a [schema], where:
-   - Other flakes are referenced as dependencies providing {term}`Nix language` code or other files.
-   - The values produced by the {term}`Nix expression`s in `flake.nix` are structured according to pre-defined use cases.
+You have to judge for yourself based on your needs.
```
normalcea: Thinking this over, I believe a stronger position should be made to discourage flakes, or even to outright reject them. 

It has ultimately, in my view, been a fatal error to allow community adoption of flakes while so many design decisions have been left up to the table (either by lack of development power or deadlocked on discussion). Flakes are an opaque interface that do not follow any established Nix rules and may very well never see any roadmap for rehabilitation. As already added, we are very much stuck on a flakes v1, a v1 of brittle components. 

We should instead point people to follow the nixpkgs way of doing things and if nixpkgs can distribute a project that builds hundreds of thousands of packages, has a standard library of hundreds of functions, implement a module system and an operating system from those modules all while never touching flakes then maybe flakes weren't the solution we needed.

In essence, would it be a good idea to present flakes as an experimental extension format on-top of Nix rather than at equal footing with stable Nix? Hobbyists on spare systems can follow flakes development but I'd imagine that anyone above that need wouldn't find an "experimental extension format" on what should be the last word of build systems (All build-systems should feed into a stable Nix that is uncomplicated and uninteresting, flakes are not that).

source/concepts/flakes.md
```diff
@@ -1,112 +1,229 @@
...
+  This makes the user responsible for versions used, solving drawbacks of flakes' implicit model.
+  - If dependencies have only flake entrypoints, if needed load their flakes using library [`flake-inputs`].
+  - Allow others to change your dependencies by having them in Nix entrypoints' module parameters _and_ exports:
+
+  ```nix
+  {
+    sources ? import ./npins,
+  }:
+  {
+    # ...
```
hsjobeki:
```suggestion

```

remove this. It's not really adding information.

We can trust the reader to understand that the file might be bigger.

source/concepts/flakes.md
```diff
@@ -1,117 +1,320 @@
...
+  };
+}
+```
+
+Flakes are aimed at sharing Nix code.
+They make it easy to build programs with the same version.
+
+Flake commands access flakes using [references] to local (e.g. `.`) or remote (e.g. `github:NixOS/nixpkgs`) project directories.
+
+[`outputs`] include various [built-in types], but can be [extended].
```
hsjobeki: This should follow-up immediately after the flake.nix file example

source/concepts/flakes.md
```diff
@@ -1,117 +1,320 @@
...
+You can find an overview of these on the [wiki].
+
+[`inputs`] let you declare dependencies.
+
+Nix creates a [`flake.lock`] to pin dependencies.
```
hsjobeki: When does it create the lock?

```suggestion
Nix creates a [`flake.lock`] to pin dependencies once you run a [nix-command].
```

source/concepts/flakes.md
```diff
@@ -1,117 +1,320 @@
...
+- Make it easy to reproduce published software, following the versions they used.
+- You can override recursive inputs.
+
+Cons:
+
+- Flakes by default follow versions from dependencies' `flake.lock`, so if you don't override these with `follows`, you may get:
+  - multiple versions of the same dependency.
+  - outdated dependencies, if their versions are not actively updated.
+- Dependencies are fetched eagerly, loading dependencies you may not use.
+- Dependencies managed by flake inputs are hard to override if you don't also use flakes.
```
hsjobeki: Addition:

- If your flake is used as a library you need to add follows statements for all recursive inputs. Otherwise downstream consumers cannot add their follows on your indirect inputs.

source/concepts/flakes.md
```diff
@@ -1,117 +1,320 @@
...
+
+Alternatives:
+
+- Handle dependencies with [`npins`].
```
hsjobeki: I would only point to the npins guide here. Not repeating any examples or pro-cons. 
This is already maintained in a different place and needs to be kept in sync now.

hsjobeki (changes requested):
Left some minor improvements.
After which i think this is good to merge.

KiaraGrouwstra:
@hsjobeki thank you for your comments! i rewrote it a bit to incorporate these now.

philiptaron:
Congrats on getting the words out. Sorry I wasn't more helpful.

PR #1198, opened 2026-01-06
explain managing nixos configurations by npins

KiaraGrouwstra:
inspired by [jade's article](https://jade.fyi/blog/pinning-nixos-with-npins/). they further mentioned setting up flake commands (nix v3 cli) to use `npins`'s `nixpkgs` pin, which nowadays translates to setting [`nixpkgs.flake.source`](https://search.nixos.org/options?channel=unstable&show=nixpkgs.flake.source&query=nixpkgs.flake.source).


philiptaron:
Yeah, there are a bunch of ways to do it. This one works.

PR #1201, opened 2026-01-11
callpackage: mention `import`

KiaraGrouwstra:
(this could link to https://nix.dev/tutorials/nix-language#import, but i'm not sure of the best way to do this)


hsjobeki:
I think this can be inferred by the reader. If its unclear, we should add a sentence what is `callPackage` specfic. 

The sentence about `callPackage` feels unclear, i would rather improve that, than adding another one.

KiaraGrouwstra:
sounds good, i can close in favor of alternatives

PR #1202, opened 2026-01-13
describe using npins dependencies in v3 cli commands using the flake registry

KiaraGrouwstra:


KiaraGrouwstra:
i dunno how github's formal 'request review' things works, so /cc @philiptaron @hsjobeki :no_mouth: :pray:

Issue #1184, opened 2025-10-23
search results link to nix-dev.pdf


raboof:
When I go to https://nix.dev/search.html?q=foo# on Firefox everything that's not a link actually links to nix-dev.pdf

**Willing to help?**

Not right now I'm afraid

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

Haschwalth00B:
Can I?

raboof:
Please do! ;)

Issue #1167, opened 2025-08-05
NixOS install guide for Raspberry Pi misses step in installation using wifi


PieterPel:
**Observations**

I recently followed [this guide](https://nix.dev/tutorials/nixos/installing-nixos-on-a-raspberry-pi.html) (with the most recent hydra image) on installing NixOS on a Raspberry Pi 400. Setting up a wireless connection at the 'Getting internet connection' step did not actually connect to the internet yet. I had to first assign a static ip address to my wlan0 using the following commands:

```bash
sudo ip addr add <ipaddress> dev wlan0
sudo ip route add default via <ipaddress>
```

This allowed me to set up a nix shell to assign a regular ip address with `dhcpcd`

```bash
nix-shell -p dhcpcd
sudo dhcpcd wlan0
```

And the guide could be completed.

Also, the guide currently mentions to use `iwconfig` to find the network interface, but this program isn't available when first booting.

**Problem**

The guide misses a step to assign an ip address to the network interface and incorrectly assumes that `iwconfig` is available. 

**Approaches**

A step needs to be inserted that explains how to assign an ip address to the wireless interface. It could be that this issue doesn't show up everywhere, but it would at least be good to mention it in the guide. Replacing `iwconfig` with `ip link` would also be an improvement

**Willing to help?**

Of course

**Priorities**

Add :+1: to [issues you find important](https://github.com/NixOS/nix.dev/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

Haschwalth00B:
Can I work on it?

fricklerhandwerk:
Hi and welcome! No need to ask for permission, just open pull requests with changes. From experience, the smaller and less impactful the change, and the easier to test it for correctness (no one here is an expert on all subjects) the more likely it gets attention and more quickly it can get merged.

PR #1208, opened 2026-02-07
Updated WiFi setup steps for installation on Raspberry pi

Haschwalth00B:
This PR addresses two issues:

#1167: Adds missing step for assigning IP address to wireless interface during NixOS Raspberry Pi installation, and replaces unavailable iwconfig command with ip link
#1184: Fixes search results page where non-link elements incorrectly link to nix-dev.pdf

### DIFF
--- a/nix/nix-versions.json
+++ b/nix/nix-versions.json
@@ -1,28 +1,33 @@
 {
+  "2.34": {
+    "url": "https://github.com/nixos/nix/tarball/7ee0a95261253665e064a197272f91ca0e97c7b1",
+    "version": "2.34.1",
+    "sha256": "14mywvdf19jwpsa67g05kxis69g9hljrvfp1mmgqr5p2g4yvr3z1"
+  },
   "2.33": {
-    "url": "https://github.com/nixos/nix/tarball/377fc95bce8714e54883f40f7ac7edbfc63801a4",
-    "version": "2.33.3",
-    "sha256": "0jm8wdg6iprhpja35v80cwi17nwdbklf71caq75y7d2rxzhimj6q"
+    "url": "https://github.com/nixos/nix/tarball/db62d8d931d626856243c37ec1e7f4ce88a39412",
+    "version": "2.33.4",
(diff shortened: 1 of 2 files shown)


source/tutorials/nixos/installing-nixos-on-a-raspberry-pi.md
```diff
@@ -72,13 +72,27 @@ Run `sudo -i` to get a root shell for the rest of the tutorial.
...
-Once you see in your terminal that connection is established, run `host nixos.org` to check that the DNS resolves correctly.
+If `dhcpcd` is not available, load it with:
+```shell-session
+# nix-shell -p dhcpcd
+# dhcpcd wlan0
+```
+
+After a few seconds, verify the connection works by running `host nixos.org` to check that DNS resolves correctly.
+
+If DNS resolution fails, you may need to manually add a nameserver:
```
fricklerhandwerk: Hm, mutating files is not good taste when you have Nix. Also why would we add the google NS of all things? The problem of explaining how to figure out your network connection explodes pretty quickly if you don't know what learners already know. One way to avoid that issue is to state the prerequisites explicitly, e.g. (I put no thought into this, so just for illustration):
- A working DHCP server on the network your Raspberry Pi is connected to
- Experience debugging network issues

Because then you can just assume it will work or resort to people's pre-existing experience to fix it on their own.

Haschwalth00B:
@fricklerhandwerk 
Thanks for the review and guidance, really grateful for it,  I'll fix it and also create separate PRs for thise issues, and also remove the venv folder from the .gitignore, I added it cuz i created a virtual environment to build the html file

And thanks again for the guidance

Haschwalth00B:
@fricklerhandwerk I've made the changes you suggested. Hope I got it right

Haschwalth00B:
@fricklerhandwerk hey, I've submitted my GSoC proposal for the nixpkgs PR review, would you be open to being a mentor for my project?

fricklerhandwerk:
@Haschwalth00B please use the appropriate venues for this (there's a Matrix channel), but short answer is no since I'm already mentoring a project, and Nixpkgs review is not my area of expertise.

PR #1209, opened 2026-02-09
flakes: clarify flake-compat/flake-inputs

KiaraGrouwstra:
Update the /concepts/flakes page to note flake-compat and flake-inputs both work from the publisher's as well as the consumer's perspective.

Note that while flake-compat is documented better, flake-compat seems to still lack support for
[overrides](https://git.lix.systems/lix-project/flake-compat/issues/82).

### DIFF
--- a/source/concepts/flakes.md
+++ b/source/concepts/flakes.md
@@ -83,12 +83,12 @@ Flakes are an experimental extension format with outstanding issues.
 Its functionality can generally be achieved without them as well.
 
 If you need to run existing software that already used flakes, or want to contribute to their development, feel free to use them.
-If you want to write Nix code yourself, consider also our guide on [dependency management].[^flake-inputs]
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra:
CI complains 429 (Too Many Requests)

hsjobeki:
@KiaraGrouwstra What is the state of this PR? I think we should simply drop the reference to "flake-inputs" ?

KiaraGrouwstra:
technically pending on review of https://git.lix.systems/lix-project/flake-compat/pulls/87 (~~if not also https://git.lix.systems/lix-project/flake-compat/issues/85~~)

KiaraGrouwstra:
heck, might as well not await that `flake-compat` PR given it seems unlikely to land (recursive overrides need concensus on design)

Issue #1212, opened 2026-02-16
`/concepts/flakes`: add tables comparing alternatives


KiaraGrouwstra:
currently, `/concepts/flakes` lists pros/cons/alternatives for some stages of flake usage.
while some alternatives may specifically address specific cons without being mutually exclusive with other options, overall this current structure leaves a lot of information out, essentially leaving readers guessing which alternatives resolves which drawbacks (and perhaps at what cost of their own).

a way to address this would be to for each of these sections add proper comparison tables, so that readers would have a better overview of the different options available.

c.f. [nixtamal's comparison](https://nixtamal.toast.al/#comparison)

PR #1213, opened 2026-02-16
/concepts/flakes: add dependency management alternative `flake = false;`

KiaraGrouwstra:
clarifies that specifying `flake = false;` on flake inputs can help resolve some of its drawbacks - in this case (if used consistently) getting [duplicate versions](https://zimbatm.com/notes/1000-instances-of-nixpkgs) and having to set `follows`. (this would not solve eager fetching, which would instead be addressed by [lazy trees](https://github.com/NixOS/nix/pull/13225).)

to make it more explicit what alternatives address what, a proper approach might be #1212.

### DIFF
--- a/source/concepts/flakes.md
+++ b/source/concepts/flakes.md
@@ -254,6 +254,7 @@ Alternatives:
 
 - Handle dependencies with [`npins`].
 - Handle dependencies inline using functions such as the [fetchers] or [`builtins.fetchTree`].
+- Use `inputs.<name>.flake = false;`
 
 [fetchers]: https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers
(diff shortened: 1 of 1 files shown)


friedow (changes requested):
Explaining the usage of `flake = false` is very valuable to the flakes documentation. I think just mentioning this as an alternative under dependency management does not explain the feature well enough though.

I'd suggest explaining what `flake = false` does in an own paragraph, or two :). Would you want to change this PR this way?

hsjobeki:
I'd add a heading just for this concept and add a reference here to that heading. So this section stays focused.
But we can add this in a follow up. This PR on its own makes sense to add.

friedow (approved):

PR #1214, opened 2026-02-16
fix github star button not respecting the dark theme

themarcel:
This is not a complete fix but I think it works well enough and the change is minimal.

The correct thing to do would be to integrate the theme with the ghbtns.com iframe but [seems there is no dark theme yet](https://github.com/mdo/github-buttons/issues/129).

Before the fix:
image

After the fix:
image


fricklerhandwerk (approved):
Thank you very much!

PR #1217, opened 2026-03-14
docs: update declarative-shell.md shellHook comment

andrew-field:
I believe the comment about when the shellHook attribute is executed as a command should be updated to after initialisation and not before entering the shell environment. How would the bash command have access to those packages before that environment designed to fetch those packages is setup? I think this is also consistent with the note on shellHook described here: https://nix.dev/tutorials/nix-language#worked-examples

Please disregard this pull request if my understanding is wrong!

### DIFF
--- a/source/tutorials/first-steps/declarative-shell.md
+++ b/source/tutorials/first-steps/declarative-shell.md
@@ -151,7 +151,7 @@ If you need to override these protected environment variables, use the `shellHoo
 
 ## Startup commands
 
-You may want to run some commands after the shell environment has initialized.
+You may want to run some shell commands before entering the interactive shell environment.
 These commands can be placed in the `shellHook` attribute provided to `mkShellNoCC`.
 
 Set `shellHook` to output a colorful greeting:


fricklerhandwerk:
When this was written, the focus was on accessibility for beginners. Even now I had to think for a second what "after environment initialization" even means. Sure, it's technically correct that Bash starts, shellHook commands run, then you do things. The emphasis was on "first some stuff happens, then you do things", which is why it said "before entering the shell" -- it was not supposed to mean "before the Bash process runs".

I'm all for being technically correct, but I would really like it be legible to the interested but inexperienced audience. So I'd suggest to expand the phrase to be more explicit about what happens, and ideally avoid the term "shell initialization" in case we're not defining it (which would take reading time, and that's always at odds with getting towards the learning goal).

andrew-field:
> When this was written, the focus was on accessibility for beginners. Even now I had to think for a second what "after environment initialization" even means. Sure, it's technically correct that Bash starts, shellHook commands run, then you do things. The emphasis was on "first some stuff happens, then you do things", which is why it said "before entering the shell" -- it was not supposed to mean "before the Bash process runs".
> 
> I'm all for being technically correct, but I would really like it be legible to the interested but inexperienced audience. So I'd suggest to expand the phrase to be more explicit about what happens, and ideally avoid the term "shell initialization" in case we're not defining it (which would take reading time, and that's always at odds with getting towards the learning goal).

Thanks very much! I understand it better myself.

fricklerhandwerk (approved):

PR #1232, opened 2026-06-01
document pinning nixpkgs pin for NixOS by npins from `system.nix`

KiaraGrouwstra:
now that 26.05 is finally stable, i guess we can now document using npins with nixos on here now 😶

### DIFF
--- a/source/guides/recipes/dependency-management.md
+++ b/source/guides/recipes/dependency-management.md
@@ -110,7 +110,7 @@ All the imported entries will be updated, so they won't necessarily point to the
 ## Managing NixOS configurations
 
 NixOS defaults to using channels to locate `nixpkgs`.
-You can instead pin a version using `npins` from the `system.nix` entrypoint:
+You can instead pin a version using `npins` from the `system.nix` entrypoint (available since 26.05):
 
 ```nix
 let
@@ -121,7 +121,13 @@ import "${sources.nixpkgs}/nixos" {
 }
 ```
 
-You can then disable channels in your configuration:
+Before NixOS 26.05's `system.nix` use:
+
+```bash
+sudo NIX_PATH="nixos-config=configuration.nix:nixpkgs=$(nix-instantiate --raw --eval npins -A nixpkgs.outPath)" nixos-rebuild switch
+```
+
+If you use npins with `system.nix`, disable channels in your configuration:
 
 ```nix
 # configuration.nix
--- a/source/guides/recipes/dependency-management.md
+++ b/source/guides/recipes/dependency-management.md
@@ -109,11 +109,26 @@ All the imported entries will be updated, so they won't necessarily point to the
 
 ## Managing NixOS configurations
 
-NixOS relies on the `NIX_PATH` environment variable to locate `nixpkgs`, which defaults to using channels.
-To instead use a `nixpkgs` version managed by `npins`, one can manually override this environment variable on system rebuilds:
+NixOS defaults to using channels to locate `nixpkgs`.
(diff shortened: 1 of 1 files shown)


source/guides/recipes/dependency-management.md
```diff
@@ -109,11 +109,26 @@ All the imported entries will be updated, so they won't necessarily point to the
 
 ## Managing NixOS configurations
 
-NixOS relies on the `NIX_PATH` environment variable to locate `nixpkgs`, which defaults to using channels.
-To instead use a `nixpkgs` version managed by `npins`, one can manually override this environment variable on system rebuilds:
+NixOS defaults to using channels to locate `nixpkgs`.
+You can instead pin a version using `npins` from the `system.nix` entrypoint:
```
fricklerhandwerk: That makes it sound like "the `system.nix` entry point" is some sort of first-class concept.

KiaraGrouwstra: its [release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2605.section.md) entry mentions one could straight-up have their entire config in there, i.e. have a working system for the purpose of those `nixos-rebuild` / `nixos-install` without use of any other files.
that makes me wonder: what do you mean here when you'd consider it not first-class?

fricklerhandwerk: I wasn't aware of that change, but now that you remind me, I remember seeing some discussion around it when it was still in the works.

In that case, how about linking to the NixOS (or more precisely, `nixos-rebuild` and `nixos-install`) reference documentation on the subject? Otherwise the initial impression of "you have to know to know" still holds.

KiaraGrouwstra: as per the release notes, its use-case that needs passing some extra thing is when it's used as an attrset of nixos configurations, in which case the commands need to know which one to use.
that feature isn't invoked in the example here, and i've been able to use the `system.nix` entrypoint myself without really having needed to be aware of how it can be used: as far as on-par functionality goes, it 'just works'.

source/guides/recipes/dependency-management.md
```diff
@@ -109,13 +109,34 @@ All the imported entries will be updated, so they won't necessarily point to the
...
+```nix
+let
+  sources = import ./npins;
+in
+import "${sources.nixpkgs}/nixos" {
+  configuration = ./configuration.nix;
+}
+```
+
+Before NixOS 26.05's `system.nix`, one may explicitly specify the `nixpkgs` pin like:
```
hsjobeki:
```suggestion
Before NixOS 26.05's `system.nix` use:
```

I'll unify the nix.dev stylguide with the nixpkgs one.
But essentially: "Address the user as 'you'" and "Make the subject - In this case the user - do the action"

source/guides/recipes/dependency-management.md
```diff
@@ -109,13 +109,34 @@ All the imported entries will be updated, so they won't necessarily point to the
...
 
 ```bash
 sudo NIX_PATH="nixos-config=configuration.nix:nixpkgs=$(nix-instantiate --raw --eval npins -A nixpkgs.outPath)" nixos-rebuild switch
 ```
 
+When using npins this way, you can disable channels in your configuration:
```
hsjobeki:
```suggestion
If you use npins with system.nix, disable channels in your configuration:
```

hsjobeki (commented):
LGTM overall. 

I'll port over the new styleguide from nixpkgs in a seperate PR. The current one in nix.dev is a bit lax.

hsjobeki (approved):

PR #1243, opened 2026-06-28
Clarify a confusing intro sentence

eoghanmurray:
This was quite confusing to a newcomer, I initially read it as

"using a virtual machine ... on a “bare metal” computer"

i.e. that the virtual machine necessitated a bare metal computer in order to be installed, i.e. probably exact opposite of intended communication.

(I read "independent of an installation" being a standalone aside instead of being connected to 'bare metal')

### DIFF
--- a/source/tutorials/nixos/nixos-configuration-on-vm.md
+++ b/source/tutorials/nixos/nixos-configuration-on-vm.md
@@ -4,7 +4,7 @@
 
 One of the most important features of NixOS is the ability to configure the entire system declaratively, including packages to be installed, services to be run, as well as other settings and options.
 
-NixOS configurations can be used to test and use NixOS using a virtual machine, independent of an installation on a "bare metal" computer.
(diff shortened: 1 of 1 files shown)


source/tutorials/nixos/nixos-configuration-on-vm.md
```diff
@@ -4,7 +4,7 @@
 
 One of the most important features of NixOS is the ability to configure the entire system declaratively, including packages to be installed, services to be run, as well as other settings and options.
 
-NixOS configurations can be used to test and use NixOS using a virtual machine, independent of an installation on a "bare metal" computer.
+NixOS configurations can be used to test and use NixOS using a virtual machine, without having to do a full "bare metal" installation.
```
eoghanmurray:
```suggestion
NixOS configurations can be used to test and use NixOS using a virtual machine, which is a lighter weight option compared to a full "bare metal" installation.
```

hsjobeki (approved):

PR #1255, opened 2026-07-07
fix instructions to optionally build manuals in devmode

fricklerhandwerk:
This also fixes the local display of links to the Nix manuals when
they're disabled.


hsjobeki:
Why dont we use direnv?

fricklerhandwerk:
How is this related?

hsjobeki:
Ah i see, i thought we could leverage direnv switch somehow to switch into the other shell.
But it seems thats not possible.

friedow (approved):
lgtm

PR #1258, opened 2026-07-07
link to build helpers instead of trivial builders in language tutorial

fricklerhandwerk:
For readers at that stage, trivial build helpers ("trivial builders" as
a term has been deprecated for a while already) are a special case, and
what's trivial about them makes no sense in that context yet.


hsjobeki:
As a user I find it very confusing to determine build-helpers from trivial-builders. 

This is some nixpkgs maintainer organization model presented as if it were a learning journey. I don't think we need to teach users this distinction at all.

PR #1261, opened 2026-07-07
add more authors and editors to tutorials and guides

fricklerhandwerk:
Closes #543

Depends on #1260

This is done mostly from memory and cursory inspection of commit history.
It may still be incomplete or a bit inaccurate, but it should do its job as a first way-better-than-nothing approximation.

PR #1262, opened 2026-07-07
show list of contributors and their articles

fricklerhandwerk:
Depends on #1261

image


hsjobeki (approved):
Looks good to me. I would like to give it a try myself before merging

PR #1264, opened 2026-07-12
Prototype site: some appearance tweaks

trespaul:
I figured each of these changes were too small for their own PRs:

- ~add *NixOS* Documentation to site title~
- set favicon in Astro config, remove redundant copy, optimise SVG (-60%)
- remove custom components (I personally like the sections at the top of the sidebar — are there reasons for putting them in the header that I'm not realising?)
- use colours from the brand guideline
- remove hairline borders (so page feels more open)
- make banner less bright, round corners to stylistically fit in when borders are removed

image
image

In general, though, I personally think a custom template would be best, instead of Starlight. Over the past year, I've been seeing the exact same website over and over again, and I think it would be good if we could differentiate this one a bit. If the branding team creates a unified UI components style, that would be great. But with these changes I at least tried to make it look a bit more like its own thing for now, e.g. by removing the hairline borders. (I know it's early days and the plan probably was never to stay with the default Starlight theme, but I thought let me play around with some colours and things so long.)

My argument would just be that, if basically every component has to be overridden, I'm not sure if the remaining benefits of Starlight would make it worth keeping over just making a custom site. Is there a benefit of Starlight we would be missing?


hsjobeki:
The reason I didn't put NixOS in the title was that I thought it might be confusing. Since 
NixOS is not nixpkgs and if you are viewing the nixpkgs docs. A nixos title would be potentially confusing

hsjobeki:
I like the improvements quite a lot. +1 
I also put the topic switchers in the sidebar in my prototype. @friedow voted for having them in the header. I personally find them better in the sidebar. 

For a custom template. Yes please feel free to go ahead. I wouldnt wait for the marketing team here and go straight ahead. I don't think we have that much to share beyond colors and typography, since the docs are mostly prose content based.

If you want to discuss any ideas i'm available the whole next week.

trespaul:
> A nixos title would be potentially confusing

True, I did this late-ish last night and just took the title from ~some existing site~ the title in the hero section. How about "Nix Documentation"?

hsjobeki:
I'm not sure if this is good reasoning, but here is my thought behind the title:

"Nix-icon  documentation" 

And the logo itself stands uniformly for all parts of the ecosystem: nix, nixpkgs and nixos likewise.

trespaul:
> And the logo itself stands uniformly for all parts of the ecosystem: nix, nixpkgs and nixos likewise.

It does make sense, but the reason I changed it was also partly because the site title is used in the `<head>` and thus in search engine results, and is appended to page titles, so page titles would currently look like "Installation Guide | Documentation". So if we wanted to change just the header we'd have to do that with a component override.

hsjobeki (approved):

PR #1266, opened 2026-07-13
Replace Sphinx search with Pagefind

KiaraGrouwstra:
Sphinx's stock client-side search stems and splits identifiers like `system.nix` into ubiquitous terms (`system` + `nix`) and stores no positional data, so phrase/identifier ranking is impossible: searching `system.nix` [did not (highly) rank](https://discourse.nixos.org/t/discussion-of-flakes-from-2025-nixos-community-survey-report/78853/39) `guides/recipes/dependency-management.html`, which contains the exact string.

Pagefind indexes the rendered HTML post-build and does real phrase/proximity ranking. Searching `system.nix` now ranks that page first.

- `pagefind.yml`: scope indexing to the article body, exclude UI chrome.
- `default.nix`: add `pkgs.pagefind` to `nativeBuildInputs` (also exposes it in the dev shell via `inputsFrom`).
- `Makefile`: run `pagefind --site build/html` after `sphinx-build` in the `html` target for local/CI parity, guarded with a clear error if the binary is missing.
- `source/_templates/search.html`: replace the Sphinx search page with a Pagefind UI, dropping the now-unused Sphinx search scripts and pre-filling from `?q=` so the sidebar search box still works.

Verifying:

- before: https://nix.dev/search.html?q=%22system.nix%22#
- running this PR locally: `nix-shell --run 'make html' && python3 -m http.server -d build/html 8000`
- after: http://127.0.0.1:8000/search.html?q=%22system.nix%22

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,11 @@ html:
 	# hack to make live-reload on css update work
 	@cp source/_static/css/custom.css build/html/_static/css/custom.css
 
+	# build the Pagefind search index over the rendered HTML (writes build/html/_pagefind/)
+	@command -v pagefind >/dev/null 2>&1 \
+		|| { echo "error: 'pagefind' not found on PATH; install it or run inside the Nix dev shell (nix-shell)"; exit 1; }
+	pagefind --site build/html
+
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
--- a/default.nix
+++ b/default.nix
@@ -36,6 +36,7 @@ let
     nativeBuildInputs = [
       nix-dev-python-pkgs
       nix-dev-latex
+      pkgs.pagefind
     ];
     buildPhase =
       let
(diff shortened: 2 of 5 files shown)


hsjobeki:
I like the idea.

Tested this locally. 

I think excerpts could be longer.

New

image

Current

image


I'd like to try some other queries still. But pagefind is a sane default for a fully static page imho.

KiaraGrouwstra:
> I think excerpts could be longer.

i bumped the snippets up to 30 words now

hsjobeki:
@KiaraGrouwstra : Resolved the conflicts. 
The pagefind styling now follows the main theme, so dark mode also works.

Most notably i changed the description of your original commit message. 
It still reads a bit AI-generated but at least is more understandable and doesn't just re-state the file-list I can already get from the diff. 

I tested the change locally. Using pagefind makes sense to me, it also aligns better with the upcoming docs.nixos.org

PR #1268, opened 2026-07-13
add warning about caches as an attack vector

fricklerhandwerk:
For [Spectrum](https://spectrum-os.org), [funded by NLnet](https://nlnet.nl/project/Spectrum-virt/)


xokdvium (approved):
Thanks! I guess it's not going into details intentionally - that would be better suited in the nix manual, but the current page on security is quite anemic https://nix.dev/manual/nix/2.35/installation/nix-security.html

hsjobeki (approved):

PR #1279, opened 2026-08-29
site: balance styling against nixos branding identity 

hsjobeki:
cc @trespaul


site/src/styles/global.css
```diff
@@ -73,5 +74,9 @@
   --color-grey-90: oklch(0.30 0 0); /* #2e2e2e */
   --color-grey-94: oklch(0.22 0 0); /* #1b1b1b */
   --color-grey-95: oklch(0.18 0 0); /* #121212 */
+  --color-grey-96: oklch(0.17 0 0); /* #0f0f0f */
 
+  /* More colors for dark mode */
+  --color-steel-95: oklch(0.21 0.008 274.5) /* #17181c; */
```
trespaul: Semicolon!

site/src/layouts/Base.astro
```diff
@@ -9,8 +9,9 @@ const { title, description } = Astro.props;
 
 <html lang="en">
   <Head title={title} description={description} />
-  <body class="flex flex-col min-h-dvh gap-y-8
-    bg-grey-05 text-grey-95 dark:bg-grey-95 dark:text-grey-05"
+  <body
+    class="flex flex-col min-h-dvh gap-y-8
+    bg-white text-black dark:bg-steel-95 dark:text-mainsail-05"
```
trespaul: `mainsail-10`?

trespaul (commented):
Missing semicolon and `mainsail-10` is undefined (or `mainsail-05` is unused).

Making light mode higher contrast is fair, I'm probably biased towards making it less bright. But dark mode is also now lower-constrast. I prefer the slightly higher contrast — what's your reasoning?

But we don't have to bikeshed about the colours (since they'll probably change anyway as we combine design efforts with the marketing team). so I'm fine accepting these colours.

PR #1280, opened 2026-08-29
site: add prettier for formatting

hsjobeki:
Adds a CI gate on formatted changes.

I noticed formatting churn when trying to change some simple things. (css variables, astro code) 

We can switch to another formatter later if we like.


.github/workflows/site-format.yml
```diff
@@ -0,0 +1,36 @@
...
+  contents: read
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: site
+    steps:
+      - uses: actions/checkout@v4
```
friedow: The latest version of checkout is v7.


```suggestion
      - uses: actions/checkout@v7
```

.github/workflows/site-format.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: site/package-lock.json
```
friedow: Similar to above, setup-node's latest version is v7. I think using the default node version here should be fine which defaults to the latest lts (currently 24).


```suggestion
      - uses: actions/setup-node@v7
        with:
          cache: npm
          cache-dependency-path: site/package-lock.json
```

site/package.json
```diff
@@ -7,7 +7,9 @@
     "start": "astro dev",
     "build": "astro build",
     "preview": "astro preview",
-    "astro": "astro"
+    "astro": "astro",
+    "format": "prettier --write .",
```
friedow, line 11: It would be nice to have prettier added as a nix formatter flake output to enable the use of `nix fmt` on the repository. The easiest way to do this is probably using treefmt-nix. Furthermore, the github action using the formatter could then run the formatter through nix.

This ensures that we have no differences in environment (mainly nodejs version) between local development environments and ci.

friedow: Nevermind, I've just noticed that this repo does not use flakes.

PR #1283, opened 2026-09-15
docs site: fix styles and make interactive

trespaul:
- use Astro's fontProvider to add fonts, use inter-ui
  - fontProvider is the (new-ish) standard way of adding fonts in Astro; fonts can be added from different sources, including local files and npm packages. Usually it fetches packages on the fly, but this doesn't work in nix, so we install the package as a dependency and set `remote: false`. Processing the font files through fontProvider also means Astro performs some optimisations on the font files, instead of serving them straight from ./public
  - inter-ui: fontsource provides the same font files as Google Fonts, which aggressively strips font features such as extended weights and ligatures. inter-ui gives us the option to use those features if we choose.

- ExpressiveCode style: use built-in themes; hook into theme switcher
  - Custom code theme used my new colours, and was more of an intellectual exercise. catppuccin is perfectly serviceable and is bluish to fit with the Nix brand colours.
  - ExpressiveCode by default uses a theme media query to determine whether to use the dark theme; with the custom theme switcher which uses a class, we need to tell ExpressiveCode which class indicates which theme.

- header styling
  - don't use backdrop-blur, use a slightly darker or lighter shade of grey
  - elevate header z-index to be above ExpressiveCode blocks
  - centre items
  - increase icons sizes slightly
  - increase space between header and content

- add Alpine
  - Alpine is very small and straightforward; it mostly uses attributes on elements to set behaviour.
  - add entrypoint with stores to hold state (state is usually in `x-data`), but since the first common ancestor of the controller and controlled components is `<body>`, using `x-data` there would mean reactivity is scoped to the entire site. Stores are registered in the entrypoint.
  - `Window` interface in ./src/env.d.ts registers Alpine with the language server for intellisense.
  - `[x-cloak]` CSS utility hides conditionally rendered components before render so they don't flash before Alpine is initialised.

- make sidebar interactive
  - An empty `x-data` is necessary to make elements reactive.
  - Toggle button triggers nav store's `toggle()`; different icons are conditionally rendered.
  - Sidebar `<section>` can't used `x-show`, because it needs to show unconditionally on larger screens. Instead, hide using `hidden`, show unconditionally using `md:block`, and conditionally cancel out `hidden` with `!block` (`!` being `!important` in Tailwind).

- prose style: fix table padding & code font size
  - Table padding was wrong, I think this was just a typo.
  - Code font is slightly larger than heading font, which made it not fit well, so reduce it to 95%.

- add route for nixpkgs manual content

- use details element to collapse outline entries
  - Browsers provide `<details>` with native interactivity. We just need to set which should start open, and remember state between navigations.
  - Also add collapse depth parameters to ToC item so long.

- add example code for outline to sidebar


trespaul:
Preview with nixpkgs manual: https://preview-nix-docs.trespaul.workers.dev/

hsjobeki (approved):
Went through the commits and the changes. Obviously there is more follow up work needed. 

Just a short list, so we don't forget:

- Mobile nav seems to have a dead reference.  aria-controls="mobile-menu"
- tocStore seems dead code. alpine.ts registers $store.toc
- Every nav item is wrapped in `details`,  as discussed in the weekly meeting. I think this needs a seperate sidebar PR (planning to do that)

This PR aligns a lot of the fundametals. 👍 

The follow up's can be done in the next PRs.

trespaul:
> Mobile nav seems to have a dead reference.  aria-controls="mobile-menu"

Yes, forgot about that, sorry. I do already have "check aria attributes" on my personal to-do list, at least.

> * tocStore seems dead code. alpine.ts registers $store.toc

Not dead, just not born yet — intended for the collapsible table of contents.

> * Every nav item is wrapped in `details`,  as discussed in the weekly meeting. I think this needs a seperate sidebar PR (planning to do that)

I was planning on setting the depth up to which the `<details>` should start open, but if you want to change it to something else, that's also good.

hsjobeki:
> I was planning on setting the depth up to which the `<details>` should start open, but if you want to change it to something else, that's also good.

The nav.json of each manual describes what should be open. See: https://github.com/NixOS/nix.dev/pull/1289 
though it also doesn't wire it up yet.

PR #1289, opened 2026-09-18
Render the nixpkgs manual

hsjobeki:
How the source preprocessing works:

1. Sync the doc sources out of nixpkgs into site/nixpkgs-doc. This happens in the devShell
2. astro build or astro dev starts and syncs the content layer, running three loaders in content.config.ts
3. Each loader awaits loadDocIndex() (build an anchor map from the manual)

The index is then used for during the conversion from nixpkgs-markdown to unfified commonmark replacing the internal anchors with docs.nixos.org ones.

We then use sätteri to do regular markdown parsing and rendering.


site/src/content.config.ts
```diff
@@ -1,15 +1,109 @@
 import { defineCollection } from "astro:content";
-import { glob } from "astro/loaders";
+import { type Loader } from "astro/loaders";
 import { z } from "astro/zod";
 
+import { loadDocIndex } from "./lib/nixpkgs-doc";
+
+/** One entry per markdown file the nixpkgs nav lists. */
```
friedow, line 7: This currently produces empty pages for sections that only had a headline like "Using Nixpkgs":

Image

https://nixos.org/manual/nixpkgs/stable/#part-using

site/src/content.config.ts
```diff
@@ -1,15 +1,109 @@
...
+            summary: entry.summary,
+            source: entry.source,
+          },
+        }),
+      });
+    }
+  },
+};
+
+/** The groups the functions are filed under, in the order nixdoc exports them. */
```
friedow, line 55: Can we render one more level of sidebar for the lib functions? Currently I have to go to a function set to see its subpages. Instead I'd like to be able to see the functions in the sidebar

Image

site/src/pages/nixpkgs/[...slug].astro
```diff
@@ -1,19 +1,93 @@
...
+          title: `lib.${props.group.data.id}`,
+          description: props.group.data.summary,
+          functions: props.functions,
+        }
+      : {
+          doc: props.entry,
+          title: props.entry.data.attrPath,
+          description: props.entry.data.summary,
+          functions: undefined,
+        };
```
Comment (friedow) [inline site/src/pages/nixpkgs/[...slug].astro:62]: Not the biggest fan of those ternary operator chains. Instead simple `if else` would be better to read.

How about putting an agent skill into this repo that documents those style choices to enforce them continuously?

site/src/content.config.ts
hsjobeki: I'll have a look but my quick guess is that this is because the file has no content in nixpkgs itself.
Deleting the file in nixpks would be the fix :D

site/src/content.config.ts
hsjobeki: But they are ~200 items in total. That's why i wanted to defer building a widget or something more dedicated for them in a later PR.

site/src/pages/nixpkgs/[...slug].astro
Comment (trespaul) [reply site/src/pages/nixpkgs/[...slug].astro:62]: I think you'd need:

```typescript
(props => {
  if ("page" in props) {
    return ...;
  }

  if ("group" in props) {
   return ...;
  }

  return ...;
})(props)
```

site/src/content.config.ts
bew: FWIW For me ideally the left sidebar would stay about pages, and there'd be a right sidebar for items in the current page (maybe be collapsed by default) 🤔

hsjobeki:
I have split out the most important work into #1292 so we bring the lines changed of this PR down.
After #1292 is merged, i'll rebase this PR.

site/src/content.config.ts
hsjobeki: https://github.com/NixOS/nixpkgs/pull/565872

PR #1290, opened 2026-09-18
site: improve header, footer, nav, tables

trespaul:
- decrease text size in table to avoid very narrow columns
- set nav position fixed to prevent scrolling off the screen, and set top-0 and bottom-0 to constrain size, so that overflow-scroll works
- header tweaks
  - set icons to same size
  - move mobile menu toggle into header-right
  - keep theme switcher on smaller screens, only collapse social icons
  - remove redundant div inner wrapper
  - set header height to keep constant
- improve footer spacing


friedow (approved):
lgtm

